PR #4, opened 2019-12-04
Add spearman job

oyvindeide:


jondequinor:
Could you describe what this does? Preferably in the code.

semeio/jobs/scripts/spearman_correlation.py
```diff
@@ -0,0 +1,47 @@
...
+        cophenetic distance than `t`.
+        """,
+    )
+    parser.add_argument(
+        "--output-file",
+        required=False,
+        type=str,
+        help="Name of the outputfile. The format will be yaml.",
+    )
+    parser.add_argument(
```
jondequinor: Why would you want a dry run?

semeio/jobs/spearman_correlation_job/job.py
```diff
@@ -0,0 +1,82 @@
...
+    measured_data.remove_inactive_observations()
+
+    simulated_data = measured_data.get_simulated_data()
+
+    correlation_matrix = _calculate_correlation_matrix(simulated_data)
+
+    clusters = _cluster_analysis(correlation_matrix, threshold)
+
+    columns = correlation_matrix.columns
+    data = list(
```
jondequinor: Perhaps a small comment as to why this is done?

tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+import os
+import pytest
+
+@pytest.fixture()
+def setup_tmpdir(tmpdir):
+    cwd = os.getcwd()
+    tmpdir.chdir()
+    yield
+    os.chdir(cwd)
```
jondequinor: There's a newline missing here.

semeio/jobs/scripts/spearman_correlation.py
oyvindeide: If the user would like to know what clusters are created, but would not necessarily run the scaling. But doesn't make much sense without more printing i suppose.

oyvindeide: Printing added

PR #15, opened 2020-01-03
Porting design2param

oysteoh:
Depends on #20


tests/jobs/design2params/test_design2params.py
```diff
@@ -0,0 +1,99 @@
...
+    with open("refdesignparameters_w_default.txt") as ref_designparameters_file:
+        ref_designparameters = ref_designparameters_file.read()
+    assert designparameters == ref_designparameters
+
+    with open(design2params._TARGET_FILE_TXT, "r") as status_file:
+        status = status_file.read()
+    assert status == "OK\n"
+
+
+def test_open_excel_file_not_existing(tmpdir, input_data):
```
oyvindeide: `tmpdir` is never used

tests/jobs/design2params/test_design2params.py
```diff
@@ -0,0 +1,99 @@
...
+    with pytest.raises(SystemExit):
+        design2params._read_excel("not_existing.xlsx", "DesignSheet01")
+
+
+def test_open_excel_file_not_xls(tmpdir, input_data):
```
oyvindeide: `tmpdir` is never used

tests/jobs/design2params/test_design2params.py
```diff
@@ -0,0 +1,99 @@
...
+    with pytest.raises(SystemExit):
+        design2params._read_excel("parameters.txt", "DesignSheet01")
+
+
+def test_open_excel_file_header_missing(tmpdir, input_data):
```
oyvindeide: `tmpdir` is never used

tests/jobs/design2params/test_design2params.py
```diff
@@ -0,0 +1,99 @@
...
+    with pytest.raises(SystemExit):
+        design2params.run(
+            1, "design_matrix_missing_header.xlsx", "DesignSheet01", "DefaultValues"
+        )
+
+
+def test_open_excel_file_value_missing(tmpdir, input_data):
```
oyvindeide: `tmpdir` is never used

semeio/jobs/design2params/design2params.py
```diff
@@ -0,0 +1,173 @@
+import logging
+import sys
+
+import numpy as np
+import pandas as pd
+
+
+def _create_handler(formatter, stream, level):
```
oyvindeide: Should there be a separate log handler on a job basis? Maybe in a shared location or be the responsibility of `ert`? @markusdregi, @jondequinor, we had a discussion on this earlier, any thoughts?

oysteoh: Yes , we should probably make a common one which we can use across the jobs . Lets discuss on monday !

semeio/jobs/scripts/design2params.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def valid_file(arg):
+    if os.path.isfile(arg):
+        return arg
+    raise argparse.ArgumentTypeError("{} is not a existing file!".format(arg))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description=description)
```
oyvindeide: Might be better to move the parser into its own function for readability?

tests/jobs/design2params/test_design2params.py
```diff
@@ -0,0 +1,99 @@
...
+
+    cwd = os.getcwd()
+    tmpdir.chdir()
+
+    yield
+
+    os.chdir(cwd)
+
+
+def test_run(tmpdir, input_data):
```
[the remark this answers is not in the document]
oysteoh: nice one :+1:

oysteoh:
@oyvindeide  have a second look? pulled out logging initiation as a starting point of our discussion further with regards to a basic logging func - how do you think this is now? should we do this differently , opinions?

oyvindeide:
> @oyvindeide have a second look? pulled out logging initiation as a starting point of our discussion further with regards to a basic logging func - how do you think this is now? should we do this differently , opinions?

Think this is much better, but any reason you're returning the new logger instead of adding a handler to the `logging` class and calling it in `__init__.py`? I feel the benefit of doing that is each module is more general by just doing:

```
import logging
logger = logging.getLogger(__name__)
```

and replacing
`return logger` with `logger.addHandler(handler)`

Could also make it easier in the future to change the handlers if it will be handled by `ert` in some way, https://docs.python.org/3/howto/logging.html#library-config.

What do you think?

semeio/jobs/scripts/design2params.py
```diff
@@ -0,0 +1,68 @@
...
+        realization=args.realization,
+        xlsfilename=args.xlsfilename,
+        designsheetname=args.designsheetname,
+        defaultssheetname=args.defaultssheetname,
+        log_level=args.log_level,
+    )
+
+
+if __name__ == "__main__":
+    main()
```
oyvindeide: Missing `args`, should perhaps have a test on the entry point?

semeio/jobs/scripts/design2params.py
```diff
@@ -0,0 +1,68 @@
...
+Requires a matrix with column header as strings in topmost row in Excel.
+Row 2 in Excel must then correspond with the data you want for 'realization 0'
+Column 1 in Excel should contain the realization number, and will be ignored by this script.
+You must run this script as a FORWARD_MODEL from your ERT config, before
+you run DESIGN_KW.
+
+Usage:
+ $ design2params.py <realization-id> <xls-filename> <worksheet-name> [<defaultssheetname>]
+defaultssheetname is an optional argument, you may choose not to provide it (just give an empty string)
+Haavard Berland, Oseberg Petec, havb@statoil.com, April-June 2015
```
oyvindeide: A bit nit picky, but this should perhaps be removed? Guess we'll be building documentation based on the argparser, and we're taking over maintenance.

semeio/jobs/scripts/design2params.py
oysteoh: ah, didn't see your suggestion before reimplementing - but i think i've done almost what you suggested

semeio/jobs/scripts/design2params.py
```diff
@@ -0,0 +1,68 @@
...
+    parser.add_argument(
+        "defaultssheetname",
+        type=str,
+        nargs="?",
+        help="Defaults sheet name",
+        default=None,
+    )
+
+    parser.add_argument(
+        "--log-level", required=False, default="WARNING", type=logging.getLevelName,
```
oyvindeide: Could add a single letter option to make `ert` integration easier.

semeio/jobs/design2params/design2params.py
```diff
@@ -0,0 +1,166 @@
...
+
+import numpy as np
+import pandas as pd
+
+import logging
+
+
+# Filenames created/modified by this script. Don't change unless you understand the consequences.
+_TARGET_FILE_TXT = "DESIGN2PARAMS.OK"
+_PARAMETERS_TXT = "parameters.txt"
```
oyvindeide: This requires that we run the job in the same folder as `parameters.txt`, will that always be the case? Would be nice to do the parsing of `parameters.txt` before this job to make it more independent of the file system. Could add it as an optional argument with a default value.

semeio/jobs/scripts/design2params.py
```diff
@@ -0,0 +1,68 @@
...
+ $ design2params.py <realization-id> <xls-filename> <worksheet-name> [<defaultssheetname>]
+defaultssheetname is an optional argument, you may choose not to provide it (just give an empty string)
+"""
+
+
+def parse_args(args):
+    parser = argparse.ArgumentParser(description=description)
+
+    parser.add_argument(
+        "realization", type=int, help="Which realization",
```
oyvindeide: When I use a different realization than `1` the job fails, example of failing test:
```
@pytest.mark.usefixtures("input_data")
@pytest.mark.parametrize("realization_id", [(1), (2), (3)])
@pytest.mark.parametrize(
    "test_file, expected_file",
    [
        (design2params._PARAMETERS_TXT, "refparameters_w_default.txt"),
        (design2params._DESIGN_MATRIX_TXT, "refdesignmatrix.txt"),
        (design2params._DESIGN_PARAMETERS_TXT, "refdesignparameters_w_default.txt"),
    ],
)
def test_run_with_default(test_file, expected_file, realization_id):
    design2params.run(
        realization_id,
        "design_matrix_tampered_defaults.xlsx",
        "DesignSheet01",
        "DefaultValues",
        log_level=logging.DEBUG,
    )

    with open(design2params._TARGET_FILE_TXT, "r") as status_file:
        status = status_file.read()
        assert status == "OK\n"
```

semeio/jobs/scripts/design2params.py
```diff
@@ -0,0 +1,68 @@
...
+   can be given to CUSTOM_KW
+Requires a matrix with column header as strings in topmost row in Excel.
+Row 2 in Excel must then correspond with the data you want for 'realization 0'
+Column 1 in Excel should contain the realization number, and will be ignored by this script.
+You must run this script as a FORWARD_MODEL from your ERT config, before
+you run DESIGN_KW.
+
+Usage:
+ $ design2params.py <realization-id> <xls-filename> <worksheet-name> [<defaultssheetname>]
+defaultssheetname is an optional argument, you may choose not to provide it (just give an empty string)
```
oyvindeide: `you may choose not to provide it (just give an empty string)` being optional you shouldn't have to provide an empty string?

oyvindeide: An on the command line at least you don't have to give an empty string either.

semeio/jobs/scripts/design2params.py
lars-petter-hauge: Seconded, if we want to use e.g. the documentation in argparse in sphinx we need a function that returns the parser only

PR #18, opened 2020-01-06
Gendata rft

lars-petter-hauge:


tests/jobs/rft/data/norne/norne.py
```diff
@@ -0,0 +1,87 @@
+from itertools import product
```
jondequinor: How is this test run? Or; is it a test, or something used in testing?

Could the answer be included in the file itself as a comment?

tests/jobs/rft/data/norne/expected_result/RFT_B-4AH_0_active
```diff
@@ -0,0 +1,29 @@
+1
+1
+1
+1
+1
+1
```
oyvindeide: Are these files really necessary to check in? Can't they be mocked?

tests/jobs/rft/data/norne/norne.py
lars-petter-hauge: This was the script created to generate the data required as input for the `gen_data` job. We made it early on to test both implementations.

This doesn't generate the input completely though, the zone descriptions have been added manually (e.g. `zone2_and_3`). So some modifications would be required to have the script completely generate everything.

I'm not sure if I would like this script to generate the test data at each run, or have them pre-generated (as they are now), and remove this script.

Either way, I don't think we need both

jondequinor: Okay, then let's remove it.

tests/jobs/rft/data/norne/expected_result/RFT_B-4AH_0_active
lars-petter-hauge: Let's add them in the test, fixing

tests/jobs/rft/test_trajectory.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for line, expected_zone in zip(["0 1 2.2 3", "0 1 2.2 3 zone"], [None, "zone"]):
+            point = TrajectoryPoint(*parse_trajectory_line(line))
+            self.assertEqual(point.utm_x, 0)
+            self.assertEqual(point.utm_y, 1)
+            self.assertEqual(point.measured_depth, 2.2)
+            self.assertEqual(point.true_vertical_depth, 3)
+            self.assertEqual(point.zone, expected_zone)
+
+    def test_invalid_load_from_line(self):
```
oyvindeide: You know I like `parameterized`, but wouldn't this be more readable:
```suggestion
@pytest.mark.parametrize(
    "lines,expected_error",
    [
        ("1", "Trajectory data file not on correct format"),
        ("a b c", "Trajectory data file not on correct format"),
        ("a b c d e", "Error: Failed to extract data from line"),
    ],
)
def test_invalid_load_from_line(lines, expected_error):
    with pytest.raises(argparse.ArgumentTypeError) as e_info:
        TrajectoryPoint(*parse_trajectory_line(lines))
    assert expected_error in str(e_info.value)
```

tests/jobs/rft/test_gendata_rft.py
```diff
@@ -0,0 +1,168 @@
...
+        "zonemap.txt",
+    ]
+
+    main_entry_point(arguments)
+
+    for expected_file in os.listdir(EXPECTED_RESULTS_PATH):
+        assert expected_file in os.listdir(tmpdir.strpath)
+        if not expected_file.endswith("inactive_info"):
+            assert filecmp.cmp(
+                expected_file, os.path.join(tmpdir.strpath, expected_file)
```
lars-petter-hauge: So this is super stupid, but the `expected_file` is only the base name, and as such what this function does is to only check that the same file, the one in `tmpdir.strpath` is identical.. 

In fact, there was a bug in the `validate_zonemap` which caused every well that did not have a zone described to be given `-1` for pressure.

tests/jobs/rft/test_trajectory.py
```diff
@@ -0,0 +1,103 @@
...
+            point.pressure = pressure
+            point.valid_zone = validation
+            self.assertEqual(all([ijk, pressure, validation]), point.is_active())
+            self.assertEqual(
+                1 if all([ijk, pressure, validation]) else -1, point.get_pressure()
+            )
+
+
+class TestTrajectory(unittest.TestCase):
+    @pytest.fixture(autouse=True)
```
oyvindeide: If you're using unittest, wouldn't it be better to just use `def setUp`? Or just drop the `class` and use pure pytest?

tests/jobs/rft/data/norne/norne.py
lars-petter-hauge: removed

tests/jobs/rft/data/norne/expected_result/RFT_B-4AH_0_active
lars-petter-hauge: added

tests/jobs/rft/test_gendata_rft.py
```diff
@@ -0,0 +1,168 @@
...
+ECL_BASE = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "data/norne/NORNE_ATW2013"
+)
+EXPECTED_RESULTS_PATH = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "data/norne/expected_result"
+)
+
+
+@pytest.fixture
+def input_data(tmpdir):
```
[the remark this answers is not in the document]
lars-petter-hauge: Added writing of parts of the test data, that added more complexity to `input_data`, let's see if I can get it all to play nice with your suggestion

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,59 @@
...
+
+    for p in trajectory_points:
+        p.set_ijk(eclGrid.find_cell(p.utm_x, p.utm_y, p.true_vertical_depth))
+        p.update_pressure_from_rft(rft)
+        p.validate_zone(zonemap)
+
+    return trajectory_points
+
+
+def run(well_times, trajectories, eclGrid, eclRftFile, zonemap):
```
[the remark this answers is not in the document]
lars-petter-hauge: Unsure.. fixed anyway!

lars-petter-hauge:
The reason it fails on 2.7 currently is the precision in py2 vs py3 has changed, might need to change `filecmp.cmp` to reading both files and asserting almost equals.

tests/jobs/rft/test_trajectory.py
lars-petter-hauge: Yeh, this is a leftover. The previous repo used unittest, hence why it was written in unittest first. Started removing unittest from `test_gendata_rft`, an apparently that is how far I got :smile: 

Will have everything run pytest like the rest of this repo

lars-petter-hauge: fixed

semeio/jobs/scripts/gendata_rft.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
+
+import argparse
+import datetime
```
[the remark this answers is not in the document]
lars-petter-hauge: fixed!

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,59 @@
+import sys
```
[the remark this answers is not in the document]
lars-petter-hauge: fixed

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def write_inactive_info(points, fname, zonemap):
+    with open(fname + "_inactive_info", "w+") as fh:
+        fh.write(
+            "\n".join([p.inactive_info(zonemap) for p in points if not p.is_active()])
+        )
+
+
+def populate_trajectory_points(
```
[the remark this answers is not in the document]
lars-petter-hauge: privatized!

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,59 @@
...
+def write_points(points, fname):
+    with open(fname, "w+") as fh:
+        fh.write("\n".join([str(p.get_pressure()) for p in points]))
+    logger.info("Forward model script gendata_rft.py: Wrote file {}".format(fname))
+
+    with open(fname + "_active", "w+") as fh:
+        fh.write("\n".join([str(int(p.is_active())) for p in points]))
+
+
+def write_inactive_info(points, fname, zonemap):
```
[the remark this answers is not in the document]
lars-petter-hauge: privatized!

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,59 @@
+import sys
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def write_points(points, fname):
```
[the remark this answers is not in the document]
lars-petter-hauge: privatized!

tests/jobs/rft/test_trajectory.py
lars-petter-hauge: I do know you like your `parameterized`, but this can just as easily be written as:
```
def test_invalid_load_from_line():
    lines_errors = [
        ("1", "Trajectory data file not on correct format"),
        ("a b c", "Trajectory data file not on correct format"),
        ("a b c d e", "Error: Failed to extract data from line"),
    ]

    for (line, error) in lines_errors:
        with pytest.raises(argparse.ArgumentTypeError) as err_context:
            TrajectoryPoint(*parse_trajectory_line(line))
        assert error in err_context.value.args[0]
```
So you don't need the pytest fixture to change it into the other way of writing ;)

Some really like the `zip`, I'm a bit back and forth.. But I agree that it's easier to match which input provides which error, so I'll change it. And I'll even use `parametrize`!

tests/jobs/rft/test_gendata_rft.py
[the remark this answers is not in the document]
lars-petter-hauge: So, I did a grep on `tmpdir` and found, as you said, a range of different variations on how it is used.
The 
```
@pytest.fixture()
def setup_tmpdir(tmpdir):
```
Is defined in duplicate places (and those only have a `yield`, without any finally. So if you change directory in between, the directory stays unchanged when you continue with your testing, resulting in consecutive tests failiing although it is really just the first that fails..)

To me it seems that we should define the fixture in one location and set it up generic. I don't see how the change in implementation above will make that easier.

My suggestion is that we create a PR afterwards that tries to combine every place where this is done.

semeio/jobs/config_jobs/GENDATA_RFT
```diff
@@ -0,0 +1,12 @@
+STDOUT        GENDATA_RFT.stdout
+STDERR        GENDATA_RFT.stderr
+EXECUTABLE    ../scripts/gendata_rft.py
+ARGLIST       <ECL_BASE> <PATH_TO_TRAJECTORY_FILES> <WELL_AND_TIME_FILE> <ZONEMAP>
```
ManInFez: The argument should use the corresponding short-name  in the args-list, i.e. `-e <ECL_BASE> .. -z <ZONEMAP>`

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+from ecl.grid import EclGrid
+from semeio.jobs.rft.zonemap import ZoneMap
+
+
+def strip_comments(line):
+    return line.partition("--")[0].rstrip()
+
+
+def existing_directory(path):
+    if not os.path.exists(path):
```
ManInFez: maybe use `os.path.is_dir()` instead of `os.path.exists`?

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+        raise argparse.ArgumentTypeError("The path {} does not exist".format(path))
+    return path
+
+
+def load_zonemap(path):
```
ManInFez: Why is this not a function of `ZoneMap`

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+            "Warning: Trajectory file {filename} not found!".format(filename=filename)
+        )
+
+    with open(filename, "r") as f:
+        trajectory_lines = f.readlines()
+
+    trajectory_lines = [strip_comments(l) for l in trajectory_lines]
+
+    for line in trajectory_lines:
+        if not line:
```
ManInFez: be explicit. `line is None or line is ""` or something along those lines.

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+
+    with open(filename) as well_time_file:
+        lines = well_time_file.readlines()
+
+    well_time_lines = [(strip_comments(l), i + 1) for i, l in enumerate(lines)]
+
+    for line, line_number in well_time_lines:
+        tokens = line.split()
+
+        if not tokens:
```
ManInFez: `tokens is []`

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+    Returns
+    -------
+    Dict
+        Returns a dict with k_value as keys and a list of corresponding
+        zonenames
+    """
+
+    zones_at_k_value = {}
+
+    with open(filename, "r") as f:
```
ManInFez: No test whether file exist...

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+        zonemap_lines = f.readlines()
+
+    zonemap_lines = [(strip_comments(l), i + 1) for i, l in enumerate(zonemap_lines)]
+    basic_err_msg = "Line {line_number} in ZoneMap file {filename} not on proper format: 'k zonename <zonename> ...'. "
+    for line, line_number in zonemap_lines:
+        zonemap_line = line.split()
+
+        if not zonemap_line:
```
ManInFez: `zonemap_line is []`

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+            continue
+
+        if len(zonemap_line) < 2:
+            raise argparse.ArgumentTypeError(
+                basic_err_msg
+                + "Number of zonenames can be 1 or more".format(
```
ManInFez: "can be" -> "must be"

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+    ----------
+    filename : string
+        Filename to open
+
+    Returns
+    -------
+    Tuple
+        Returns a tuple with an ecl grid instance and an rft instance
+    """
+    try:
```
ManInFez: verify if path exist?

semeio/jobs/scripts/gendata_rft.py
```diff
@@ -0,0 +1,130 @@
...
+        required=False,
+        default="WARNING",
+        type=logging.getLevelName,
+        help="Sets the log level",
+    )
+
+    return parser
+
+
+def main_entry_point(args=None):
```
ManInFez: Why does args default to None. It will not result in a happy path to just call `main_entry_point()`?

tests/jobs/rft/test_gendata_rft.py
```diff
@@ -0,0 +1,219 @@
...
+        file2_content = fh.readlines()
+
+    assert len(file1_content) == len(file2_content)
+
+    for line1, line2 in zip(file1_content, file2_content):
+        try:
+            line1, line2 = float(line1), float(line2)
+        except ValueError:
+            continue
+        npt.assert_almost_equal(line1, line2, decimal=7)
```
ManInFez: can we just use  `numpy.testing.asser...`? it us only being used once.

semeio/jobs/scripts/gendata_rft.py
sondreso: Calling parse_args() with `None` will make the function use `sys.argv` (https://docs.python.org/2/library/argparse.html#argparse.ArgumentParser.parse_args)

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,58 @@
...
+            well, time, trajectories[well], ecl_grid, ecl_rft, zonemap
+        )
+        if trajectory_points:
+            fname = "RFT_{}_{}".format(well, report_step)
+
+            _write_points(trajectory_points, fname)
+            _write_inactive_info(trajectory_points, fname, zonemap)
+
+    with open("GENDATA_RFT.OK", "w") as fileH:
+        fileH.write("GENDATA RFT completed OK")
```
oysteoh: details - but would like fileH renamed to file_handler or something ... at least not camelCase

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,58 @@
...
+):
+    try:
+        rft = ecl_rft.get(well, date)
+    except KeyError:
+        logger.error(
+            "Forward model script gendata_rft.py: No RFT data written for well ({}) and date ({})".format(
+                well, date
+            )
+        )
+        return []
```
oysteoh: I'm not sure - but think its more clear if None is returned at this point - and by that the test on line 51 would make more sense - and could also be rewritten to `if not trajectory_points is None` or something.

semeio/jobs/rft/trajectory.py
```diff
@@ -0,0 +1,69 @@
+import sys
+import os
+
+from semeio.jobs.rft.utility import strip_comments
+
```
oysteoh: Thinks all imports are unused

semeio/jobs/rft/trajectory.py
```diff
@@ -0,0 +1,69 @@
...
+        if self.grid_ijk is None:
+            return "TRAJECTORY_POINT_NOT_IN_GRID {}".format(str(self))
+        if self.pressure is None:
+            return "TRAJECTORY_POINT_NOT_IN_RFT {}".format(str(self))
+        if not self.valid_zone:
+            if self.grid_ijk[2] not in zonemap:
+                return "ZONEMAP_MISSING_VALUE {} {} {}".format(
+                    str(self), self.grid_ijk[2], None
+                )
+            else:
```
oysteoh: This else is not necessary

semeio/jobs/rft/trajectory.py
```diff
@@ -0,0 +1,69 @@
...
+            self.valid_zone = zonemap is None or zonemap.has_relationship(
+                self.zone, self.grid_ijk[2]
+            )
+
+    def is_active(self):
+        return (
+            self.grid_ijk is not None and self.pressure is not None and self.valid_zone
+        )
+
+    def inactive_info(self, zonemap):
```
oysteoh: what happens if `grid_ijk is not None` and `pressure is not None` and `valid_zone`. i guess nothing - and - is that ok?

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+                + "Number of zonenames can be 1 or more".format(
+                    line_number=line_number, filename=filename
+                )
+            )
+
+        try:
+            raw_k = int(zonemap_line[0])
+        except ValueError:
+            raise argparse.ArgumentTypeError(
+                basic_err_msg
```
oysteoh: unused arguments in format i guess?

semeio/jobs/config_jobs/GENDATA_RFT
lars-petter-hauge: Thanks, fixed!

semeio/jobs/rft/utility.py
lars-petter-hauge: thanks, fixed!

semeio/jobs/scripts/gendata_rft.py
ManInFez: :+1:

semeio/jobs/rft/utility.py
lars-petter-hauge: When we started this, all the file parser were indeed part of their corresponding classes. So the `load_zonemap` was a class function of `ZoneMap`. My personal opinion is that I don't think the peculiarity of the file formats that exist today should be in any of the classes. So instead I thought it would be better design to have all the parsers in one location, and have the classes be more general.

Agree with my thoughts?

semeio/jobs/rft/utility.py
lars-petter-hauge: I see what you did there ;)

semeio/jobs/rft/utility.py
```diff
@@ -0,0 +1,265 @@
...
+
+    well_time_lines = [(strip_comments(l), i + 1) for i, l in enumerate(lines)]
+
+    for line, line_number in well_time_lines:
+        tokens = line.split()
+
+        if not tokens:
+            continue
+
+        elif len(tokens) != 5:
```
oysteoh: This `elif` could just be an `if`i guess?

semeio/jobs/rft/utility.py
lars-petter-hauge: But the recommended practice from PEP8 is to: 

>For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

```
Yes: if not seq:
     if seq:

No:  if len(seq):
     if not len(seq):
```

semeio/jobs/rft/zonemap.py
```diff
@@ -0,0 +1,31 @@
+import os
```
oysteoh: unused

semeio/jobs/rft/utility.py
lars-petter-hauge: This goes for the same as above, this is the recommended way

semeio/jobs/rft/zonemap.py
```diff
@@ -0,0 +1,31 @@
...
+            for zone_name in zone_names:
+                if not zone_name in self._k_values_at_zone:
+                    self._k_values_at_zone[zone_name] = []
+
+                self._k_values_at_zone[zone_name].append(k_value)
+
+    def __contains__(self, item):
+        if isinstance(item, int):
+            return item in self._zones_at_k_value
+        elif isinstance(item, str):
```
oysteoh: Unnecessary `elif` after `return`

semeio/jobs/rft/zonemap.py
```diff
@@ -0,0 +1,31 @@
...
+            return item in self._k_values_at_zone
+        return False
+
+    def __getitem__(self, item):
+        if isinstance(item, int):
+            return self._zones_at_k_value[item]
+        elif isinstance(item, str):
```
oysteoh: Unnecessary `elif` after `return`

semeio/jobs/rft/utility.py
lars-petter-hauge: It is tested in the caller (`load_zonemap(path)`). The reason it is split is because the `load_zonemap` returns a `ZoneMap` object, while `load_and_parse_zonemap_file` returns the content of the zonemap file in a well defined data structure.

It could perhaps be one function though..

semeio/jobs/rft/utility.py
lars-petter-hauge: Same as PEP above

semeio/jobs/rft/utility.py
lars-petter-hauge: Yepp, fixing!

semeio/jobs/rft/utility.py
lars-petter-hauge: The `EclGrid` and `EclRFT` takes a path, it should be their responsibility to check that the path exists (and they do). In case of non-existing path, they throw an `IOError` which we catch.

tests/jobs/rft/test_gendata_rft.py
lars-petter-hauge: agreed, fixing

semeio/jobs/rft/gendata_rft.py
lars-petter-hauge: Completely agree, not sure how this has managed to creep in!

semeio/jobs/rft/gendata_rft.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for p in trajectory_points:
```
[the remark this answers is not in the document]
lars-petter-hauge: Agree, fixed!

semeio/jobs/rft/gendata_rft.py
lars-petter-hauge: I think I disagree.

I like that a function always returns objects with the same attributes. So a function that usually returns a list of content should return an empty list if nothing happens.

The check on line 51 should be good. It's pythonic to check the emptyness of lists by using `if my_list`.

But I'm happy to be disproved :=)

semeio/jobs/rft/trajectory.py
lars-petter-hauge: jeez, fixing

semeio/jobs/rft/trajectory.py
lars-petter-hauge: Although I agree, I tried to take it away and I must say I didn't feel it looked better.. But, sure, let's remove it.

semeio/jobs/rft/utility.py
lars-petter-hauge: No? The arguments are included in the `basic_err_msg` string, initialized further up. Unused arguments would have thrown an error by the way

semeio/jobs/rft/utility.py
lars-petter-hauge: fixed

semeio/jobs/rft/zonemap.py
lars-petter-hauge: fixed

semeio/jobs/rft/zonemap.py
lars-petter-hauge: You mean it could be `if` instead? Is it _that_ bad to use `elif`?

oysteoh: I'm not entirely sure - for me it doesn't feel right, and for pylint it doesn't either..... `Unnecessary "elif" after "continue" (no-else-continue)`

Don't think it's that bad - feel free to just leave it as is :+1:

semeio/jobs/rft/gendata_rft.py
oysteoh: as said - i'm not sure.. i'm totally fine with the code as is - just wanted to discuss the exception handling and returning an empty list :+1: and you have a good point - it's always good to return objects with same attributes!

semeio/jobs/rft/utility.py
lars-petter-hauge: After some discussions we figured it is better to keep the parsing logic within each classes. Moved functionality

semeio/jobs/rft/utility.py
lars-petter-hauge: This is refactored never the less, and they are now in the same location

semeio/jobs/rft/utility.py
ManInFez: But the error message is hidden. The user does not get the error produced by EclGrid or EclRFTFile. The error message should be propegated to the user.

lars-petter-hauge: I agree with you. The checking of file also happens further down into the c code in ecl. I've added checks for if the file exists (as this is likely one of the most prominent issues and should preferably have a small error message), and I've included the propagated error message from libecl.

Cheers!

PR #19, opened 2020-01-08
add STEA job

xjules:
Moving and updating STEA job from spinning jenny.
Stea can be an open-source since anybody having the stea account can make use of it.
Resolves https://github.com/equinor/ert/issues/597


.travis.yml
```diff
@@ -37,6 +37,7 @@ addons:
       - gcc
 
 install:
+    - pip install git+https://git@github.com/equinor/stea.git
```
lars-petter-hauge: It is possible to add a repository directly to the requirements.txt file, please do that instead.

i.e. `git+git://github.com/equinor/stea` should make it possible to install stea via `pip install -r requirements.txt`

xjules: will do!

tests/jobs/stea/test_stea.py
```diff
@@ -0,0 +1,50 @@
...
+    from mock import patch
+
+from stea import SteaResult, SteaKeys, SteaInput
+import os
+
+TEST_STEA_PATH, _ = os.path.split(os.path.abspath(__file__))
+
+
+def test_import_stea():
+    assert importlib.import_module("stea")
```
lars-petter-hauge: I might be a bit confused, but what is the purpose of this test?

xjules: testing that stea is installed (legacy of previous testing) ... but might also go away, since stea is now part of setup.py dependencies installation.

xjules: removed

semeio/jobs/configs/STEA
```diff
@@ -0,0 +1,4 @@
+EXECUTABLE     ../scripts/fm_stea
```
lars-petter-hauge: I tought I had a comment here, but the `../scripts/fm_stea` should be `../scripts/fm_stea.py`, right?

xjules: right!

semeio/jobs/scripts/fm_stea.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+
+import stea
+import argparse
+import os
+from argparse import ArgumentTypeError
```
lars-petter-hauge: unused imports

tests/jobs/stea/test_stea.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+@patch("stea.calculate")
+@pytest.mark.usefixtures("setup_stea")
+def test_stea(mock_stea):
+    mock_stea.side_effect = calculate_patch
+    main_entry_point(["-c", "stea_input.yml"])
+    mock_stea.assert_called_once()
+    files = os.listdir(os.getcwd())
+    assert "NPV_0" in files
```
lars-petter-hauge: And the reason `NPV_0` should exist in the result files is because in the `yml` config the following is stated:

```
# What do you want stea to calculate
results:
   - NPV
```
??

If so I think this looks good. Thanks for updating!

semeio/jobs/scripts/fm_stea.py
xjules: going away

tests/jobs/stea/test_stea.py
xjules: you think we should mention it as a comment? maybe a good idea

xjules: comment added

PR #20, opened 2020-01-09
Add common utils to semeio

oysteoh:
Blacked files
Adding black tests
Add logging and file checker to semeio init


lars-petter-hauge:
Thanks, good to have this in a single PR, let's get it in! :)

PR #21, opened 2020-01-09
Fix default UPDATE_KEYS with threshold in CALCULATE_KEYS

oyvindeide:


semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -72,7 +72,9 @@ def _realize_list(input_string):
 def _expand_input(input_value):
     expanded_values = deepcopy(input_value)
     if "CALCULATE_KEYS" in expanded_values and "UPDATE_KEYS" not in expanded_values:
-        expanded_values.update({"UPDATE_KEYS": expanded_values["CALCULATE_KEYS"]})
+        expanded_values.update(
```
lars-petter-hauge: Should this not have been caught in the `test_expand_input`?

oyvindeide: Good catch, yes it should, is now.

PR #28, opened 2020-01-15
Overburden timeshift (OTS) job

xjules:
Resolves https://github.com/equinor/ert/issues/596

Part of the job is to replace the config content with a yml config.
It contains a lot of dependencies (e.g., segyio, scipy, etc). Also requires substantial cleaning.

- add config suite build schema
- test ots fixed
- fixed test_ots_reservoir; add conftest
- fixed test_ots_velocity
- add norne ots data
- add configsuite schema
- create_mock_segy and refactor ots_integration test
- add OTS entry point
- add segyio to setup.py
- add ots to hook + black
- fix strpath for py27
- fix contextvalidator


semeio/jobs/ots/ots_res_surface.py
```diff
@@ -0,0 +1,99 @@
...
+    @property
+    def ny(self):
+        return self._ny
+
+    def __len__(self):
+        return len(self.x)
+
+    def __str__(self):
+        s = "x: " + str(self.x)
+        s += "\ny: " + str(self.y)
```
oyvindeide:
```Suggestion:
return = "x: {}\ny: {}\nz: {}".format(x, y, z)``

File: semeio/jobs/ots/ots_res_surface.py [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+        s += "\nz: " + str(self.z)
+        return s
+
+    @property
+    def cell_corners(self):
+        return self._getTopCorners()
+
+    def _calculateSurface(self, grid, above):
```
oyvindeide: Why camelCase?

semeio/jobs/ots/ots_res_surface.py
```diff
@@ -0,0 +1,99 @@
...
+
+        z -= above
+
+        self._x = x
+        self._y = y
+        self._z = z
+        self._nx = nx
+        self._ny = ny
+
+    def _getTopCorners(self):
```
oyvindeide: Why camelCase?

semeio/jobs/ots/ots_vel_surface.py
```diff
@@ -0,0 +1,198 @@
...
+            nt = len(f.samples)
+
+            for i, h in enumerate(f.header):
+                scalar = h[TraceField.SourceGroupScalar]
+                if scalar < 0:
+                    scalar = -1.0 / scalar
+                x[i] = h[TraceField.CDP_X] * scalar
+                y[i] = h[TraceField.CDP_Y] * scalar
+
+            # TODO nx <-> ny??
```
oyvindeide: Is there an issue for this, or is it fixed?

semeio/jobs/ots/ots_vel_surface.py
```diff
@@ -0,0 +1,198 @@
...
+            ny = len(f.xlines)
+            nx = len(f.ilines)
+
+            # TODO memory issue?
+            traces = f.trace.raw[:]
```
oyvindeide: Is there an issue for this, or is it fixed?

semeio/jobs/ots/ots_vel_surface.py
```diff
@@ -0,0 +1,198 @@
...
+        # average extent of the grid in axis direction divided by resolution gives us average extent of one voxel
+        size = np.mean(np.max(res_corners, axis) - np.min(res_corners, axis)) / res_n
+        # the number of grid voxels that can be fit into the segy volume defined by CDP_X and CDP_Y
+        nn = np.ceil(
+            np.mean(np.max(vel_axis, axis) - np.min(vel_axis, axis)) / size
+        ).astype(np.int)
+
+        n = vel_axis.shape[axis]
+        ups = np.floor((n - 1) / nn).astype(np.int)
+        # always only upsscaling
```
oyvindeide: Typo

semeio/jobs/ots/ots_vel_surface.py
```diff
@@ -0,0 +1,198 @@
...
+        # skip samples if needs, ups >= 2
+        traces_upscaled = traces[
+            0 : (nxx - 1) * upsx + 1 : upsx,
+            0 : (nyy - 1) * upsy + 1 : upsy,
+            0 : (ntt - 1) * upst + 1 : upst,
+        ]
+
+        return x, y, traces_upscaled, nt, dt
+
+    def _mapReservoirSurfaceToVelocity(self, res_surface, vcube):
```
oyvindeide: why camelCase?

semeio/jobs/ots/ots_config.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@configsuite.validator_msg("Vintages must contain at least an entry!")
+def _min_length(value):
+    return len(value) > 1
+
+
+@configsuite.transformation_msg("Converting list of strings to list of dates")
+def _str2dates(value):
+    value = deepcopy(value)
```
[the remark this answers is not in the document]
xjules: Due to not being sure whether datetime is passed by value or reference.

semeio/jobs/config_jobs/OTS
```diff
@@ -0,0 +1,4 @@
+UTABLE     ../scripts/ots_job.py
```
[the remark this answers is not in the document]
xjules: yes, fixed now.

semeio/jobs/ots/ots_config.py
```diff
@@ -0,0 +1,148 @@
...
+        if isinstance(x, str):
+            dates.append(datetime.datetime.strptime(x, "%Y-%m-%d").date())
+        else:
+            dates.append(x)
+    return dates
+
+
+@configsuite.validator_msg("OTS vintage date must be present in .UNRST file!")
+def _vintage_present_in_rst(elem, context):
+    if isinstance(elem, datetime.date):
```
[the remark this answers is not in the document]
xjules: in essence, you are right this test can be removed.

semeio/jobs/ots/ots_vel_surface.py
xjules: well, this comment comes from before and IMHO it works it as now.

semeio/jobs/ots/ots_vel_surface.py
xjules: this might become an issue when an entire volume is being read (for extremely large segy files), so left it there. We might get to solve it though.

semeio/jobs/ots/ots_vel_surface.py
xjules: yes, fixed now.

semeio/jobs/ots/ots_res_surface.py
xjules: legacy from ert-statoil, will update all camelCases ....

tests/jobs/ots/test_ots_integration.py
```diff
@@ -0,0 +1,123 @@
+from semeio.jobs.ots import ots_run
+from ecl.util.geometry import Surface
+from ecl.grid import EclGrid
+from ecl.eclfile import EclFile
+import pytest
+from ots_util import mock_segy
```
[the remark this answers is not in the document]
xjules: ok changed

semeio/jobs/ots/ots_vel_surface.py
xjules: I improved comment a bit

semeio/jobs/ots/ots_vel_surface.py
xjules: fixed

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,572 @@
...
+                    ts.append(tshift_ts_simple[iv][point])
+                for iv in range(len(vintage_pairs.dpv)):
+                    ts.append(tshift_dpv[iv][point])
+                for iv, _ in enumerate(vintage_pairs.ts_rporv):
+                    ts.append(tshift_ts_rporv[iv][point])
+                f.write(line.format(xy[0], xy[1], surface_horizon[point], *ts))
+
+
+class OTS(object):
+    def __init__(self, parms, verbose=False):
```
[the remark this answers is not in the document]
xjules: ok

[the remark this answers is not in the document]
xjules: done

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,572 @@
...
+        case = os.path.splitext(parms.eclbase)[0]
+        self._init_file = EclFile("%s.INIT" % case)
+        self._rst_file = EclFile("%s.UNRST" % case)
+        self._grid = EclGrid("%s.EGRID" % case, apply_mapaxes=parms.mapaxes)
+
+        self.subsidence = EclSubsidence(self._grid, self._init_file)
+
+        self._seabed = parms.seabed
+        if not isinstance(parms.youngs, float):
+            parms.youngs = 0
```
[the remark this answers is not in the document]
xjules: yes, it goes away

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,572 @@
...
+        self._youngs_modulus = parms.youngs * 1e9
```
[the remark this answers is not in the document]
xjules: Maybe the value user enters is in different units / scale so I will keep it this way

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,572 @@
...
+        self._poisson_ratio = parms.poisson
+        self._r_factor = parms.rfactor
+        self._convention = parms.convention
+        self._verbose = verbose
+
+        self._surface = res_surface = OTSResSurface(grid=self._grid, above=parms.above)
```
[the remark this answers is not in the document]
xjules: well spotted, legacy of old code :)

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,572 @@
...
+        overburden timeshift calculations.
+        """
+
+        if parms.eclbase is None:
+            raise RuntimeError
+
+        case = os.path.splitext(parms.eclbase)[0]
+        self._init_file = EclFile("%s.INIT" % case)
+        self._rst_file = EclFile("%s.UNRST" % case)
+        self._grid = EclGrid("%s.EGRID" % case, apply_mapaxes=parms.mapaxes)
```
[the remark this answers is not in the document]
xjules: we decided to keep it as is - mainly due to tests

xjules:
test this please!

semeio/jobs/ots/ots_vel_surface.py
```diff
@@ -0,0 +1,198 @@
...
+        # skip samples if needs, ups >= 2
+        traces_upscaled = traces[
+            0 : (nxx - 1) * upsx + 1 : upsx,
+            0 : (nyy - 1) * upsy + 1 : upsy,
+            0 : (ntt - 1) * upst + 1 : upst,
+        ]
+
+        return x, y, traces_upscaled, nt, dt
+
+    def _map_reservoir_surface_to_velocity(self, res_surface, vcube):
```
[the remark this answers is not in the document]
xjules: Will spawn this as a stand alone issue after this has been merged.

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+
+
+def write_surface(vintage_pairs, ts, output_dir, type_str):
+    for iv, vp in enumerate(vintage_pairs):
+        d0 = vp[0].strftime("%Y_%m_%d")
+        d1 = vp[1].strftime("%Y_%m_%d")
+        ts[iv].write(output_dir + type_str + "/ots" + "_" + d0 + "_" + d1 + ".irap")
+
+
+def ots_run(parameter_file, verbose=False):
```
lars-petter-hauge, line 54: There is no way to set the `verbose` to True. The caller of this function only passes the configuration file (which, to be honest, I would think could've been parsed up front).

I believe most jobs in semeio uses builtin logger in python. I would think that having a command line argument where one could set `--debug`, and then instead set the logger level would make this much simpler.

All the `if verbose` checks would also go away

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+    config = configsuite.ConfigSuite(
+        config,
+        build_schema(),
+        deduce_required=True,
+        extract_validation_context=extract_ots_context,
+    )
+    assert config.valid
+
+    input_data = config.snapshot
+    return input_data, input_data.vintages
```
lars-petter-hauge, line 44: it is a bit odd to split the return into both returning the snapshot, and a part of the snapshot. I mean, you're already returning the snapshot..

But otherwise I think it's interesting that the `vintages` turns into `vintage_pairs` on the other side. Are they vintages or vintage pairs?

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+    num_pairs = (
+        len(vintage_pairs.ts)
+        + len(vintage_pairs.ts_simple)
+        + len(vintage_pairs.dpv)
+        + len(vintage_pairs.ts_rporv)
+    )
+    line = "{}, {}, {}" + ", {}" * num_pairs + "\n"
+
+    if parms.ascii is not None:
+        with open(parms.ascii, "w") as f:
```
lars-petter-hauge, line 99: Split this writing out into a separate function as well, as you've done with the surface writing?

Also, I can't help but think that there must be a better way of writing this.. :P 

I'll look at the rest and come back here afterwards for suggestion

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+        yinc = (np.max(y) - ystart) / (ny - 1)
+
+        surf = Surface(
+            nx=nx, ny=ny, xinc=xinc, yinc=yinc, xstart=xstart, ystart=ystart, angle=0
+        )
+
+        irap_x = np.empty(nx * ny)
+        irap_y = np.array(irap_x)
+
+        for i, s in enumerate(surf):
```
lars-petter-hauge, line 190: Why enumerate over `surf` if you don't use it?

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+
+        self._restart_views[name] = restart_view
+
+        return restart_view
+
+    @staticmethod
+    def _divide_negative_shift(ts):
+        for i, _ in enumerate(ts):
+            if ts[i] < 0:
+                ts[i] /= 5.0
```
lars-petter-hauge, line 223: What does this do? What does the number 5 represent? (would've made more sense to have such a parameter an argument)

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+
+    def geertsma_ts_rporv(self, vintage_pairs):
```
lars-petter-hauge, line 225: The `geertsma_ts_rporv`and `geertsma_ts_simple` are identical except for the line `if len(vintage_pairs) < 1:`(0 in one of the function, 1 in the other), and the function they are calling (which could've been a function argument).

I woul recommend that those two functions are merged.

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+                        monitor_survey=None,
+                        pos=r2,
+                        youngs_modulus=self._youngs_modulus,
+                        poisson_ratio=self._poisson_ratio,
+                        seabed=self._seabed,
+                    )
+                    du[iv, point] = dz2 - dz1
+
+        for ivp, vp in enumerate(vintage_pairs):
+            if self._verbose:
```
lars-petter-hauge, line 288: As mentioned, using logger instead would help a bit on this part. And still the amount of logging here kind of scrambles what the function really does. The messages for instance are identical, the thing different is the order of the dates.

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+                        )
+                    )
+                if self._convention == -1:
+                    print(
+                        "{:%x %X} TS_RPORV: Calculating shift"
+                        " {:%Y.%m.%d}-{:%Y.%m.%d} in {} points".format(
+                            dt.now(), vp[1], vp[0], num_points_calculated
+                        )
+                    )
+                if num_points_calculated == 0:
```
lars-petter-hauge, line 303: This variable does not change during the loop, I think. Put the logging outside of the loop instead.

lars-petter-hauge: In fact, it is over-written on each vintage start, and as such will only report on the final vintage pair. Is that really how it should be? This part should probably have been placed on the last line of the previous loop

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+            ts = -self._r_factor * (du[monitor_index] - du[base_index])
+            self._divide_negative_shift(ts)
+
+            ts = ts * self._convention
+
+            ts_surfaces.append(self._create_surface(ts))
+
+        return ts_surfaces
+
+    def geertsma_ts(self, vintage_pairs):
```
lars-petter-hauge, line 411: So this is the function that does the proper time shift evaluation, while the other two are turning some corners. I am not sure if it is as easy to merge them all three into a function, but all three functions are doing the same kind of logging. Splitting the content of the functions a bit will certainly help on that.

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,597 @@
...
+                        num_points_calculated += 1
+                        sum_pv = 0
+                        for k in range(self._grid.getNZ()):
+                            v = self._grid.cell_volume(ijk=(i, j, k))
+                            sum_pv += pressure[i, j, k] * v
+
+                        pv[iv, point] = sum_pv
+                    except ValueError:
+                        pv[iv, point] = 0
+        for ivp, vp in enumerate(vintage_pairs):
```
lars-petter-hauge, line 569: The fourth time I've seen the same logging now.. Split the functionality and have each of the other functions call it.

semeio/jobs/ots/ots_config.py
```diff
@@ -0,0 +1,195 @@
...
+            },
+            "above": {
+                MK.Type: types.Number,
+                MK.Description: "Distance in meters above the reservoir"
+                " where shift is calculated. The distance from the "
+                "shallowest cell, eg. 100",
+            },
+            "convention": {
+                MK.Type: types.Number,
+                MK.Description: "Positive or negative shift can be either 1 or -1,"
```
lars-petter-hauge, line 59: If it can be only 1 or -1, I think we should add an element validator here.

lars-petter-hauge:
Looks a lot better than it did, especially the testing has been increased!

I've added a couple comments though, as I still think we can improve parts of it.

Otherwise there's a fair amount of single letter, two letter variables in this code. Although I usually find out what they mean close by, it is a bit of additional work to always have to deduce what they mean (especially if you haven't looked at the code for a while).

semeio/jobs/ots/ots_config.py
xjules: Sounds like a good idea!

semeio/jobs/scripts/ots.py
```diff
@@ -0,0 +1,34 @@
...
+    "based on eclipse models and seismic velocity volume."
+    "Input yml needs to contains vintages section, where at"
+    " least one of the four categories"
+    "for the surface computation needs to be set."
+)
+
+category = "modeling.surface"
+
+
+def _get_args_parser():
```
lars-petter-hauge, line 18: There's a good deal of information in the configsuite schema. It is possible to create a RST output of the content with `configsuite.create_rst` or something similar. Could you add an argument to the job that will output this information? Or add it as part of the description? Part of the description if probably nice, as this will be included on the docs pages.

semeio/jobs/ots/ots.py
xjules: Well spotted also I agree with having a logger instead. Will fix it.

semeio/jobs/ots/ots.py
xjules: Yes, 4 times as it has 4 methods for computing the surface shift. Anyway, as you suggested it'll be base logging now only.

semeio/jobs/ots/ots.py
xjules: I see your point, but prefer to have it split. Especially if logging message is slightly different for each of the four methods. Splitting them would bring some additional ifs and more parameters for the functions. What can be done is to group logging code into semantically same-y methods.

semeio/jobs/ots/ots.py
xjules: Was bothering my head with this one too. Not really sure, how `num_points_calculated` relates to each vintage pair - as you can see it's reported with each vintage pair. and as such is should be num_points_calcilated for each of the vintage pair (since it's overwritten in current implementation) - So left it like in original, but might be that the answer is like this:
```python
num_points_calculated[vp] 
```

semeio/jobs/ots/ots.py
xjules: The logging (as mentioned bellow) can be extracted to helper methods.

semeio/jobs/ots/ots.py
xjules: If I would know more about details than perhaps it might be a good idea, but since the extracted description that is in CS schema says something I'm leaving it as is.

semeio/jobs/ots/ots.py
xjules: Pretty it does what it says, but why 5 is a good question :) Can do it as parameter defaulting to 5

semeio/jobs/ots/ots.py
xjules: Flake8's complaining when terating over indices, lists, etc. without enumeration, but this concrete case is from older ert-statoil times, so `s` can redundant.

semeio/jobs/scripts/ots.py
xjules: I would say that having a specific parameter for creating the rst file sounds like a very good idea!

semeio/jobs/ots/ots.py
xjules: So I read a bit on this and it's even more ambiguous than before. Firstly, `num_points_calculated` is computed for each date specified, ie. when having 4 pairs of dates (base + monitor) this will generate 4 surface calculations with 4 distinct  `num_points_calculated`. But after that we do this:
```python
for ivp, vp in enumerate(vintage_pairs):
            if self._convention == 1:
                logging.info(
                    "{:%x %X} TS_SIMPLE: Calculating shift"
                    " {:%Y.%m.%d}-{:%Y.%m.%d} in {} points".format(
                        dt.now(), vp[0], vp[1], num_points_calculated)
)
```
Ie. we report `num_points_calculated` for each pair of surfaces. I need to check what it actually means.

semeio/jobs/ots/ots.py
lars-petter-hauge: What? So if you do:

`for value in mylist:` that would be a flaw?! That makes no sense at all

Or do you mean the `for i in range(len(surf)):` would be an error?

semeio/jobs/ots/ots.py
lars-petter-hauge: I mean, this is such a special thing to do.. I think we at least should state it in the documentation. It isn't nullified, it's just divided by the number 5.

semeio/jobs/ots/ots.py
lars-petter-hauge: The implementation should really not be effected by what the documentation says.

To be honest, I think it makes things much more transparent if we were to merge the functions, because then the differences between the implementations would be obvious. Instead of how it is now, where we need to diff the entire function to see what's different.

semeio/jobs/ots/ots.py
lars-petter-hauge: I think this is a copy+paste issue. The logging was placed correctly in the original implementation (betting `geertsma_ts` - in which this looks to make sense), and then someone wanted the same logging in a different function and just copy pasted the entire thing, not seeing how it got wrong.

semeio/jobs/ots/ots.py
lars-petter-hauge: Something like:

```
    def _report(self, func_name, base, monitor, num_points_calculated):
        if self._convention == 1:
            start_date, end_date = monitor, base
        elif self._convention == -1:
            start_date, end_date = base, monitor
        else:
            raise ValueError("Convention must be 1 or -1, was {}".format(self._convention))
        logging.DEBUG(
                "{:%x %X} TS: Calculating shift"
                " {:%Y.%m.%d}-{:%Y.%m.%d} in {} points".format(
                    dt.now(), start_date, end_date, num_points_calculated
                )
            )
```
And have that called in each of the four places.

semeio/jobs/ots/ots.py
lars-petter-hauge: Made a suggestion above, and then we can

```
for ivp, vp in enumerate(vintage_pairs):
    self._report("function name", vp[0], vp[1], num_points_calculated)
``

File: semeio/jobs/ots/ots.py (hunk 23 above)
Comment (lars-petter-hauge) [reply semeio/jobs/ots/ots.py:222]: > The geertsma_ts_rporvand geertsma_ts_simple are identical except for the line `if len(vintage_pairs) < 1`:(0 in one of the function, 1 in the other), and the function they are calling (which could've been a function argument).

A small brainfart here.. it's doing:

```
if len(vintage_pairs) < 1:
    return 0, []
```

File: semeio/jobs/ots/ots.py (hunk 19 above)
Comment (xjules) [reply semeio/jobs/ots/ots.py:43]: Those are vintage pairs to start with, but agree this can just output `input_data` as I did this for old-sake compatibility :)

File: semeio/jobs/ots/ots.py (hunk 20 above)
Comment (xjules) [reply semeio/jobs/ots/ots.py:97]: Agree, there could be a better way to writing this to a file, but I think this would require also a better data structure in the first place. So I left it as is currently, let's get back to it.

File: semeio/jobs/ots/ots.py (hunk 21 above)
Comment (xjules) [reply semeio/jobs/ots/ots.py]: I changed it to `range(len(surf))` as we agreed that it performs faster.

File: semeio/jobs/ots/ots.py (hunk 23 above)
Comment (xjules) [reply semeio/jobs/ots/ots.py:221]: For the sake of wider coverage I took
```python
if len(vintage_pairs) < 1:
    return 0, []
```

semeio/jobs/ots/ots.py
xjules: It's done via helper function now with `num_points_calculated` brought up before the main loop.

semeio/jobs/ots/ots.py
lars-petter-hauge: A suggestion at least.. I think I got most of it correct, haven't run it or anything. And even still the default arguments of the `delta_function` could be included higher. I am not completely sure how sane it is to do the partial on those functions, but I think I favor it. If we do it even further up we could make it even more obvious, that the defaults for poisson_ratio etc is the same all over. It's really just the survey names and position that changes.

Would want a better name for `du` as well. To be honest not sure what to call it.

```
    def delta_surface_calculation(self, vintage_pairs, delta_function=self.subsidence.evalGeertsma):
        """
        Must have good doc string here!
        """

        if len(vintage_pairs) == 0:
            # Always return same type
            return np.zeros((0, 0))

        vintages = self._vintages_name_date(vintage_pairs)
        surface = self._surface
        idx_to_calculate = [i for i in len(surface) if not np.isnan(surface.z[i])]

        du = np.zeros((len(vintages.name), len(surface)))
        ts_surfaces = []
        for vintage_index, (vintage_name, vintage_date) in enumerate(zip(vintages.name, vintages.date)):
            self.add_survey(vintage_name, vintage_date)

            dz_func = partial(delta_function(
                    base_survey=vintage_name,
                    monitor_survey=None,
                    youngs_modulus=self._youngs_modulus,
                    poisson_ratio=self._poisson_ratio,
                    seabed=self._seabed,
                )

            for idx in idx_to_calculate:
                r1 = (surface.x[idx], surface.y[idx], 0)
                r2 = (
                    surface.x[idx],
                    surface.y[idx],
                    surface.z[idx] - self._seabed,
                )
                # NB: comment below from pre refactor, not sure if this is well formulated

                # subsidence and displacement have opposite sign
                # should have minus on dz1 and dz2 here, more efficient
                # when calculating ts
                du[vintage_index, idx] = dz_func(pos=r1) - dz_func(pos=r2)

        for (base, monitor) in vintage_pairs:
            self._report(delta_function.__name__, base, monitor)

            base_index = vintages.date.index(base)
            monitor_index = vintages.date.index(monitor)
            ts = -self._r_factor * (du[monitor_index] - du[base_index])
            self._divide_negative_shift(ts)

            ts = ts * self._convention

            ts_surfaces.append(self._create_surface(ts))

        return ts_surfaces
```

semeio/jobs/ots/ots.py
xjules: The question's how important is the division factor and whether it should be exposed as a parameter.

semeio/jobs/ots/ots.py
xjules: Agree, worked in already!

semeio/jobs/scripts/ots.py
xjules: Currently I made it only that `configsuite.docs.generate(schema)` appends the string to the description. Do you think it suffice?

semeio/jobs/ots/ots.py
xjules: Wrote to an author of this some days back, but still no answer. Shall we just have it as is then?

semeio/jobs/ots/ots.py
xjules: The usage of the subsidence function is so specific that I don't think having partial would bring any benefits neither the defaults for the matter. I renamed `du` to `surf_displacement` though.

semeio/jobs/scripts/ots.py
xjules: In the end I had decided for an extra parameter `--params` just showing the RST output from configsuite.

semeio/jobs/scripts/ots.py
```diff
@@ -0,0 +1,48 @@
...
+
+from semeio.jobs.ots.ots import ots_run
+from semeio.jobs.ots.ots_config import generate_rst_doc
+import argparse
+from semeio import valid_file
+
+description = (
+    "Overburden timeshift (OTS) generates evolution of reservoir surfaces"
+    "based on eclipse models and seismic velocity volume. If the shift"
+    "becomes negative it won't be nullified but only divided"
```
oyvindeide, line 11: Missing space between shift and becomes

semeio/jobs/scripts/ots.py
oyvindeide: This should also be included in the description, as the top level `description` will be compiled as documentation for `ert` and that way it will be much more accessible for the user.

tests/jobs/ots/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+
+# use this fixture for setting up norne data and entering
+# another folder while having handle to norne data dir
+@pytest.fixture
+def ots_tmpdir_enter(_setup_ots_tmppath, tmpdir):
+    tmp_path = _setup_ots_tmppath
+    cwd = os.getcwd()
+    tmpdir.chdir()
+    yield tmp_path
+    os.chdir(cwd)
```
oyvindeide, line 26: This can be simplified to:
```suggestion
with tmpdir.as_cwd() as cwd:
    yield tmp_path
```

semeio/jobs/scripts/ots.py
oyvindeide: And a few other lines come to think of it

semeio/jobs/ots/ots.py
oyvindeide: A bit hard to follow the discussion here, but what is the added benefit of calling `geertsma_ts_custom` from two different functions? From my point of view it is moving knowledge down in the call stack instead of up. I would rather split up `_gertsma_ts_custom` (which has a typo in the doc string, subsudence).

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+
+        if len(vintage_pairs) < 1:
+            return 0, []
+
+        vintages = self._vintages_name_date(vintage_pairs)
+        surface = self._surface
+        points_to_calculate = self._get_non_nan_points()
+        surf_displacement = np.zeros((len(vintages.name), len(surface)))
+        ts_surfaces = []
+        for iv, (vn, vd) in enumerate(zip(vintages.name, vintages.date)):
```
oyvindeide, line 263: Can you rename `iv`, `vn` and `vd`, better to just call them `vintages_name` and `vintages_date`, or just `name` and `date`?

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+                    base_survey=vn,
+                    monitor_survey=None,
+                    pos=r2,
+                    youngs_modulus=self._youngs_modulus,
+                    poisson_ratio=self._poisson_ratio,
+                    seabed=self._seabed,
+                )
+                surf_displacement[iv, point] = dz2 - dz1
+
+        for ivp, vp in enumerate(vintage_pairs):
```
oyvindeide, line 304: Also here, can I have more descriptive variable names?

semeio/jobs/scripts/ots.py
xjules: Might be, as wasn't really sure if the space is automatically added or not -  so apparently not :)

tests/jobs/ots/conftest.py
xjules: didn't know about the context support for tmpdir; will do!

semeio/jobs/ots/ots.py
xjules: I thought about the following: `idx_vtg`, `name_vtg`, `date_vtg`

semeio/jobs/ots/ots.py
xjules: Done

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+                        dz2 = self.subsidence.evalGeertsma(
+                            base_survey="base",
+                            monitor_survey="monitor",
+                            pos=r2,
+                            youngs_modulus=self._youngs_modulus,
+                            poisson_ratio=self._poisson_ratio,
+                            seabed=self._seabed,
+                        )
+
+                        ezz = (dz2 - dz1) * 10
```
oyvindeide, line 372: What is `ezz`, and why is it multiplied by 10?

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+                )
+                dz2 = subsidence_func(
+                    base_survey=name_vtg,
+                    monitor_survey=None,
+                    pos=r2,
+                    youngs_modulus=self._youngs_modulus,
+                    poisson_ratio=self._poisson_ratio,
+                    seabed=self._seabed,
+                )
+                surf_displacement[idx_vtg, point] = dz2 - dz1
```
oyvindeide, line 302: Would it make sense to have this as a dict instead? Looking up on the index seems a bit error prone.

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+            base_index = vintages.date.index(pair[0])
+            monitor_index = vintages.date.index(pair[1])
+            ts = -self._r_factor * (
+                surf_displacement[monitor_index] - surf_displacement[base_index]
+            )
+            self._divide_negative_shift(ts)
+
+            ts = ts * self._convention
+
+            ts_surfaces.append(self._create_surface(ts))
```
oyvindeide, line 316: Maybe this should be its own function? That way we can add unit tests (and I would more easily understand what is happening)

semeio/jobs/ots/ots.py
oyvindeide, line 302: Should perhaps be a separate function?

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+        ts_surfaces = []
+        for idx_vtg, (name_vtg, date_vtg) in enumerate(zip(vintages.name, vintages.date)):
+            try:
+                logging.info(
+                    "{:%x %X} {}: Calculating vintage {:%Y.%m.%d}".format(
+                        dt.now(), method_name, date_vtg
+                    )
+                )
+            except TypeError:
+                logging.exception("Wrong date vintage format!")
```
oyvindeide, line 271: This message should be improved to help debugging, at least it should include the previous error to show where there is an error.

oyvindeide: Or - couldnt we validate the vintage data format earlier, to avoid this problem entirely?

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+
+        :param vintage_pairs:
+        :param subsidence_func: specify subsidence method to be used
+        :param method_name: string represeting the subsudence func name
+        """
+
+        if len(vintage_pairs) < 1:
+            return 0, []
+
+        vintages = self._vintages_name_date(vintage_pairs)
```
lars-petter-hauge, line 258: From what I can see, what we are interested in is the unique list of vintages, and then we iterate over them. If you include the change to use a dict, and just lookup on the date/name, as suggested by @oyvindeide, you could instead have the `_vintages_name_date` return a list of Vintages e.g.:

```
        Vintage = namedtuple("Vintage", "name date")
        return [Vintage(name, date) for name, date in zip(vintages_name, vintages_date)]
```

And then in `_geerstma_ts_custom`:
```
for vintage in vintages:
    self.add_survey(vintage.name, vintage.date)
    ....
    surf_displacement[vintage.name][point] = dz2 - dz1
```

Not sure if it is even necessary with the names, which are basically "S0", "S1" etc...

semeio/jobs/ots/ots.py
xjules: I've checked the code of `ecl_grav_common_eval_geertsma_kernel` (being used by `subsidence.evalGeertsma`), which is essentially a convolution (over voxel grid of velocities) by applying `ecl_grav_common_eval_geertsma_kernel`. This ends up being weighted sum velocities; therefore it looks like conversion from speed (or its difference) into distance, where `time=10`, but no idea about the units (years? :)). I've contacted the original author about some other weird constants, like the negative shift division, and no reply yet. Wondering if this should be attached.

semeio/jobs/ots/ots.py
xjules: I thought about indexing it (some time back that is), but I consider numpy a quite efficient structure (performance and storage wise), so I think it is ok to use it.

lars-petter-hauge: But a dict consisting of strings as keys and numpy arrays as values should be equally efficient though, right? We won't be looking up a high number of vintages anyway.

The downside of what is now is that it is necessary to do a bunch of checking to get the correct index:
```
for pair in vintage_pairs:
    base_index = vintages.date.index(pair[0])
    monitor_index = vintages.date.index(pair[1])
    ts = -self._r_factor * (
         surf_displacement[monitor_index] - surf_displacement[base_index]
    )
```

instead of just going directly to
```
for base, monitor in vintage_pairs:
    ts = -self._r_factor * (
         surf_displacement[monitor] - surf_displacement[base]
    )
```

semeio/jobs/ots/ots.py
xjules: It computes the actual displacement surface `surf=scale*(monitor-base)`, which contains the unfortunate negative shift division. Could you maybe clarify how would this benefit from being in a stand-alone function? There's no other use of this function, ie. it is integral part of this displacement calculation. Also it involves only linear transformations - so I would need to be more convinced to do a unit test for this one ;)

semeio/jobs/ots/ots.py
xjules: Same as above, ie. don't see any benefit of having this as separate function. Besides, before line `263` there is only initialization of variables. Not saying that I cannot be convinced though :)

semeio/jobs/ots/ots.py
xjules: There's a test, which injects wrong (or None) dates into ots calculations directly:
https://github.com/xjules/semeio/blob/b16b75a8fd8905bac2fe5157657a6bb3c5b1e011/tests/jobs/ots/test_ots.py#L189

It expects that `ValueError` is risen, but since this won't be used through API (or at least that's what I reckon) that test can go away. What do you think?

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+                    poisson_ratio=self._poisson_ratio,
+                    seabed=self._seabed,
+                )
+                surf_displacement[idx_vtg, point] = dz2 - dz1
+
+        for pair in vintage_pairs:
+            self._report(method_name, pair[0], pair[1], len(points_to_calculate))
+
+            base_index = vintages.date.index(pair[0])
+            monitor_index = vintages.date.index(pair[1])
```
lars-petter-hauge, line 308: Out of curiosity, the content of the vintage pairs at this point is that the first item is the base, while the second is the monitor. However in `geertsma_ts` it is the opposite?

```
self.add_survey("base", pair[1])
self.add_survey("monitor", pair[0])
```

I mean, one would expect such a mistake, if it really is, to have been caught a long time ago.. But it is without a doubt weird that we are not handling the vintages consistently.

semeio/jobs/ots/ots.py
```diff
@@ -0,0 +1,477 @@
...
+        logging.debug(
+            "{:%x %X} {}: Calculating shift"
+            " {:%Y.%m.%d}-{:%Y.%m.%d} in {} points".format(
+                dt.now(), func_name, start_date, end_date, num_points_calculated
+            )
+        )
+
+    def _get_non_nan_points(self):
+        points_to_calculate = [
+            id for id in range(len(self._surface)) if not np.isnan(self._surface.z[id])
```
lars-petter-hauge, line 419: Overriding built-in function https://docs.python.org/3/library/functions.html#id

Although it will be locally it is bad behaviour..

semeio/jobs/ots/ots.py
xjules: Ok, must say that I misunderstood what to replace. This makes sense - can work it in I think.

semeio/jobs/ots/ots.py
lars-petter-hauge: I have to say, I still don't understand what `ezz` is supposed to be an acronym of.. But, yeah, I agree to include a question about this as well. Could be something simple, but nice to know (and likely document)

semeio/jobs/ots/ots.py
xjules: Good point! Had no idea that `id` is taken.

semeio/jobs/ots/ots.py
xjules: Exactly my point as well already added to questions!

semeio/jobs/ots/ots.py
xjules: Appending the answer from Thomas:

Observations (even though they are few) indicate that time shifts linked with compaction are smaller than when linked with stretch. This is also reported by e.g. Hatchel and Bourne (2005), which obtained that `R_stretch / R_comp` is around 5 (where R_stretch denotes R-factor for stretch and R_comp denotes the R-factor for compaction). Hatchell and Bourne obtained R_stretch=5 for Valhall field and we (Røste and Ke, 2017) obtained R_stretch=20 for sandstone fields, but we both obtained that `R_stretch / R_comp ≈ 5.` This is why we have chosen `R_comp = R_stretch / 5.`
 
The physical reason for this is not fully understood, but we believe rocks undergoing stretch changes differently than rocks undergoing compaction.

semeio/jobs/ots/ots.py
xjules: Appending an answer from Thomas:

Reversed computation for the dates fo base and monitor might be connected to the fact that Snorre uses reversed convention (and Snorre where first user of this workflow).

semeio/jobs/ots/ots.py
xjules: Appending answer from Thomas (quite different from what I thought):

ezz denotes the strain in the vertical direction, obtained from the geomechanical model. The vertical strain ezz is equal to dz/z, where dz is the thickness change within a chosen thickness z. 

Regarding the equation,

`ezz = (dz2 - dz1) * 10,`

I do not remember exactly, but by interpretation I guess that the thickness z is chosen to be 1/10 (given in meter, if dz is given in meter); which means that the strain is calculated every 0.1 m. This seems reasonable.

semeio/jobs/ots/ots.py
oyvindeide: Yes, that seems reasonable, as long as the dates are validated that should be no problem.

xjules: Yes, mainly if they are present in `ecl.UNRST` file

semeio/jobs/ots/ots.py
lars-petter-hauge: So we are definitively changing this one then. I'm a bit confused as to why there's a `convention` parameter, and still this exists. But perhaps the change was added and the parameter followed afterwards without reverting how the function runs..

Never the less, let's keep consistent! :)

semeio/jobs/ots/ots.py
lars-petter-hauge: (I got a bit help from my IDE to be honest :P )

semeio/jobs/ots/ots.py
lars-petter-hauge: I would find it a bit odd that anything that has to do with resolution of either models is burried so far down into the code. Grid cells in eclipse can be small, so I don't think that limitation exists in libecl. I'm not sure if there's a bound limit in segy file format though, but I doubt it.

I seems reasonable to calculate it every n meter (but I would think every 10 cm is too low though, for a 300m depth reservoir, that would accumulate to 300 000 points in z direction), just don't understand how it just happens to be included here.

lars-petter-hauge: Considering

> to the fact that Snorre uses reversed convention (and Snorre where first user of this workflow).

Just to make sure, this doesn't have anything to do with Snorre. We don't want field specific quantities in the code.

semeio/jobs/ots/ots.py
xjules: Still waiting for the reply from Thomas to check whether `convention` alone suffices for the reversed case.

semeio/jobs/ots/ots.py
xjules: It's been refactored now.

semeio/jobs/ots/ots.py
xjules: I've gotten the reply, which states that the implementation should be correct and as such it's indeed reversed in `geertsma_ts` job.

lars-petter-hauge: Alright.. Could be good with a comment, but I don't have anything better than
# Reversed because reasons

:P

semeio/jobs/ots/ots.py
xjules: This exception is gone as well as the (now unnecessary) test

semeio/jobs/ots/ots.py
xjules: Done as suggested by @lars-petter-hauge

semeio/jobs/ots/ots.py
xjules: > It seems reasonable to calculate it every n meter (but I would think every 10 cm is too low though, for a 300m depth reservoir, that would accumulate to 300 000 points in z direction - or am I messing something up here?), but I would expect it to depend on the resolutions of the grid cell and the segyfile, rather than a fixed number

Unless we read the paper we won't know well enough about the thickness parameter I reckon. To me it looks more like a constant, which be declared in the beginning of the file, but I wouldn't expose it as a free parameter.

xjules:
Test this please!

semeio/jobs/ots/ots.py
lars-petter-hauge: I don't really know what else to do, but to leave it as is.

semeio/jobs/ots/ots.py
lars-petter-hauge: I believe the current implementation now suffices for what I suggested at least.

semeio/jobs/ots/ots.py
halatequinor: What Thomas is saying is certainly true, but ideally the number 5 should be a parameter that could be input (through the ots_parameters file). For instance on Kvitebjørn we definitely see the "compaction halo" surrounding the reservoir, and by tuning the R-factor for streching, we could improve the match with 4D data. That could be important of fields with smaller compartments, like Valemon.

markusdregi:
@xjules Can you write up the remaining discussions in issues? 👍 

Also, the Jenkins pipeline is failing... ⬇️

lars-petter-hauge (approved):
Good job!

semeio/jobs/ots/ots.py
lars-petter-hauge: https://github.com/equinor/semeio/issues/190

semeio/jobs/ots/ots.py
lars-petter-hauge: https://github.com/equinor/semeio/issues/189

markusdregi:
🎉

PR #53, opened 2020-02-26
Write forward model for pyscal

berland:
Tested manually with minimal ert config file.

I would like comments on how ugly it is to have the forward model read numbers from `parameters.txt`.

Requires pyscal 0.4.0


ertomatic:

semeio/jobs/config_jobs/PYSCAL
```diff
@@ -0,0 +1,23 @@
+STDERR PYSCAL.stderr
+STDOUT PYSCAL.stdout
+
+TARGET_FILE PYSCAL.OK
```
markusdregi: Do we really need a target file for this job? Are we unable to detect failures and just exit with a non-zero exit code?

semeio/jobs/config_jobs/PYSCAL
```diff
@@ -0,0 +1,23 @@
+STDERR PYSCAL.stderr
```
markusdregi: This is defaulted by ERT and unnecessary.

semeio/jobs/config_jobs/PYSCAL
```diff
@@ -0,0 +1,23 @@
...
+STDOUT PYSCAL.stdout
+
+TARGET_FILE PYSCAL.OK
+
+PORTABLE_EXE ../scripts/fm_pyscal.py
```
markusdregi: I suggest using EXECUTABLE instead ;)

semeio/jobs/config_jobs/PYSCAL
```diff
@@ -0,0 +1,23 @@
...
+DEFAULT <INT_PARAM_WO_NAME> __NOINTERPOLATION__
+DEFAULT <INT_PARAM_GO_NAME> __NOINTERPOLATION__
+DEFAULT <SHEET_NAME> 0
+DEFAULT <FAMILY> 1
+DEFAULT <SLGOF> SGOF
+ARGLIST <PARAMETER_FILE> <RESULT_FILE> <SHEET_NAME> <INT_PARAM_WO_NAME> <INT_PARAM_GO_NAME> <FAMILY> <SLGOF>
+
+MIN_ARG    1
+MAX_ARG    6
+ARG_TYPE 0 STRING
```
markusdregi: The type checking does not really work and will not be fixed in its current format. Keep it as if if you like, but if it where me I would just remove it ;)

markusdregi:
> I would like comments on how ugly it is to have the forward model read numbers from parameters.txt.

Any reason why you are not reading the `parameters.json` instead? :)

markusdregi:
I think it is perfectly natural for a forward model job to read from the `parameters`-files. This is the input to the forward model :)

berland:
> Any reason why you are not reading the `parameters.json` instead? :)

Due to issue #54

markusdregi:
I had my suspicions ;)

jondequinor:
We broke the spearman job test. If you rebase on master, that should be fixed.

berland:
Ready for final(?) review.

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,162 @@
...
+import os
+import six
+
+from semeio.jobs.design_kw.design_kw import extract_key_value
+
+from pyscal import pyscalcli
+
+_logger = logging.getLogger("FM_PYSCAL")
+
+MAGIC_NONE = "__NONE__"
```
jondequinor: What is `MAGIC_NONE`? If this is some well-known thing from pyscal, could you add that as a comment?

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def main():
+    """As a forward model, all arguments are mandatory, but can have
+    defaults inserted"""
+    run(
```
jondequinor: Please use argparse here. Move defaults from `run` into the argparse and add help/semantics to each parameter.

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,162 @@
...
+import logging
+import sys
+import os
+import six
+
+from semeio.jobs.design_kw.design_kw import extract_key_value
+
+from pyscal import pyscalcli
+
+_logger = logging.getLogger("FM_PYSCAL")
```
jondequinor: What is this logger? If the logger is not defined, will it default to standard streams?

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,162 @@
...
+    parameters_file_name="parameters.txt",
+):
+    """This function is a wrapper around the Pyscal command
+    line tool. The command line tool is designed around argparse
+    and this function wraps around that design.
+
+    In contrast with the command line tool where interpolation
+    parameters are explicit, they are implicit here, and gathered
+    from parameters.txt"""
+    if not isinstance(relperm_parameters_file, six.string_types):
```
jondequinor: This, and the following, checks should go into an argparse validator.

tests/jobs/pyscal/test_pyscal.py
```diff
@@ -0,0 +1,205 @@
...
+        "",
+        "RELPERM:INTERPOLATE_WO",
+        "INTERPOLATE_GO",
+        "sgof",
+        1,
+    )
+    assert os.path.exists("relperm.inc")
+
+
+def test_fm_pyscal_errors(tmpdir):
```
jondequinor: These tests should be parametrized, meaning that the test is run multiple times with varying input. The varying input here would be e.g. `expected_exceptions`, `run_args`, `static_dframe`, `indexed_csv`, etc.

Comments on the assertions/setups can be in the parameter data.

tests/jobs/pyscal/test_pyscal.py
```diff
@@ -0,0 +1,205 @@
...
+    """Test fm_pyscal on a static xlsx file"""
+    tmpdir.chdir()
+    EXAMPLE_STATIC_DFRAME.to_excel("relperm-input.xlsx")
+    run(
+        "relperm-input.xlsx", "relperm.inc", "", "__NONE__", "__NONE__", "slgof", 1,
+    )
+    assert os.path.exists("relperm.inc")
+
+
+def test_fm_pyscal_scal(tmpdir):
```
jondequinor: These tests should be parametrized.

tests/jobs/pyscal/test_pyscal.py
```diff
@@ -0,0 +1,205 @@
...
+
+import os
+
+import pytest
+
+import pandas as pd
+
+from semeio.jobs.scripts.fm_pyscal import run, rm_genkw_prefix
+
+TEST_PYSCAL_PATH, _ = os.path.split(os.path.abspath(__file__))
```
jondequinor: Is this unused? If so you can just remove it.

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,162 @@
...
+        # Let the WO interpolation be default for GasOil:
+        do_go_interpolation = True
+        int_param_go_name = int_param_wo_name
+
+    if int_param_go_name != MAGIC_NONE:
+        int_param_go_name = rm_genkw_prefix(int_param_go_name)
+        do_go_interpolation = True
+        if not do_wo_interpolation:
+            _logger.error("WaterOil interpolation parameter missing")
+            sys.exit(1)
```
jondequinor: This is pretty dense and hard to follow. I would strongly consider implementing a small DSL here.

semeio/jobs/scripts/fm_pyscal.py
berland: This line was stolen from `fm_stea.py`, except for the explicit naming. It ends in stderr by default (so it seems).

semeio/jobs/scripts/fm_pyscal.py
berland: Done

semeio/jobs/scripts/fm_pyscal.py
berland: Done

semeio/jobs/scripts/fm_pyscal.py
berland: Documented.

tests/jobs/pyscal/test_pyscal.py
berland: Removed

semeio/jobs/scripts/fm_pyscal.py
berland: Rewritten, hopefully for the better.

tests/jobs/pyscal/test_pyscal.py
berland: Done.

tests/jobs/pyscal/test_pyscal.py
berland: Done.

jondequinor:
I am happy with the changes you made. I'll let @markusdregi do the final review.

berland:
Possible feature request to PR: Allow explicit interpolation numbers as forward model arguments. This will allow setting base case from the ert config file (override design matrix parameter), and will allow testing a SCAL xlsx file without intervening in the design matrix.

The parameter INT_PARAM_WO_name could possibly be reused as both a paramter name and a value, renamed to INT_PARAM_WO. If value is a float between -1 and 1, it is interpreted as an explicit value, if it is not a float, assume it is a string with a parameter name.

setup.py
```diff
@@ -51,13 +51,15 @@ def package_files(directory):
...
+        "six>=1.12.0",
         "scipy",
         "xlrd",
         "stea",
+        "pyscal>=0.4.0",
     ],
     setup_requires=["pytest-runner", "setuptools_scm"],
-    tests_require=["pytest", "mock", 'black; python_version>="3.6"'],
+    tests_require=["pytest", "mock", "openpyxl<=2.6.4", 'black; python_version>="3.6"'],
+    # The openpyxl restriction is relevant as long as Python2.7 is to be supported.
```
markusdregi: I suggest we rephrase the `tests_require` as `'openpyxl<=2.6.4; python_version<"3"'` and remove the comment then ;) Better in code than in comments...

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,201 @@
...
+            output=output_filename,
+            sheet_name=sheet_name,
+            int_param_wo=int_param_wo,
+            int_param_go=int_param_go,
+            slgof=slgof == "slgof",
+            family2=family == 2,
+        )
+    except ValueError as e_msg:
+        _logger.error(str(e_msg))
+        sys.exit(1)
```
markusdregi: I would suggest that you do `raise e_msg` instead. Then you still get the message of the logger on `stdout`, but in `stderr` you will still get the original stack trace. Which is extremely useful if you need to debug something :)

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,201 @@
...
+    """Remove anything before the first colon in a string, or remove
+    the same before every key in a dict"""
+    if isinstance(str_or_dict, six.string_types):
+        if ":" in str_or_dict:
+            parts = str_or_dict.split(":")
+            return ":".join(parts[1:])
+        return str_or_dict
+    if isinstance(str_or_dict, dict):
+        return {rm_genkw_prefix(key): value for key, value in str_or_dict.items()}
+    raise TypeError
```
markusdregi: Perhaps a message to follow the exception would be nice?

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,201 @@
...
+    parser.add_argument("output_filename", type=str)
+    parser.add_argument("sheet_name", type=str)
+    parser.add_argument("int_param_wo_name", type=str)
+    parser.add_argument("int_param_go_name", type=str)
+    parser.add_argument("slgof", type=str)
+    parser.add_argument("family", type=int)
+    return parser
+
+
+def run(
```
markusdregi: Can we split this function into multiple private functions? It is a tad long ;)

semeio/jobs/scripts/fm_pyscal.py
```diff
@@ -0,0 +1,201 @@
...
+def _get_args_parser():
+    """Construct an argparse parser for fm_pyscal"""
+    description = (
+        "ERT forward model argument parser for wrapper around pyscal. "
+        "This parser should only be used in the forward model context, "
+        "where all arguments must always be passed. For other uses, "
+        "head to the pyscal client from the pyscal package"
+    )
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument("relperm_parameters_file", type=str)
```
markusdregi: A brief description of all the arguments is extremely useful if we are to generate documentation based on the arg_parser.

markusdregi:
@berland I did not fully get why you need a new executable here. Is it because not all the arguments will necessarily be provided by the user? If we could replace `__NONE__` by the empty string, and hence no value would be provided to the executable unless set by the user, would that moot the point of the additional executable?

Had some final comments on the code as well. But after those are addressed and I understand why we need the wrapper this is good to go 🚀

setup.py
berland: fixed

semeio/jobs/scripts/fm_pyscal.py
berland: fixed

semeio/jobs/scripts/fm_pyscal.py
berland: Fixed

semeio/jobs/scripts/fm_pyscal.py
berland: Fixed.

semeio/jobs/scripts/fm_pyscal.py
berland: Refactored out one function.

berland:
The main reason for having a wrapper in the ERT context is to be able to access parameters.txt directly, which I don't think makes sense for the pyscal client in the pyscal package - it has no clue of ERT whatsoever.

Secondly, I thought there were issues with optional arguments for forward models, therefore I have written the fm_pyscal arguments only as positional. This assumption might be wrong though, only learned from experience.

markusdregi:
> The main reason for having a wrapper in the ERT context is to be able to access parameters.txt directly, which I don't think makes sense for the pyscal client in the pyscal package - it has no clue of ERT whatsoever.

Makes sense. I would argue that we should perhaps introduce a data model for our forward models down the road, but that is a bigger discussion then this PR ;)

markusdregi:
So, whenever tests are passing I'm ready to have a final look and merge it.

berland:
Fixed, rebased and squashed.

PR #61, opened 2020-03-10
Proof-of-concept CSV_EXPORT1 reimplementation

berland:
Reimplemented CSV_EXPORT1 (as "CSV_EXPORT2") using fmu.ensemble.

https://equinor.github.io/fmu-ensemble/advancedusage.html#merging-data

### DIFF
--- a/semeio/jobs/scripts/csv_export2.py
+++ b/semeio/jobs/scripts/csv_export2.py
@@ -1,0 +1,36 @@
+import argparse
+import os
+import sys
+
+import pandas as pd
+from fmu import ensemble
+from res.enkf import ErtScript
+from semeio.jobs.csv_export2 import csv_export2
+
+
+class CsvExport2Job(ErtScript):
+    def run(self, *args):
+        main(args)
+
+
+def main(args):
+    """Setup parser"""
+    parser = argparse.ArgumentParser()
+    parser.add_argument("runpathfile", type=str)
+    parser.add_argument("outputfile", type=str)
+    parser.add_argument("time_index", type=str, default="monthly")
+    parser.add_argument("column_keys", nargs="+", default=None)
+    args = parser.parse_args(args)
+
+    csv_export2.csv_exporter(
+        runpathfile=args.runpathfile,
+        time_index=args.time_index,
+        outputfile=args.outputfile,
+        column_keys=args.column_keys,
+    )
+
+    print("{} csv-export written to {}".format(args.time_index, args.outputfile))
+
+
+if __name__ == "__main__":
+    main(sys.argv[1:])
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
         "xlrd",
         "stea",
         "pyscal>=0.4.0",
+        "fmu-ensemble",
     ],
     setup_requires=["setuptools_scm"],
     package_data={"": job_files},
--- a/tests/jobs/csv_export2/conftest.py
+++ b/tests/jobs/csv_export2/conftest.py
@@ -1,0 +1,49 @@
+import os
+import shutil
+
+import pytest
+from res.enkf import ResConfig
+
+TEST_DATA_DIR = os.environ.get("ERT_TESTDATA_ROOT")
+TEST_DATA_DIR_BERGEN = "/d/proj/bg/enkf/ErtTestData"
+TEST_DATA_DIR_STAVANGER = "/project/res-testdata/ErtTestData"
+
+
+def find_available_test_data():
+    if TEST_DATA_DIR is None:
+        if os.path.isdir(TEST_DATA_DIR_STAVANGER):
+            return TEST_DATA_DIR_STAVANGER
+        elif os.path.isdir(TEST_DATA_DIR_BERGEN):
+            return TEST_DATA_DIR_BERGEN
+
+    return TEST_DATA_DIR
+
+
+def has_equinor_test_data():
+    return os.path.isdir(find_available_test_data())
+
+
+def pytest_runtest_setup(item):
+    if item.get_closest_marker("equinor_test") and not has_equinor_test_data():
+        pytest.skip("Test requires Equinor data")
+
+
+@pytest.fixture()
+def ert_statoil_test_data(tmpdir):
+    cwd = os.getcwd()
+    tmpdir.chdir()
+    test_data_dir = os.path.join(find_available_test_data(), "ert-statoil", "spotfire")
+    shutil.copytree(test_data_dir, "test_data")
+    os.chdir(os.path.join("test_data"))
+
+    yield
+
+    os.chdir(cwd)
+
+
+@pytest.fixture()
+def setup_tmpdir(tmpdir):
+    cwd = os.getcwd()
+    tmpdir.chdir()
+    yield
+    os.chdir(cwd)
--- a/tests/jobs/csv_export2/test_integration.py
+++ b/tests/jobs/csv_export2/test_integration.py
@@ -1,0 +1,124 @@
+import sys
+
+import pytest
+from semeio.jobs.csv_export2 import csv_export2
+
+from tests.jobs.csv_export2 import conftest
+
+test_header = "ENSEMBLE,REAL,DATE,FOPR,X,FLUID_PARAMS:SWL,FLUID_PARAMS:SGCR,A,H\n"
+
+
+@pytest.mark.skipif(
+    conftest.find_available_test_data() is None, reason="no ert-statoil test-data"
+)
+@pytest.mark.usefixtures("ert_statoil_test_data")
+def test_failed_realization_no_summary_file(ert_statoil_test_data):
+    path_file = "pathfile_path_to_failed_realizations.txt"
+    export_file = "export.txt"
+    csv_export2.csv_exporter(
+        runpathfile=path_file,
+        time_index="monthly",
+        outputfile=export_file,
+        column_keys="FOPR",
+    )
+    verifyExportedFile(
+        export_file,
+        test_header,
+        {("iter-1", 1), ("iter-2", 2), ("iter-1", 0), ("iter-1", 2)},
+    )
+
+
+@pytest.mark.skipif(
+    conftest.find_available_test_data() is None, reason="no ert-statoil test-data"
+)
+@pytest.mark.usefixtures("ert_statoil_test_data")
+def test_one_iteration(ert_statoil_test_data):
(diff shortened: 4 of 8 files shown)


ertomatic:

markusdregi:
Thanks for the proof-of-concept :) As mentioned earlier we would prefer to make a solution on top of the new data API if we manage in time. But this is good to have around anyway 👏

berland:
Self-assessment: 
* Does not reproduce CSV_EXPORT1 exactly, it will handle inconsistent columns from each realization correctly (pandas is doing the job of concatenating frames, not string cat as in CSV_EXPORT1).
* Support for reading from ERT runpathfile is present, but not extensively used/tested. Surprises may pop up.
* fmu-ensemble is well tested. Testing for this PR only applies to checking the forward model argument is followed.
* DateInterval support in CSV_EXPORT1 has approximate support in fmu-ensemble, fmu-ensemble supports raw, daily, weekly, monthly, yearly. Not DateInterval=25 days as CSV_EXPORT1 supports.

semeio/jobs/scripts/csv_export2.py
```diff
@@ -41,6 +32,6 @@ def _get_args_parser():
     parser = argparse.ArgumentParser()
     parser.add_argument("runpathfile", type=str)
     parser.add_argument("outputfile", type=str)
-    parser.add_argument("datesetting", type=str)
-    parser.add_argument("vectors", nargs="+")
+    parser.add_argument("--vectors", nargs="+")
+    parser.add_argument("--datesetting", type=str, default="monthly")
```
berland, line 36: We should perhaps change these to fmu-ensemble's object names, so `--column_keys` and `--time_index` as option names (this is in turn derived from libecl..pandas_frame()). This makes it similar to the syntax of `ecl2csv summary` https://equinor.github.io/ecl2df/scripts.html#summary

oyvindeide:
Thanks for the PR!

>     * Does not reproduce CSV_EXPORT1 exactly, it will handle inconsistent columns from each realization correctly (pandas is doing the job of concatenating frames, not string cat as in CSV_EXPORT1).
> 
Do you think users will need to make a lot of changes to their jobs based on this difference?

If we are changing the API here, that will require everyone to change their jobs accordingly. While that should be fine to do at one point, we should evaluate the new API to make sure we don´t need to change it again when when the new storage based API is ready in hopefully not too long.

berland:
I am not really worried about the inconsistency bug being fixed. Those CSV files  have never been really usable anyway, so users have fixed that inconsistency  upstream (in Eclipse).

We are changing the API for the DateInterval, by ignoring it. This will create a different but not unusable CSV file for those who have had something peculiar in their DateInterval.

(If not, we recreate the DateInterval support, it is just a matter of creating a list of datetime.dates and send it to fmu.ensemble)

Bottom line, this is possible to just plug in, without users needing to modify anything, but in some cases the output will be slightly different.

berland:
With respect to the new storage  API, that can be naturally solved by making an analogue class to ScratchEnsemble (ErtEnsemble?) and nothing  needs to be changed in this export job.

semeio/jobs/scripts/csv_export2.py
```diff
@@ -0,0 +1,37 @@
...
+from res.enkf import ErtScript
+from fmu import ensemble
+
+
+class CsvExport2Job(ErtScript):
+    def run(self, *args):
+        parser = _get_args_parser()
+        args = parser.parse_args(args)
+        csv_exporter(args.runpathfile, args.time_index, args.outputfile, vectors=args.column_keys)
+        print("{} csv-export written to {}".format(args.datesetting, args.outputfile))
```
berland, line 15: datesetting -> time_index

semeio/jobs/scripts/csv_export2.py
```diff
@@ -0,0 +1,37 @@
...
+import argparse
+
+import pandas as pd
+
+from res.enkf import ErtScript
+from fmu import ensemble
+
+
+class CsvExport2Job(ErtScript):
+    def run(self, *args):
```
berland, line 11: Can  we pull this run() out of the class, so that we can also  facilitate a command line endpoint doing the same with the same arguments?

oysteoh: ErtScript requires a run() method in the class inherit from it. Do you want an additional command line endpoint? And if so, where?

markusdregi: But since the `run` function is not using `self` at all I'm guessing this does not even need to be an internal script ;) Which makes sense as it crawls the file system ;)

markusdregi:
I would like to see some tests :)

semeio/jobs/scripts/csv_export2.py
berland: I think I want an additional endpoint also yes, very useful for debugging, and for relaunching exports without interferring with a config file.

oysteoh: ok, i'll move it!

oysteoh:
> I would like to see some tests :)

some tests i can write - is it sufficient mocking the fmu_ensemble interface and only test our side of this?

markusdregi:
> some tests i can write - is it sufficient mocking the fmu_ensemble interface and only test our side of this?

For most of it yes, but would we not like an integration test as well? :)

berland:
You might want to do some review of the function: https://github.com/equinor/fmu-ensemble/blob/master/src/fmu/ensemble/ensemble.py#L223

Export via runpath files are not well tested in fmu-ensemble - how much can the files vary?

fmu.ensemble tests versus this file for each CI run: https://github.com/equinor/fmu-ensemble/blob/master/tests/data/ert-runpath-file (which has relative paths, supported in fmu.ensemble for the sake of CI)

oysteoh:
> > some tests i can write - is it sufficient mocking the fmu_ensemble interface and only test our side of this?
> 
> For most of it yes, but would we not like an integration test as well? :)

i looked more thoroughly in the code, if i write a test it will be an integration test - nothing is going on in semeio. I will try to get something up and running ! 👍

oysteoh:
> This will fail, Realization has been renamed to REAL in the FMU standard, Iteration is renamed to ITER, and Date to DATE. There might be a column called ENSEMBLE as well.

Yes i know - it was just the initial work of getting tests in - copied from the old csv job as a starting point ... rest of the day i struggled to get my computer working! will fix - but thanks anyways - good to see you keep an eye on the issue! 👍

semeio/jobs/scripts/csv_export2.py
markusdregi: Are we sure we should expose another endpoint? With future design in mind an export from ERT should be done with a configuration file/experiment in mind and not an ensemble on disk. Otherwise we immediately have to break this behaviour again when we fetch the data from the API.

Which use case is it that we are not covering by having the workflow `CSV_EXPORT2` available and supporting to run it via the ERT cli as: `ert workflow CSV_EXPORT2 my_config.ert`?

berland: The endpoint may belong to fmu-ensemble (it has been planned for years but I have not been able to come up with a name, or how the command line api should look like). The command line api is simpler for this semeio job, since the difficult part is put into the runpath file.

The use-case for an endpoint is being able to do reliable (test/debugging) exports, perhaps on other users data,  without relying on whether the config file has been tampered with, or knowing which config file to pick. Or if you generate your own runpath file. Not  sure how relevant it is for regular users, so perhaps the endpoint should be somewhat hidden.

markusdregi: That makes sense, but I think the data that ERT serves should be based around configuration files/experiments. And this is why we need to do persist configuration files after they have been ran to be able to serve this even more reliably in the future.

I would suggest putting that second end point in `fmu-ensembles` or another repository.

For ERT I'm thinking that we should be able to do `ert cases` to get a list of the cases available and then `ert export <case_name>` to get an export of the data from that case instead of basing it on the scratch-disk. This should still fly as long as you have access to storage, but will shift the perceived location of master data from scratch to storage. Which makes 100% sense given that storage is a place you can trust with your data ;)

setup.py
```diff
@@ -56,7 +56,7 @@ def package_files(directory):
         "xlrd",
         "stea",
         "pyscal>=0.4.0",
-        "fmu-ensemble",
+        "fmu-ensemble<=1.2.2",
```
oyvindeide, line 59: Think you need to pin it < 1.2.2 or maybe better with != 1.2.2, think that should work?

oysteoh: ah, i thought you ment i should use 1.2.2... will try <1.2.2

berland:
A special reason to pin the version to current stable? (it will work with a lot older releases, but perhaps not very relevant to test that)

oysteoh:
> A special reason to pin the version to current stable? (it will work with a lot older releases, but perhaps not very relevant to test that)

It was just something we tried, it didn't work out so we'll have to fix this in other ways !

tests/jobs/csv_export2/test_integration.py
```diff
@@ -0,0 +1,126 @@
...
+            outputfile=export_file,
+            column_keys="FOPR",
+        )
+
+
+def verifyExportedFile(exported_file_name, result_header, result_iter_rel):
+    with open(exported_file_name, "r") as exported_file:
+        lines = exported_file.readlines()
+
+        print("Header")
```
markusdregi, line 113: Please remove the `print` statements

tests/jobs/csv_export2/conftest.py
```diff
@@ -0,0 +1,49 @@
+import os
+import shutil
+
+import pytest
+from res.enkf import ResConfig
+
+TEST_DATA_DIR = os.environ.get("ERT_TESTDATA_ROOT")
```
markusdregi, line 7: I'm only letting this slide as this is considered a temporary solution. In general I would like us to stay away from dragging in not open test data.

oysteoh: yes... but it is already done this way for other tests as well ... agree it could slide considered a temp solution, but there are more permanent code in here with the same stuff ( correlated_observation_scaling ) . If this is not considered ok we should probably spawn an issue on this - at least for the other tests?

markusdregi:
The single commit is in dire need of a new message 😬

tests/jobs/csv_export2/conftest.py
markusdregi: I've already made an issue for this some time back ;)

PR #65, opened 2020-03-16
Move tests to new folder for komodo-testing

ManInFez:
Prevent the test from removing libres test-data


dotfloat:
I think it makes more sense to `git clone` libres _after_ removing everything but the tests. Ie., `git clone` right before `pytest`.

ManInFez:
Still would have to be in a separate directory or pytest would pick up libres-tests

Issue #43, opened 2020-02-18
[COS] Provide `--outfile` option


oyvindeide:
Correlated Observations scaling

Provide output to file in addition to terminal?

jondequinor:
What will go into the file? The terminal output is currently 
```
Calculation scaling factor, nr of primary components: 2, number of observations: 4
Scaling factor calculated from (named_dict(index=None, key='WPR_DIFF_1'),)
Keys: ['WPR_DIFF_1'] scaled with scaling factor: 1.41421356237
Calculation scaling factor, nr of primary components: 2, number of observations: 3
Scaling factor calculated from (named_dict(index=(400, 800, 1200), key='WPR_DIFF_1'),)
Keys: ['WPR_DIFF_1'] scaled with scaling factor: 1.22474487139
```
Should it instead be exported into some portable data format?

jondequinor:
We are currently discussing how to handle output from workflows here: https://github.com/equinor/semeio/pull/66 — this discussion will most likely affect this issue/make it more actionable.

markusdregi:
The output files are dictated by the reporter. Later, the data will be available through storage instead.

Issue #42, opened 2020-02-18
[COS] Improve stdout formatting


oyvindeide:
Correlated Observations scaling

Important information is quickly hidden in and overflow of metadata (when all keys are listed, and the datastructures leak through to the terminal)

jondequinor:
We are currently discussing how to handle output from workflows here: https://github.com/equinor/semeio/pull/66 — this discussion will most likely affect this issue/make it more actionable.

markusdregi:
The output will be structured differently when using the reporter. For now superseded by #112

PR #66, opened 2020-03-17
Cos export

### LINKED ISSUE #45: [COS] Output singular values
Correlated Observations scaling

Users would like output of singular values to stdout

https://github.com/equinor/semeio/blob/2e779f8d4b3c52bafbc519796dbe5f928ea57af7/semeio/jobs/correlated_observations_scaling/scaled_matrix.py#L81

jondequinor:
**Main points**:
- Creates an `obs_util` package, for a lack of a better name. Also splits up `test_scaling_job` into more fitting constituents.
- Stamps out `facade` from both COS and SC. This simplifies testing, but also moves both jobs to the same abstraction layer which also eases interoperability.
- SC is now more of a config-generator, and the code reflects this. And SC uses the `CorrelatedObservationsScalingJob`, which improves encapsulation (SC knows less about COS).

**Suggested `Reporter`**:

Principles:

Data flowing out of workflows

- into other workflows 
- to the user

Fixes #65 
Arguably fixes #45 #43 #42

### DIFF
--- a/semeio/jobs/spearman_correlation_job/job.py
+++ b/semeio/jobs/spearman_correlation_job/job.py
@@ -45,7 +45,7 @@ def _spearman_correlation(facade, obs_keys, threshold, dry_run):
         )
     )
 
-    clustered_data = _cluster_data(data)
+    clustered_data = _remove_singular_obs(_cluster_data(data))
 
     job_configs = _config_creation(clustered_data)
 
@@ -77,6 +77,19 @@ def _cluster_data(data):
             groups[nr] = {}
         groups[nr].update({key: [index for _, _, index in cluster_group]})
     return groups
+
+
+def _remove_singular_obs(clusters):
+    """Removes clusters with a singular observation."""
+    new_cluster_index = 0
+    multiobs_clusters = {}
+    for _, cluster in clusters.items():
+        if sum(map(len, cluster.values())) > 1:
+            multiobs_clusters[new_cluster_index] = cluster
+            new_cluster_index += 1
+        else:
+            print("Removed cluster with singular observation: {}".format(cluster))
+    return multiobs_clusters
 
 
 def _config_creation(clusters):
--- a/tests/jobs/spearman_correlation_job/unit/test_spearman_job.py
+++ b/tests/jobs/spearman_correlation_job/unit/test_spearman_job.py
@@ -61,6 +61,23 @@ def test_make_clusters(test_input, expected_result):
 @pytest.mark.parametrize(
     "test_input,expected_result",
     [
+        ({}, {}),
+        ({1: {"KEY_1": [1]}}, {}),
+        ({1: {"KEY_1": [1, 2]}}, {0: {"KEY_1": [1, 2]}}),
+        (
+            {1: {"KEY_1": [1], "KEY_2": [2]}, 2: {"KEY_2": [2]}},
+            {0: {"KEY_1": [1], "KEY_2": [2]}},
+        ),
+    ],
+)
+def test_single_obs_clusters(test_input, expected_result):
+    result = spearman._remove_singular_obs(test_input)
+    assert result == expected_result, "bad removal of singular obs clusters"
+
+
+@pytest.mark.parametrize(
+    "test_input,expected_result",
+    [
         (
             {1: {"KEY_1": [1]}},
             [{"CALCULATE_KEYS": {"keys": [{"key": "KEY_1", "index": [1]}]}}],


semeio/jobs/correlated_observations_scaling/job.py
```diff
@@ -1,232 +1,112 @@
...
-                    obs_node.set_std_scaling(scale_factor)
-            elif obs_vector.getImplementationType().name != "SUMMARY_OBS":
-                obs_node.updateStdScaling(scale_factor, event.active_list)
-    print(
-        "Keys: {} scaled with scaling factor: {}".format(
-            [event.key for event in events], scale_factor
+
+
+class ScalingJob(object):
+    def __init__(self, facade, user_config_dict):
```
oyvindeide: I´ve been thinking a bit about this, and was thinking I should refactor to make the scaling job less knowledgeble about the facade, and send observation keys, etc. instead. First of, do you agree? If so this could be a good oppertunity to do so, if not, nvm ;)

tests/jobs/correlated_observations_scaling/unit/__init__.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+import pytest
+
+
+@pytest.fixture()
+def setup_tmpdir(tmpdir):
```
oyvindeide: I`ve no problem with moving this to a new file, but should be a `conftest.py` file (could potentially be the root one, or a new one in this location).

markusdregi:
Thanks for the write up @jondequinor!

Up until this point I've envisioned a clear separation between data and status reporting. Where we offer data posting and data subscription through one interface and the possibility to report your status through a different interface. I'm not completely sure why however. Maybe it is an artificial separation?

For the messages to the user; would this be more or less what a traditional logger interface would equip you with?

semeio/jobs/correlated_observations_scaling/job.py
```diff
@@ -1,232 +1,118 @@
...
-            elif obs_vector.getImplementationType().name != "SUMMARY_OBS":
-                obs_node.updateStdScaling(scale_factor, event.active_list)
-    print(
-        "Keys: {} scaled with scaling factor: {}".format(
-            [event.key for event in events], scale_factor
+from semeio.jobs.correlated_observations_scaling.scaled_matrix import DataMatrix
+from semeio.jobs.correlated_observations_scaling.validator import has_keys, is_subset
+
+
+class ScalingJob(object):
```
markusdregi: To me the API of `ScalingJob` and it's usage hints towards a reusable object that can scale different sets of measured data. However, I think the current local assumption of the export location prevents this. Do we intend to open up for using the scaling job multiple times (and if so, does that yield any value?) or is it a create once, use once object?

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -154,6 +160,11 @@ def build_schema():
                     }
                 },
             },
+            "EXPORT_DIR": {
+                MK.Type: types.String,
+                MK.Required: False,
+                MK.ElementValidators: (_path_exists,),
```
markusdregi: Should it even be a folder perhaps?

semeio/jobs/correlated_observations_scaling/obs_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Loops through the user input and identifies wildcards in observation
+    names and expands them.
+    """
+    new_dict = deepcopy(user_dict)
+    for main_key, value in user_dict.items():
+        new_entries = []
+        if main_key in ("UPDATE_KEYS", "CALCULATE_KEYS"):
+            for val in value["keys"]:
+                if "*" in val["key"]:
```
markusdregi: This appears to be an optimisation of `_expand_wildcard`? Should we move it into the `_expand_wildcard` function (and hence also move the knowledge of `*` being the wildcard)?

Then we could rewrite the body of the main loop as something like:
```python
    if main_key not in ("UPDATE_KEYS", "CALCULATE_KEYS"):
        continue

    new_entries = []
    for val in value["keys"]:
        new_entries.extend(_expand_wildcard(obs_list, val, val["key"]))
    new_dict[main_key]["keys"] = new_entries
```

jondequinor: > Up until this point I've envisioned a clear separation between data and status reporting. Where we offer data posting and data subscription through one interface and the possibility to report your status through a different interface. I'm not completely sure why however. Maybe it is an artificial separation?

Currently, I don't value that distinction. Because to me, both (data and status) should be thought as events. And I think this is also intuitively true for many users.
 
> For the messages to the user; would this be more or less what a traditional logger interface would equip you with?

The logger interface is semantically limiting on its own. The only thing that varies is the log level. You could argue that structured logging could add the semantics needed, but then you'd have to juggle the difference between non-user records and user feedback.

Also, I think the python logger offers too many features and should not be offered to our workflow authors as a part of a _workflow api_ for getting messages back to our users. I am heavily influenced by Dave Cheney: https://dave.cheney.net/2015/11/05/lets-talk-about-logging

jondequinor:
> Had an initial go trough of your PR for my own educational purposes although it is not currently in the review column. Made some comment on the way, but feel free to silently ignore them for now if you feel like it ;)

Yes, the reporting/exporting part was actually made before this discussion started, and should be considered stale.

semeio/jobs/correlated_observations_scaling/exporter.py
```diff
@@ -0,0 +1,49 @@
...
+        return json.JSONEncoder.default(self, obj)
+
+
+class Exporter(object):
+    def __init__(self, config):
+        self._data = {}
+        self._dir = config.snapshot.EXPORT_DIR
+
+    def export(self, key, data, message=None):
+        if message is not None:
```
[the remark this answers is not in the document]
jondequinor: It seemed like this was a common thing in COS. Printing data to the terminal, and possibly having someone interpret that output into data. This combines both, for no particular reason other than it might be useful during the scaling (in e.g. the status window in the GUI), and after the workflow has ended (on disk). But I am not confident it should be a part of such an interface.

semeio/jobs/correlated_observations_scaling/exporter.py
```diff
@@ -0,0 +1,49 @@
...
+            print(message)
+
+        if not self._dir:
+            return
```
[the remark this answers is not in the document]
jondequinor: The report/export directory wouldn't currently be vital for the operation of neither COS nor SC. But we can always print stuff in the terminal.

I was thinking of using `cwd` if not specified, maybe that's the better approach.

semeio/jobs/correlated_observations_scaling/job.py
```diff
@@ -1,232 +1,118 @@
...
+            raise RuntimeError("get_errors called before validation")
+        return self._validation_errors
+
+    def get_calc_keys(self):
+        return [event.key for event in self._config.snapshot.CALCULATE_KEYS.keys]
+
+    def get_index_lists(self):
+        return [event.index for event in self._config.snapshot.CALCULATE_KEYS.keys]
+
+    def _setup_configuration(self, config_data):
```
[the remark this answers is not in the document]
jondequinor: I haven't thought about this code as it is more or less untouched.

But I do agree that the expansion should happen at some other time, unless there's really good reason.

semeio/jobs/correlated_observations_scaling/job.py
jondequinor: If the scaling job is idempotent (I'm not entirely sure, maybe @oyvindeide can chime in), then there's no reason for why it shouldn't run multiple times.

That the export location might give rise to data overwriting/corrution, shouldn't the concern of the scaling job. Rather, it should be of some parent workflow object.

semeio/jobs/correlated_observations_scaling/job_config.py
jondequinor: What do you mean?

jondequinor:
> Up until this point I've envisioned a clear separation between data and status reporting. Where we offer data posting and data subscription through one interface and the possibility to report your status through a different interface. I'm not completely sure why however. Maybe it is an artificial separation?

Currently, I don't value that distinction. Because to me, both (data and status) should be thought as events. And I think this is also intuitively true for many programmers/super users/workflow authors.
 
> For the messages to the user; would this be more or less what a traditional logger interface would equip you with?

The logger interface is semantically limiting on its own. The only thing that varies is the log level. You could argue that structured logging could add the semantics needed, but then you'd have to juggle the difference between non-user records and user feedback.

Also, I think the python logger offers too many features and should not be offered to our workflow authors as a part of a _workflow api_ for getting messages back to our users. I am heavily influenced by Dave Cheney: https://dave.cheney.net/2015/11/05/lets-talk-about-logging

jondequinor:
Moved the refactor to https://github.com/equinor/semeio/pull/74

Discussion of cos-export could continue here.

semeio/jobs/correlated_observations_scaling/job_config.py
markusdregi: It is only validated that it exists... I guess we want it to be a folder?

semeio/jobs/correlated_observations_scaling/job.py
markusdregi: I agree, but then I think the consumer of the ScalingJob should also provide the exporter, as otherwise usage multiple times is very unnatural.

semeio/jobs/correlated_observations_scaling/exporter.py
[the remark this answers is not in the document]
markusdregi: It just seems to me like the method is trying to do two potentially unrelated things that can be left to the caller, namely printing a message and exporting some data.

semeio/jobs/correlated_observations_scaling/job_config.py
jondequinor: Right. Good point.

semeio/jobs/correlated_observations_scaling/job.py
jondequinor: Right, this exporter would have to be moved into whatever replaces `ErtScript`. The exporter will exist in `ScalingJob` given `ErtScript`.

dotfloat:
I think at this point we can benefit greatly by having a proper message passing bus that can pass data to and from an ERT instance. As far as I can tell, we can make the following assumptions:
* Each workflow/forward model (henceforth job) can publish some data
* Each piece of data can be uniquely identified by a job ID and a key
* Data can be one of the JSON primitives, vector spaces (numpy data) or time-series (and other scientific-computing types maybe)
* Jobs can run in parallel but not concurrently. That is, job A cannot send data to job B and then wait for job B to respond, but they may run at the same time
* Data generated by a job is immutable after the job is finished

As such, I propose that ERT is equipped with a key-value store for each job that is capable of keeping track of data. The implementation of how this should actually be stored is out of scope for this comment. I reckon we can start with a dumb Python dictionary.

Imaginary code example of a future Python job:
```python
from ert.service import Reporter

# Reporter gets connected
Reporter.connect("http://localhost:1337", "my_job_id")

# An entry named "key" will appear in ERT with the value "value" under the job "my_job_id"
Reporter.report("key", "value")

# Report and pretty-print
Reporter.report("key2", 42, description="The answer", stdout=True)

# Report numpy matrix efficiently that will show up in ERT GUI an may be rendered as a matrix
Reporter.report("singular_values", computed_singular_values, description="Caluclated Singular Values", stdout=True)

# Report time series that ERT will know is a time-series and can plot it accordingly
Reporter.report("summary", my_timeseries)

# Cool feature to have: Report streams that can be updated (ie, modify existing values)
# If we have this then implementing streaming stdout and stderr or a proper logger for a job is easy
myio = StringIO()
Reporter.report("custom_log", myio)
myio.write("Hello, world!")  # Reporter detects new data, it magically shows up in ERT

# Reporter automatically captures stdout and stderr, sends it to ERT
print("Hello, world!")
```

ERT will coalesce this data as such in some kind of database:
```
- "PLOT:1":
  - stdout: Bytestring Stream (Closed)
  - stderr: Bytestring Stream (Closed)
  - status: Enum
- "COS:2":
  - stdout: Bytestring Stream (Open)
  - stderr: Bytestring Stream (Open)
  - status: Enum
  - singular_values: Dense NxM float64 Matrix
```

# Implementation

We can use [ZeroMQ](https://zeromq.org/) using Python [PyZMQ](https://github.com/zeromq/pyzmq) to create and manage sockets, so that we don't have to implement low-level message passing and queues ourselves. We can use [MessagePack](https://msgpack.org/) to efficiently generate and send messages. __In particular, if we do this we can avoid serialising and then deserialising numpy objects by simply passing the data buffer as-is__ Numpy's ABI is stable and in the future if we use a different language we can convert Numpy buffers to whatever is native in the other language. In particular this'll easily allows us to efficiently transfer sparse matrices.

ERT message bus initialisation steps:
1. Use ZeroMQ to intialise a server
2. Start listening in a separate thread
3. When `job_dispatch` is ran, do:
   1. Set `ERT_MESSAGE_BUS` envvar to server addres (eg. `http://localhost:1337` or `unix:///var/run/1234/ert.server`)
    2. Set `ERT_JOB_ID` envvar to a unique job ID (eg. `COS:2`)

Job initialisation steps:
1. `from ert.service import Reporter` is called
2. Reporter automatically connects using data in `ERT_MESSAGE_BUS`, `ERT_JOB_ID`
3. Reporter redirects stdout and stderr

Reporting data:
1. Metadata is stored in some MessagePack object. This contains 'action' (ie. `Assign`, `StringAppend`, `InsertMatrixColumn`, as needed), 'key', 'description', maybe some other information.
2. Reporter knows how to send data that we are interested in. Ie, maybe there's a dictionary of `send_data` that we have implemented where `send_data[type(val)](socket, val)`. This will allow us to send numpy data directly without unnecessary copying.
3. If it's a streaming data type (ie a `StringIO`) start a thread or something that polls data

# Closing Remarks
This type of message bus is something that will solve many existing problems in libres (`status.json`) as well as make our efforts in COS simpler. It is also language-agnostic for the most part, and doesn't require an EnKFMain instance. :cloud:-ready!

Because we own both sides of the channel (both ERT and this Reporter class) we can implement this efficiently and without the need for flask and serialising massive data to CSV or the like (*shudder*).

If we stick to Unix sockets, we don't need to think about network security. It's possible to forward these via SSH, although it's not trivial and may not work on RHEL 6.

Future efforts may allow jobs to fetch data from previous jobs that have finished fairly trivially by asking ERT for it via the message bus.

Issue #68, opened 2020-03-18
Implement first version of AUTO_CORRELATE_OBSERVATIONS


markusdregi:
Implement a first version of the workflow `AUTO_CORRELATE_OBSERVATIONS` described in https://github.com/equinor/semeio/issues/67. This will not be using pipes and filters, but instead be a Python implementation on top of the existing tooling. Then we can request user feedback on the setup, while we reimplement behind the scenes without breaking changes.

PR #70, opened 2020-03-19
Misfit preprocessor

markusdregi:
Resolves: https://github.com/equinor/semeio/issues/68

**Misson of the PR**
- Establish the user interface of the observation correlation tooling that is to pushed out in the company as described in #68.
- Add a bigger test base to ensure that results are as expected. In particular I've mocked parts of the `MeasuredData` structure from `ERT`, which allows me to quickly evaluate thousands of forward models without depending on libres.

**Remaining work for this PR**
- Find a better name for the workflow. The idea of `AUTO_CORRELATE_OBSERVATIONS` was that it would be semi-automatic. However, this is a very overloaded term in statistics, so we should call it something else. Perhaps `SCALE_OBSERVATION_CLUSTERS`?

**Remaining work**
- Write documentation. I've started filling out the schema. Will try to use configsuite tooling afterwards to generate a first iteration.
- Unite the schema of spearman correlation in this workflow and the spearman job.

### DIFF
--- a/semeio/jobs/spearman_correlation_job/job.py
+++ b/semeio/jobs/spearman_correlation_job/job.py
@@ -4,7 +4,14 @@ import itertools
 from scipy.cluster.hierarchy import fcluster, linkage
 
 
-def spearman_job(measured_data, threshold):
+def spearman_job(
+    measured_data,
+    threshold,
+    criterion="inconsistent",
+    depth=2,
+    method="single",
+    metric="euclidean",
+):
     """
     Given measured_data and threshold, it returns configurations that describe
     scaling of these data.
@@ -17,7 +24,9 @@ def spearman_job(measured_data, threshold):
 
     correlation_matrix = _calculate_correlation_matrix(simulated_data)
 
-    clusters = _cluster_analysis(correlation_matrix, threshold)
+    clusters = _cluster_analysis(
+        correlation_matrix, threshold, criterion, depth, method, metric
+    )
 
     columns = correlation_matrix.columns
 
@@ -87,6 +96,6 @@ def _calculate_correlation_matrix(data):
     return data.rank().corr(method="pearson")
 
 
-def _cluster_analysis(correlation_matrix, threshold):
-    a = linkage(correlation_matrix, "single")
-    return fcluster(a, threshold)
+def _cluster_analysis(correlation_matrix, threshold, criterion, depth, method, metric):
+    a = linkage(correlation_matrix, method, metric)
+    return fcluster(a, threshold, criterion=criterion, depth=depth)
--- a/semeio/jobs/config_workflow_jobs/MISFIT_PREPROCESSOR
+++ b/semeio/jobs/config_workflow_jobs/MISFIT_PREPROCESSOR
@@ -1,0 +1,2 @@
+INTERNAL    True
+SCRIPT      ../scripts/misfit_preprocessor.py
--- a/semeio/jobs/misfit_preprocessor/__init__.py
+++ b/semeio/jobs/misfit_preprocessor/__init__.py
@@ -1,10 +1,14 @@
+from semeio.jobs.misfit_preprocessor.exceptions import ValidationError
 from semeio.jobs.misfit_preprocessor.config import (
     MisfitPreprocessorConfig,
     assemble_config,
 )
+from semeio.jobs.misfit_preprocessor.job import run
 
 
 __all__ = (
+    "ValidationError",
     "MisfitPreprocessorConfig",
     "assemble_config",
+    "run",
 )
--- a/semeio/jobs/misfit_preprocessor/__init__.py
+++ b/semeio/jobs/misfit_preprocessor/__init__.py
@@ -1,0 +1,10 @@
+from semeio.jobs.misfit_preprocessor.config import (
+    MisfitPreprocessorConfig,
+    assemble_config,
+)
+
+
+__all__ = (
+    "MisfitPreprocessorConfig",
+    "assemble_config",
+)
--- a/semeio/jobs/misfit_preprocessor/config.py
+++ b/semeio/jobs/misfit_preprocessor/config.py
@@ -1,0 +1,396 @@
+import collections
+import configsuite as cs
+import copy
+import fnmatch
+
+
+_SCALING = "scaling"
+_THRESHOLD = "threshold"
+_STD_CUTOFF = "std_cutoff"
+_ALPHA = "alpha"
+_CLUSTERING = "clustering"
+_METHOD = "method"
+SPEARMAN_CORRELATION = "spearman_correlation"
+_FCLUSTER = "fcluster"
+_SPEARMAN_THRESHOLD = "t"
+_CRITERION = "criterion"
+_INCONSISTENT = "inconsistent"
+_DISTANCE = "distance"
+_MAXCLUST = "maxclust"
+_MONOCRIT = "monocrit"
+_MAXCLUST_MONOCRIT = "maxclust_monocrit"
+_DEPTH = "depth"
+_LINKAGE = "linkage"
+_SINGLE = "single"
+_METHODS = (
+    _SINGLE,
+    "complete",
+    "average",
+    "weighted",
+    "centroid",
+    "ward",
+)
+_METRIC = "metric"
+_EUCLIDEAN = "euclidean"
+_METRICS = (
+    "braycurtis",
+    "canberra",
+    "chebyshev",
+    "cityblock",
+    "correlation",
+    "cosine",
+    "dice",
+    _EUCLIDEAN,
+    "hamming",
+    "jaccard",
+    "jensenshannon",
+    "kulsinski",
+    "mahalanobis",
+    "matching",
+    "minkowski",
+    "rogerstanimoto",
+    "russellrao",
+    "seuclidean",
+    "sokalmichener",
+    "sokalsneath",
+    "sqeuclidean",
+    "yule",
+)
+_OBSERVATIONS = "observations"
+
+
+_DEFAULT_VALUES = {
+    _SCALING: {_THRESHOLD: 0.95, _STD_CUTOFF: 1e-6, _ALPHA: 3.0},
+    _OBSERVATIONS: (),
+    _CLUSTERING: {
+        _METHOD: SPEARMAN_CORRELATION,
+        SPEARMAN_CORRELATION: {
+            _FCLUSTER: {_CRITERION: _INCONSISTENT, _DEPTH: 2},
+            _LINKAGE: {_METHOD: _SINGLE, _METRIC: _EUCLIDEAN},
+        },
+    },
+}
+
+
+def _bounds_validator(
+    lower=None, lower_inclusive=True, upper=None, upper_inclusive=True
+):
+    msg = "Value must be within the interval {}, {}".format(
+        "<"
+        if lower is None
+        else "{}{}".format("[" if lower_inclusive else "(", lower,),
+        ">"
+        if upper is None
+        else "{}{}".format(upper, "]" if upper_inclusive else ")",),
+    )
+
+    @cs.validator_msg(msg)
+    def validator(value):
+        lower_valid = lower is None or (
+            lower <= value if lower_inclusive else lower < value
+        )
+        upper_valid = upper is None or (
+            value <= upper if upper_inclusive else value < upper
+        )
+        return lower_valid and upper_valid
+
+    return validator
+
+
+def _one_of(*valid_values):
+    msg = "Value must be one of: ({})".format(", ".join(map(str, valid_values)))
+
+    @cs.validator_msg(msg)
+    def validator(value):
+        return value in valid_values
+
+    return validator
+
+
+@cs.validator_msg("t must be an integer if a max cluster criteria is choosen")
+def _t_is_int_if_maxclust(spearman_config):
+    if spearman_config[_CRITERION] in (_MAXCLUST, _MAXCLUST_MONOCRIT):
+        return isinstance(spearman_config[_SPEARMAN_THRESHOLD], int)
+    return True
+
+
+@cs.transformation_msg(
+    (
+        "Will be defaulted to 5 for all max cluster criterias and "
+        "1.15 for all other criterias"
+    )
+)
+def _inject_default_t(config):
+    config = copy.deepcopy(config)
+    criterion = config.get(_CRITERION)
+    if _SPEARMAN_THRESHOLD in config:
+        pass
+    elif criterion in (_MAXCLUST, _MAXCLUST_MONOCRIT):
+        config[_SPEARMAN_THRESHOLD] = 5
+    elif criterion in (_INCONSISTENT, _DISTANCE, _MONOCRIT):
+        config[_SPEARMAN_THRESHOLD] = 1.15
+    else:
+        raise ValueError("Unknown criterion: {}".format(criterion))
+    return config
+
+
+SPEARMAN_CORRELATION_SCHEMA = {
+    cs.MetaKeys.Type: cs.types.NamedDict,
+    cs.MetaKeys.Description: (
+        "The {sc} clustering is supporting multiple parameters."
+    ).format(sc=SPEARMAN_CORRELATION),
+    cs.MetaKeys.Content: {
+        _FCLUSTER: {
+            cs.MetaKeys.Type: cs.types.NamedDict,
+            cs.MetaKeys.Description: (
+                "The {fcluster} implementation is backed by scipy and for a "
+                "more detailed description we refer the reader to the "
+                "documentation of scipy.cluster.hierarchy.fcluster` "
+                "(https://docs.scipy.org/doc/scipy/reference/generated"
+                "/scipy.cluster.hierarchy.fcluster.html)."
+            ).format(fcluster=_FCLUSTER),
+            cs.MetaKeys.ElementValidators: (_t_is_int_if_maxclust,),
+            cs.MetaKeys.Transformation: _inject_default_t,
+            cs.MetaKeys.Content: {
+                _SPEARMAN_THRESHOLD: {
+                    cs.MetaKeys.Type: cs.types.Number,
+                    cs.MetaKeys.Description: (
+                        "Scalar threshold for the clustering. When a "
+                        "'{maxclust}' {criterion} is used, the scalar gives "
+                        "the maximum number of clusters to be formed."
+                    ).format(maxclust=_MAXCLUST, criterion=_CRITERION),
+                    cs.MetaKeys.ElementValidators: (
+                        _bounds_validator(lower=0, lower_inclusive=False),
+                    ),
+                },
+                _CRITERION: {
+                    cs.MetaKeys.Type: cs.types.String,
+                    cs.MetaKeys.Description: (
+                        "The criterion to use in forming flat clusters. "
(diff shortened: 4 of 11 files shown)


jondequinor:
Meeeeh.. :man_shrugging: I don't know man…

Anyway, be aware of the quite major refactoring going on in https://github.com/equinor/semeio/pull/66

markusdregi:
I did not really get the meh?

markusdregi:
Regarding the refactoring, I’m keeping a close look on it and if I manage to get anything done it will be on the config side until you merge your work to avoid conflicts

README.md
```diff
@@ -1,3 +1,5 @@
 # semeio #
 
 Jobs and workflow jobs used in [ERT](https://github.com/equinor/ert).
+
+....
```
jondequinor: This was the _meh_.

markusdregi: How come? It is the perfect dummy commit 🥳

jondequinor: Please use ellipsis …

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+_EUCLIDEAN = "euclidean"
+_METRICS = (
+    "braycurtis",
+    "canberra",
+    "chebyshev",
+    "cityblock",
+    "correlation",
+    "cosine",
+    "dice",
+    _EUCLIDEAN,
```
xjules, line 40: it's quite many metrics, are all the metrics really necessary to be exposed? If yes I would perhaps suggest this (in order not to define name of each single metric):
```python
from enum import Enum

class Metrics(Enum):
         euclidean=0
         cityblock=1

# then you can access name
Metrics.euclidean.name
```

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+    def validator(value):
+        return value in valid_values
+
+    return validator
+
+
+@cs.validator_msg("t must be an integer if a max cluster criteria is choosen")
+def _t_is_int_if_maxclust(spearman_config):
+    if spearman_config[_CRITERION] in (_MAXCLUST, _MAXCLUST_MONOCRIT):
+        return isinstance(spearman_config[_T], int)
```
xjules, line 97: what's "t"? Can we replace it with something more descriptive?

xjules: _CLUSTER_THRESHOLD = "t" ?

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+    "sqeuclidean",
+    "yule",
+)
+_OBSERVATIONS = "observations"
+
+
+def _bounds_validator(
+    lower=None, lower_inclusive=True, upper=None, upper_inclusive=True
+):
+    msg = "Value must be within the interval {}, {}".format(
```
xjules, line 62: Very nice! It took me a while to parse this :)

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+                    cs.MetaKeys.Type: cs.types.String,
+                    cs.MetaKeys.ElementValidators: (_one_of(*_METRICS),),
+                },
+            },
+        },
+    },
+}
+
+
+_CLUSTERING_SCHEMA = {
```
xjules, line 175: Is the separation of `_CLUSTERING_SCHEMA` and `_SPEARMAN_CORRELATION_SCHEMA` due to readability or something else? Also why the name `_SPEARMAN_CORRELATION_SCHEMA` when it deals only with `fcluster` in scipy?

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+            cs.MetaKeys.Type: cs.types.String,
+            cs.MetaKeys.Description: "TODO",
+            cs.MetaKeys.ElementValidators: (_one_of(_SPEARMAN_CORRELATION),),
+        },
+        _SPEARMAN_CORRELATION: _SPEARMAN_CORRELATION_SCHEMA,
+    },
+}
+
+
+_SCALING_SCHEMA = {
```
xjules, line 189: I reckon this schema is work in progress.

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+    def __init__(self, observation_keys):
+        Context = collections.namedtuple("Context", ("observation_keys",))
+        self._context = Context(set(observation_keys))
+
+    def __call__(self, _):
+        return self._context
+
+
+# TODO: Should we rename this one?
+class AutoCorrelateConfig(cs.ConfigSuite):
```
xjules, line 307: Hm and why not `AutoScalingConfig`?

semeio/jobs/auto_correlate_observations/config.py
markusdregi: In general I aimed for a very thin abstraction on top of `fcluster` and `linkage` such that the users can be guided to the `scipy` documentation and expect to find the same arguments exposed in the configuration. I could of course remove some of the metrics, but all that would save me was to define the strings... The values is propagated in the same manner no matter if we support only three of them or all.

semeio/jobs/scripts/auto_correlate_observations.py
```diff
@@ -0,0 +1,51 @@
...
+
+from semeio.jobs import auto_correlate_observations
+from semeio.jobs.scripts.correlated_observations_scaling import (
+    CorrelatedObservationsScalingJob,
+)
+
+
+class AutoCorrelateObservationsJob(ErtScript):
+    def run(self, *args):
+        config_record = _fetch_config_record(args)
```
xjules, line 16: Why we need an extra function `_fetch_config_record` as this only loads the file?

semeio/jobs/auto_correlate_observations/job.py
```diff
@@ -0,0 +1,35 @@
+from . import AutoCorrelateConfig
+from semeio.jobs.auto_correlate_observations import ValidationError
+from semeio.jobs.spearman_correlation_job.job import spearman_job
+
+
+def run(auto_correlation_config, measured_data, **records):
```
xjules, line 6: What's records?

semeio/jobs/auto_correlate_observations/config.py
markusdregi: Same argument as above. The `scipy` docs refer to the argument as `t`. I'll write up the explanations and then there will be text explaining the value in the docs. Again, we can change the name and/or behaviour, but then we are not purely exposing the `scipy` parameters anymore. Which is also fine 🤷

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+                _T: {
+                    cs.MetaKeys.Type: cs.types.Number,
+                    cs.MetaKeys.ElementValidators: (
+                        _bounds_validator(lower=0, lower_inclusive=False),
+                    ),
+                },
+                _CRITERION: {
+                    cs.MetaKeys.Type: cs.types.String,
+                    cs.MetaKeys.ElementValidators: (
+                        _one_of(
```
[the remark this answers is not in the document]
markusdregi: Well, I really need all of the members to be strings (as I want to write `inconsistent` nor `0` in the configuration file). Also, I have been thinking about implementing a `one_of` type in `configsuite` and wanted to make a first test of the idea here.

semeio/jobs/auto_correlate_observations/config.py
markusdregi: The separation is purely for readability and specifically to avoid 10 indents in a dictionary. I named it `_SPEARMAN_CORRELATION_SCHEMA` because it represent all the parameters that can be configured for our clustering algorithm based on spearman correlation. Happy to change the name if you have a suggestion :)

semeio/jobs/auto_correlate_observations/config.py
markusdregi: I think we should change the name away from `Auto` and then consistently change these names.

semeio/jobs/auto_correlate_observations/job.py
markusdregi: Nothing besides allowing the general (and non-existent) framework to pass more records then the two we actually need. But I think it is a good idea to remove it until it actually exists 👍

semeio/jobs/scripts/auto_correlate_observations.py
markusdregi: Because I'm designing for replacement and abstraction, not for reuse :) And since I'm later not expecting `args` to point to files, but to contain a stream of data, I've separated that part of the code from the core logic 🤷

semeio/jobs/spearman_correlation_job/job.py
```diff
@@ -4,7 +4,14 @@
 from scipy.cluster.hierarchy import fcluster, linkage
 
 
-def spearman_job(measured_data, threshold):
+def spearman_job(
+    measured_data,
+    t,
```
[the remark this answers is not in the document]
markusdregi: I don't mind. I just changed it consistently to cohere with the `scipy` docs to make the job a really thin wrapper 🤷

We can deviate from this in the code, or also in the UI...

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+
+_SCALING = "scaling"
+_THRESHOLD = "threshold"
+_STD_CUTOFF = "std_cutoff"
+_ALPHA = "alpha"
+_CLUSTERING = "clustering"
+_METHOD = "method"
+_SPEARMAN_CORRELATION = "spearman_correlation"
+_FCLUSTER = "fcluster"
+_T = "t"
```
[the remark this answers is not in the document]
markusdregi: Sure 👍

markusdregi:
> I think this looks good so far. But how will you prevent duplicating the SC and COS config schemas, and all it's validation/transformation logic?
>
> One way to solve this would be to let ACO absorb both SC and COS—make them submodules of the former. The user would not lose the ability to run either SC or COS in isolation.

I'm not certain @jondequinor 🤔 I just wanted to make a first shoot at the UI first and then tackle the rest later. One option is to have ACO supersede both SC and COS, but then the UI will have to be expanded to also allow for manually specifying the groups that are to be scaled. Another option would be to have `SC` and `COS` expose their schemas and then plugg them into the schema of ACO.

semeio/jobs/auto_correlate_observations/config.py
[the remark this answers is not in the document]
xjules: The string is implicitly provided on the fly with `.name` property:
```python
print(Criterion.inconsistent.name)
# "inconsistent"

# validation for string exist
string_exist = "inconsistent" in Criterion.__members__
```

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,315 @@
...
+            _LINKAGE: {_METHOD: _SINGLE, _METRIC: _EUCLIDEAN,},
+        },
+    },
+}
+
+
+class _ObservationContext:
+    def __init__(self, observation_keys):
+        Context = collections.namedtuple("Context", ("observation_keys",))
+        self._context = Context(set(observation_keys))
```
xjules, line 300: Why this needs to be a set? Are we expecting an observation key to appear more than once?

semeio/jobs/spearman_correlation_job/job.py
```diff
@@ -87,6 +96,6 @@ def _calculate_correlation_matrix(data):
     return data.rank().corr(method="pearson")
```
xjules, line 96: Can't find where this data comes from?

tests/jobs/auto_correlate_observations/test_integration.py
```diff
@@ -0,0 +1,159 @@
...
+    obs = ert.getObservations()["FOPR"]
+    for index in [13, 14, 15, 16, 17, 18, 19, 20]:
+        assert obs.getNode(index).getStdScaling() == 2.8284271247461903
+
+
+@pytest.mark.skipif(TEST_DATA_DIR is None, reason="no libres test-data")
+@pytest.mark.usefixtures("setup_tmpdir")
+def test_auto_correlate_skip_clusters_yielding_empty_data_matrixes(monkeypatch):
+    def raising_scaling_job(data):
+        if data == {"CALCULATE_KEYS": {"keys": [{"index": [88, 89], "key": "FOPR"}]}}:
```
xjules, line 102: This one would benefit from a comment or two, since it's hard to see what's happening.

tests/jobs/auto_correlate_observations/unit/test_auto_correlate_observations.py
```diff
@@ -0,0 +1,239 @@
...
+import itertools
+import numpy as np
+import pandas as pd
+import pytest
+import random
+
+from semeio.jobs import auto_correlate_observations
+
+
+class MockedMeasuredData(object):
```
xjules, line 11: Could you make a comment on the data we mock here?

semeio/jobs/auto_correlate_observations/config.py
markusdregi: Nope, it just seemed like an appropriate data structure if iteration and containment testing is what the user of the data structure is supposed to use it for.

tests/jobs/auto_correlate_observations/unit/test_auto_correlate_observations.py
markusdregi: We mock the `MeasuredData` object in ERT. Would you like me to put a reference to it?

semeio/jobs/auto_correlate_observations/config.py
xjules: Lazy to bring up a good name besides `OBSERVATION_CLUSTERING` so we leave at that :)

tests/jobs/auto_correlate_observations/unit/test_auto_correlate_observations.py
```diff
@@ -0,0 +1,239 @@
...
+        pass
+
+    def remove_inactive_observations(self):
+        pass
+
+    def filter_ensemble_std(self, _):
+        pass
+
+
+def assertHomogenClusters(config):
```
xjules, line 68: Why camelCase? Also why this is called homogenCluster?

markusdregi: camelCase due to old habit of following unittest syntax for assert functions... `homogenCluster` because it asserts that all members of each cluster comes from the same polynomial.

semeio/jobs/auto_correlate_observations/config.py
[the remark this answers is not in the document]
markusdregi: I would prefer to not place them in an enum as a container the api in configsuite will only support containers (at least in the first iteration). It is quite a bit more flexible than the limited amounts of types that enums support :)

But I agree that name spacing is good. So what I've done is to collect them into a tuple at least...

semeio/jobs/auto_correlate_observations/config.py
```diff
@@ -0,0 +1,398 @@
...
+                    ).format(maxclust=_MAXCLUST, criterion=_CRITERION),
+                    cs.MetaKeys.ElementValidators: (
+                        _bounds_validator(lower=0, lower_inclusive=False),
+                    ),
+                },
+                _CRITERION: {
+                    cs.MetaKeys.Type: cs.types.String,
+                    cs.MetaKeys.Description: (
+                        "The criterion to use in forming flat clusters. "
+                        "Defaults to {}."
```
ManInFez, line 170: Name or not named parameters for formatting?

markusdregi: I don't have any strong preferences. Do you?

semeio/jobs/auto_correlate_observations/job.py
```diff
@@ -0,0 +1,31 @@
...
+
+def run(auto_correlation_config, measured_data):
+    config = assemble_auto_correlate_config(auto_correlation_config, measured_data,)
+    if not config.valid:
+        raise ValidationError(
+            "Invalid configuration of auto correlation of observations", config.errors
+        )
+
+    config = config.snapshot
+    if config.clustering.method == "spearman_correlation":
```
ManInFez, line 16: should fetch name from config namespace maybe?

semeio/jobs/scripts/auto_correlate_observations.py
```diff
@@ -0,0 +1,75 @@
...
+        config_record, measured_data,
+    )
+    if not config.valid:
+        # The config is loaded by auto_correlate_observations.run first. The
+        # second time should never fail!
+        raise ValueError("AutoCorrelateConfig not valid on second load")
+
+    scale_conf = config.snapshot.scaling
+    return {
+        "threshold": scale_conf.threshold,
```
ManInFez, line 49: fetch strings from config namespace

semeio/jobs/spearman_correlation_job/job.py
```diff
@@ -4,7 +4,14 @@
 from scipy.cluster.hierarchy import fcluster, linkage
 
 
-def spearman_job(measured_data, threshold):
+def spearman_job(
+    measured_data,
+    threshold,
+    criterion="inconsistent",
```
ManInFez, line 10: Perhaps shared information about default values should be accessible for all jobs?

semeio/jobs/scripts/auto_correlate_observations.py
markusdregi: This string actually comes from the COS job and that schema consists of magic strings. So there is no possibility of fetching this without doing some work in the schema. See https://github.com/equinor/semeio/blob/e6a958997703d8bcb2cef3c57de05f8034c87681/semeio/jobs/correlated_observations_scaling/job_config.py#L120

Persist it in an issue instead?

semeio/jobs/spearman_correlation_job/job.py
markusdregi: Yes, as @jondequinor also suggested, they should probably share schema and default values. I can either resolve that here or we can leave that for a follow up, smaler PR 🤷 Let me know what you prefer...

semeio/jobs/scripts/auto_correlate_observations.py
ManInFez: 👍

semeio/jobs/spearman_correlation_job/job.py
ManInFez: make a follow-up issue, and we are golden

markusdregi:
Thanks 👍 I'll change the name before merging it...

markusdregi:
@ManInFez: I've rebased the PR and removed all traces of `Auto Correlate Observations` at the benefit of `Misfit Preprocessor`. I've grep'ed all commits for `auto`, verified that the tests pass on all commits and the line diff count on GitHub is preserved (the last is slightly lucky as I did not have to run black due to too large change in line length). I've also searched for auto here on GitHub in the diff.

I think it is good to go, but I'll await for your agreement before I merge.

ManInFez:
I concur!

Issue #41, opened 2020-02-18
[COS] Improve error handling observation missing


oyvindeide:
Correlated Observations Scaling:

Error handling: better error message when observation specified in group do not exist in dataset

Example:
`['OBS', 'STD'] should be present in DataFrame index, missing: set(['OBS', 'STD'])`

It is not clear how to reproduce this error;
 - what is the "dataset" in this case?
 - observations specified in which group? (input config file?)

jondequinor:
I am going to assume errors on the semeio side of the configuration.

markusdregi:
I think so yes

jondequinor:
I'm unable to reproduce, so I've asked the original reporter for more information.

oyvindeide:
Seems to happen when we are trying to access observational data, but it´s not present. Could be because the obervation is preset, but is filtered out before being accessed.

tralsos:
What I meant is if the user has a typo in the configuration of the CORRELATED_OBSERVATION_SCALING and writes an observation name that does not exist in the obsfile used for the prior run. I can make a test to reproduce it if it doesn't occour for you.

tralsos:
I think this error message only occur if the erroneous key also have a `*` in the name, e.g. I had 
`key : ALB_GIIP*`   but could instead have used `key : ALB_GIIP`
With the typo `key : AL_GIIP*` I get this cryptic error message reported above, while for the typo `key : AL_GIIP`  the error message is 

> "Key: AL_GIIP has no data"

, which is good.

jondequinor:
For the `snake_oil` case, I've tried to reproduce this with the following configuration:
```CALCULATE_KEYS:
  keys: [key: WPR_DIFF_1*]
UPDATE_KEYS:
  keys: [key: WPR_DIFF_1*]
```
I can not reproduce this issue. @tralsos can you provide a test-case? Komodo `testing-py2` should work fine come Monday.

jondequinor:
Reproducible with the following config: 
```
-
    CALCULATE_KEYS:
        keys: [key: AMP*]
-
    CALCULATE_KEYS:
        keys: [key: WWCT_1*]
```
run against e.g. the `snake_oil` observations.

sondreso:
Closed by https://github.com/equinor/semeio/pull/78

PR #74, opened 2020-03-22
Increase modularity of COS

jondequinor:
- Creates an obs_util package, for a lack of a better name. Also splits up test_scaling_job into more fitting constituents.
- Stamps out facade from both COS and SC. This simplifies testing, but also moves both jobs to the same abstraction layer which also eases interoperability.
- SC is now more of a config-generator, and the code reflects this. And SC uses the CorrelatedObservationsScalingJob, which improves encapsulation (SC knows less about COS).
- Tests are mainly moved around. I've removed one test which was not providing any value (that I saw) `test_run_scaling`. Happy to reintroduce, but this path is already well-tested.

### DIFF
--- a/semeio/jobs/correlated_observations_scaling/job.py
+++ b/semeio/jobs/correlated_observations_scaling/job.py
@@ -1,232 +1,99 @@
 # -*- coding: utf-8 -*-
-import fnmatch
-
 import configsuite
-
-from copy import deepcopy
-from collections import namedtuple
-
 from semeio.jobs.correlated_observations_scaling import job_config
-from semeio.jobs.correlated_observations_scaling.scaled_matrix import DataMatrix
-from ert_data.measured import MeasuredData
-from semeio.jobs.correlated_observations_scaling.validator import (
-    valid_configuration,
-    valid_job,
+from semeio.jobs.correlated_observations_scaling.exceptions import ValidationError
+from semeio.jobs.correlated_observations_scaling.obs_utils import (
+    create_active_lists,
+    find_and_expand_wildcards,
 )
-from res.enkf import LocalObsdata, ActiveList
-
-from ecl.util.util import BoolVector
-from res.enkf import RealizationStateEnum
+from semeio.jobs.correlated_observations_scaling.scaled_matrix import DataMatrix
+from semeio.jobs.correlated_observations_scaling.validator import has_keys, is_subset
 
 
-def scaling_job(facade, user_config_dict):
-    """
-    Takes an instance of EnkFMain and a user config dict, will do some pre-processing on
-    the user config dict, set up a ConfigSuite instance and validate the job before control
-    is passed to the main job.
-    """
+class ScalingJob(object):
+    def __init__(self, obs_keys, obs, obs_with_data, user_config_dict):
+        """Creates a ScalingJob instance with the given obs_keys, obs,
+        obs_with_data and user_config_dict."""
+        self._obs = obs
+        self._obs_with_data = obs_with_data
+        self._obs_keys = obs_keys
+        self._config = self._setup_configuration(user_config_dict)
+        self._validate()
 
-    observation_keys = [
-        facade.get_observation_key(nr) for nr, _ in enumerate(facade.get_observations())
-    ]
-    obs_with_data = keys_with_data(
-        facade.get_observations(),
-        observation_keys,
-        facade.get_ensemble_size(),
-        facade.get_current_fs(),
-    )
-    config_dict = _find_and_expand_wildcards(observation_keys, user_config_dict)
+    def scale(self, measured_data):
+        """
+        Collects data, performs scaling and applies scaling, assumes validated input.
+        """
+        config = self._config.snapshot
 
-    config = setup_configuration(config_dict, job_config.build_schema())
+        measured_data.remove_failed_realizations()
+        measured_data.remove_inactive_observations()
+        measured_data.filter_ensemble_mean_obs(config.CALCULATE_KEYS.alpha)
+        measured_data.filter_ensemble_std(config.CALCULATE_KEYS.std_cutoff)
 
-    if not valid_configuration(config):
-        raise ValueError("Invalid configuration")
-    if not valid_job(config, observation_keys, obs_with_data):
-        raise ValueError("Invalid job")
-    _observation_scaling(facade, config.snapshot)
+        matrix = DataMatrix(measured_data.data)
+        matrix.std_normalization(inplace=True)
 
+        scale_factor = matrix.get_scaling_factor(config.CALCULATE_KEYS)
 
-def _observation_scaling(facade, config):
-    """
-    Collects data, performs scaling and applies scaling, assumes validated input.
-    """
-    calculate_keys = [event.key for event in config.CALCULATE_KEYS.keys]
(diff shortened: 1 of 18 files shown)


semeio/jobs/scripts/spearman_correlation.py
```diff
@@ -1,19 +1,36 @@
...
+        measured_data = MeasuredData(facade, obs_keys)
+
         parser = spearman_job_parser()
         args = parser.parse_args(args)
 
-        spearman_job(facade, args.threshold, args.dry_run)
+        scaling_configs = spearman_job(measured_data, args.threshold)
+
+        if not args.dry_run:
+            for config in scaling_configs:
```
markusdregi: I belive we could also pass a single configuration listing the different scalings that are to be done. Any particular reason you choose to pass them one by one?

semeio/jobs/correlated_observations_scaling/job.py
```diff
@@ -1,232 +1,104 @@
...
+        errors.extend(is_subset(calc_keys, application_keys))
+        errors.extend(
+            has_keys(self._obs_keys, calc_keys, "Key: {} has no observations")
+        )
+        errors.extend(has_keys(self._obs_with_data, calc_keys, "Key: {} has no data"))
+        self._validation_errors = errors
+        return len(self._validation_errors) == 0
+
+    def get_errors(self):
+        if self._validation_errors is None:
```
markusdregi: This seems slightly off to me. Why should I call `valid` before `get_errors`. I would suggest the following; we implement to properties `valid` and `errors` and ensure validation before these are called. We can either do this at creation time (I don't really see why we don't want to do that) and compute `_errors` then. Or we can lazily compute the errors the first time either `valid` or `errors` is called. But the later discussion is an implementation detail really...

tests/jobs/spearman_correlation_job/test_integration.py
```diff
@@ -48,24 +29,38 @@ def test_main_entry_point_gen_data(monkeypatch):
...
+
+    shutil.copytree(test_data_dir, "test_data")
+    os.chdir(os.path.join("test_data"))
+
+    res_config = ResConfig("snake_oil.ert")
+
+    ert = EnKFMain(res_config)
+    sc.SpearmanCorrelationJob(ert).run(*["-t", "1.0"])
+
+    # FIXME: assertsion please
```
markusdregi: .....

tests/jobs/correlated_observations_scaling/unit/test_config.py
```diff
@@ -0,0 +1,113 @@
...
+def test_valid_configuration():
+    valid_config_data = {
+        "CALCULATE_KEYS": {"keys": [{"key": "POLY_OBS"}]},
+        "UPDATE_KEYS": {"keys": [{"key": "POLY_OBS"}]},
+    }
+
+    schema = build_schema()
+    config = configsuite.ConfigSuite(valid_config_data, schema)
+
+    # assert validator.valid_configuration(config)
```
markusdregi: ...

markusdregi:
The changes in the tests are rather large. Could you comment on whether you have mostly moved them around or actually changed them as well?

semeio/jobs/scripts/spearman_correlation.py
jondequinor: No, just went with what was there. Good call.

semeio/jobs/correlated_observations_scaling/job.py
jondequinor: Yeah no real reason. Will fix.

tests/jobs/spearman_correlation_job/test_integration.py
jondequinor: I had forgot/suppressed this. I’ll add the assertions. 

And you should start using … like god intended

jondequinor:
> The changes in the tests are rather large. Could you comment on whether you have mostly moved them around or actually changed them as well?

Yes, will do.

semeio/jobs/scripts/spearman_correlation.py
jondequinor: Wait, are you talking about multiple configs inside one config? I'd rather not introduce that as it seems buggy (ref https://github.com/equinor/semeio/issues/41).

I don't have to do a generator though, will return a list of configs.

markusdregi: >  I'd rather not introduce that as it seems buggy

You would not introduce it, as the job already supports it (at least in the yaml, but not in if passed as a programmatic object). I think we should either utilise it or remove the feature...

markusdregi: I have nothing against the generator ;)

tests/jobs/spearman_correlation_job/test_integration.py
markusdregi: I just did not know what to say, the `FIXME` already screamed it loud enough. Can I add empty comments though?

markusdregi: I cannot 😭

semeio/jobs/correlated_observations_scaling/job.py
jondequinor: This is my bad, it already _did_ raise an exception in the _initializer_ (just a function). However, I refactored this because I didn't want the `ScalingJob` to decide what to do with the errors. It used to be printed right in the `valid_job` function, but it needs to be propagated to the top. There it can be printed or whatever.

I will re-introduce the exception, but allow the `errors` to be carried on that exception (so a subclass of `ValueError` called `ValidationError`. OK?

markusdregi: I don't think the two are exclusive are they? I do however agree to leave it to the caller to deal with the errors occurring. Whether you introduce an exception or error code convention for propagating failures I leave to you 👍

tests/jobs/spearman_correlation_job/test_integration.py
jondequinor: damn m$

semeio/jobs/scripts/spearman_correlation.py
jondequinor: It is already used by a site, but okay, let's move to multiple configs in a config.

(Aside) Any reason for not using yaml streams here?

markusdregi: Or we revert to only doing a single one in the same config?

jondequinor:
test this please

jondequinor:
test this please

markusdregi:
"Restarted" the PR to get Travis going 👍

semeio/jobs/correlated_observations_scaling/job.py
```diff
@@ -1,232 +1,100 @@
...
+
+
+class ScalingJob(object):
+    def __init__(self, obs_keys, obs, obs_with_data, user_config_dict):
+        """Creates a ScalingJob instance with the given obs_keys, obs,
+        obs_with_data and user_config_dict."""
+        self._obs = obs
+        self._obs_with_data = obs_with_data
+        self._obs_keys = obs_keys
+        self._validation_errors = None
```
markusdregi: This is redundant now

jondequinor:
Thanks for the review. I second pylinting.

PR #77, opened 2020-03-25
Document Correlated Observation Scaling (COS) job, suggest docs structure

jondequinor:
- Add argparse to `COS`
- Add sphinx docs that include this argparse, and in detail documents the configuration format

I don't think this is mergeable, because there's an ongoing discussion on what to do with the ert/semeio (and other repos of jobs) documentation combo.

I've argued dumping the rst-files from semeio somewhere in ERT in the sib-docs script, others that the combo should be drawn up using python.

Maybe the PCA team can take the discussion further. Either way, I think there's some value in the COS docs themselves.

### DIFF
--- a/docs/correlated_observation_scaling.rst
+++ b/docs/correlated_observation_scaling.rst
@@ -1,0 +1,133 @@
+Correlated Observations Scaling
+===============================
+
+Configuration
+-------------
+
+Given a configuration in the format:
+
+.. code-block:: salt
+
+    CALCULATE_KEYS:
+        keys:
+            - key: FOPR
+
+This configuration will let COS calculate a scaling factor from all data
+points in ``FOPR`` and update all data points in ``FOPR``.
+You can also specify which indices will be updated on the ``FOPR`` key:
+
+Wildcards
+^^^^^^^^^
+
+Keys can be given as wildcards:
+
+.. code-block:: salt
+
+    CALCULATE_KEYS:
+        keys:
+            - key: WOPR_OP*
+
+This will calculate a scaling factor for all keys matching ``WOPR_OP*`` where
+the asteriks will match everything. The patterns are Unix style:
+
+::
+
+    *       matches everything
+    ?       matches any single character
+    [seq]   matches any character in seq
+    [!seq]  matches any char not in seq
+
+
+
+Key indices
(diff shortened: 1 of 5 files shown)


docs/correlated_observation_scaling.rst
```diff
@@ -0,0 +1,133 @@
...
+
+This will calculate a scaling factor for all keys matching ``WOPR_OP*`` where
+the asteriks will match everything. The patterns are Unix style:
+
+::
+
+    *       matches everything
+    ?       matches any single character
+    [seq]   matches any character in seq
+    [!seq]  matches any char not in seq
```
oyvindeide: Great that you´ve added documentation! Though this is not strictly correct, only strings with `*` will actually be evaluated by the expanding function. That being said, I´m in favor of keeping the documentation as is, and fixing the code, but that might belong in a separate issue and is outside the scope of this PR.

jondequinor: Ah, I assumed it used `fnmatch`? Why just a subset?

oyvindeide: It uses `fnmatch`, but https://github.com/equinor/semeio/blob/9861ce8bc5780711df3809f3a028a77f7e2cef9c/semeio/jobs/correlated_observations_scaling/obs_utils.py#L39 means it might not hit `fnmatch`, which is what I meant should be changed.

docs/correlated_observation_scaling.rst
```diff
@@ -0,0 +1,133 @@
+Correlated Observations Scaling
+===============================
+
+Configuration
```
lars-petter-hauge, line 4: Configsuite includes `rst` generation functionality based on the content in the schema and also has a sphinx extension that would generate an html page with a slightly more interactive setup.

I would want to point to the documenation page, but it seems even the `latest` does not include the last commit (which is the one containing the documentation).

In particular though, the extension is a bit rough (or the product of the extension is rough). But I think it would be beneficial spending time improving that part instead. We will need it for later (and several jobs in spinningjenny will greatly benefit from an improved extension)

oyvindeide:
Surpassed by https://github.com/equinor/semeio/pull/124 and #168

PR #78, opened 2020-03-25
Fail if CALC key wild cards produce empty config

jondequinor:
The empty CALCULATE_KEYS config has special meaning: it means use all
keys. This is most likely not what the user wants. No matches for a
wildcard has to be assumed to be a user error, and the scaling job
should therefore fail.

### DIFF
--- a/semeio/jobs/correlated_observations_scaling/exceptions.py
+++ b/semeio/jobs/correlated_observations_scaling/exceptions.py
@@ -6,3 +6,10 @@ class ValidationError(ValueError):
     def __init__(self, message, errors):
         super(ValidationError, self).__init__(message)
         self.errors = errors
+        self.message = message
+
+    def __str__(self):
+        msg = "{}\n".format(self.message)
+        for error in self.errors:
+            msg += "\t{}".format(error)
+        return msg
--- a/semeio/jobs/correlated_observations_scaling/obs_utils.py
+++ b/semeio/jobs/correlated_observations_scaling/obs_utils.py
@@ -4,6 +4,7 @@ from copy import deepcopy
 
 from ecl.util.util import BoolVector
 from res.enkf import ActiveList, LocalObsdata, RealizationStateEnum
+from semeio.jobs.correlated_observations_scaling.exceptions import ValidationError
 
 
 def _wildcard_to_dict_list(matching_keys, entry):
@@ -34,14 +35,23 @@ def find_and_expand_wildcards(obs_list, user_dict):
     new_dict = deepcopy(user_dict)
     for main_key, value in user_dict.items():
         new_entries = []
+        non_matching_wildcards = []
         if main_key in ("UPDATE_KEYS", "CALCULATE_KEYS"):
             for val in value["keys"]:
                 if "*" in val["key"]:
-                    new_entries.extend(_expand_wildcard(obs_list, val, val["key"]))
+                    expanded = _expand_wildcard(obs_list, val, val["key"])
+                    if len(expanded) == 0:
+                        non_matching_wildcards.append(val["key"])
+                    new_entries.extend(expanded)
                 else:
                     new_entries.append(val)
             new_dict[main_key]["keys"] = new_entries
 
+            if len(non_matching_wildcards) > 0:
+                raise ValidationError(
+                    "Invalid {}".format(main_key),
+                    ["{} had no match".format(wc) for wc in non_matching_wildcards],
+                )
     return new_dict
 
 
--- a/tests/jobs/correlated_observations_scaling/test_integration.py
+++ b/tests/jobs/correlated_observations_scaling/test_integration.py
@@ -4,8 +4,8 @@ import shutil
 import numpy as np
 import pytest
 import yaml
-
 from res.enkf import EnKFMain, ResConfig
+
 from semeio.jobs.scripts.correlated_observations_scaling import (
     CorrelatedObservationsScalingJob,
 )
--- a/semeio/jobs/scripts/correlated_observations_scaling.py
+++ b/semeio/jobs/scripts/correlated_observations_scaling.py
@@ -5,7 +5,6 @@ import yaml
 from ert_data.measured import MeasuredData
 from ert_shared.libres_facade import LibresFacade
 from res.enkf import ErtScript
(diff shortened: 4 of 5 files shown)


tests/jobs/correlated_observations_scaling/test_integration.py
```diff
@@ -173,6 +174,28 @@ def test_main_entry_point_summary_data_calc():
...
+
+    test_data_dir = os.path.join(TEST_DATA_DIR, "local", "snake_oil")
+
+    shutil.copytree(test_data_dir, "test_data")
+    os.chdir(os.path.join("test_data"))
+    res_config = ResConfig("snake_oil.ert")
+    ert = EnKFMain(res_config)
+
+    try:
+        CorrelatedObservationsScalingJob(ert).run(cos_config)
```
markusdregi: Should this not result in a `ValidationError`?

oyvindeide:
This could potentially also be solved in the injection of default values, which is the reason for this not failing like it should. `CALCULATE_KEYS` is required for this very reason, but the injection of default values means that validation is null and void.

jondequinor:
Okay, thanks for the clarification re: empty `keys`. However, `CALCULATE_KEYS` is not empty, it's the `keys` that is empty.

Anyway, the configuration is valid, there are keys, but they do not expand to anything that exists.

This validation should exist in the application, not configsuite. I.e. after config validation.

That means a potential large refactor where you do as little as possible with the config, but create an intermediary between the config and the scaling. @markusdregi @oyvindeide what do you think?

semeio/jobs/correlated_observations_scaling/exceptions.py
```diff
@@ -6,3 +6,9 @@ class ValidationError(ValueError):
     def __init__(self, message, errors):
         super(ValidationError, self).__init__(message)
         self.errors = errors
+
+    def __str__(self):
+        msg = "{}\n".format(str(super(self)))
```
oyvindeide: This gives some slightly non standard behavior, so can´t for example check the error message when running pytest in the standard way:

```
with pytest.raises(ValueError) as err_msg:
    raise ValueError("my error")
```
Then `str(err_msg.value)` should give: `my error`, but with ValidationError it throws an error.

semeio/jobs/correlated_observations_scaling/obs_utils.py
```diff
@@ -42,6 +43,16 @@ def find_and_expand_wildcards(obs_list, user_dict):
...
 
+        # CALCULATE_KEYS.keys went from having some values, to having no values
+        if (
+            main_key == "CALCULATE_KEYS"
+            and len(value["keys"]) > 0
+            and len(new_entries) == 0
+        ):
+            raise ValidationError(
+                "Invalid configuration",
+                ["CALCULATE_KEYS {} matched no keys".format(value["keys"])],
```
oyvindeide: What if I have some valid and some invalid keys, then this will tell me all of them are invalid?

oyvindeide: Think it would be good to add a test for that, to check that only the invalid keys are in the error message.

markusdregi:
> That means a potential large refactor where you do as little as possible with the config, but create an intermediary between the config and the scaling. @markusdregi @oyvindeide what do you think?

I would prefer that the world view of the executor of the configuration and the user is the same and that whatever needs to be processed or translated in-between is as transparent as possible. I agree with the other comments of @oyvindeide and suggest that we merge this fix first and then make an evaluation of the strategy forward. That will also make your post-rotation burden much smaller @jondequinor (given that you're willing to fix the current comments of @oyvindeide that is. If not, we'll fix it in the squad) ;)

semeio/jobs/correlated_observations_scaling/obs_utils.py
jondequinor: I've added a case for this (I think), but if there's a non-wildcard key there, this will not be thrown. If there's a wildcard in there that will match, this will not be thrown.

Let me know if there are additional things you wanted to test.

markusdregi: I think maybe what @oyvindeide is referring to here is a case where one of the wildcard keys has a match and another key does not? I would be willing to argue that this should also be considered invalid and then adding the test @oyvindeide refers to would make sense. Am I correct Øyvind?

But that could potentially be resolved in a separate PR also...

oyvindeide: Both, in the first iteration any valid and invalid key would give both as invalid. But I have no problem with creating an issue for the remaining 👍

markusdregi: Yes, so then you would expect the keys `FOPT, TYPO*` to yield an error right?

semeio/jobs/correlated_observations_scaling/obs_utils.py
```diff
@@ -34,14 +35,24 @@ def find_and_expand_wildcards(obs_list, user_dict):
...
     for main_key, value in user_dict.items():
         new_entries = []
+        non_matching_wildcards = []
         if main_key in ("UPDATE_KEYS", "CALCULATE_KEYS"):
             for val in value["keys"]:
                 if "*" in val["key"]:
-                    new_entries.extend(_expand_wildcard(obs_list, val, val["key"]))
+                    expanded = _expand_wildcard(obs_list, val, val["key"])
+                    if len(expanded) == 0:
+                        non_matching_wildcards.append(val["key"])
```
xjules: If there is a `non_matching_wildcards` entry shouldn't it trigger `ValidationError` too?

jondequinor: I agree. Done!

xjules:
Test this please!

jondequinor:
test this please

jondequinor:
(not sure what failed, but the workspace went missing)

markusdregi:
Jenkins: Test this please

markusdregi:
> (not sure what failed, but the workspace went missing)

I deleted everything in `/tmp` and restarted the workers at multiple occasions. So it might have been me...

PR #83, opened 2020-04-14
Add tox

oyvindeide:


markusdregi:
This looks promising 👍

oyvindeide:
> This looks promising 👍

Well... Now it works - but it´s not very pretty. We end up passing on a bunch of packages into the virtualenv created by `tox`, which kind of defies the point a little bit. It will be much prettier once `libecl` and `ert` can be pip installed from `pypi`. That being said, I like the advantage of splitting out the `pylint` and `black` parts. There are not that many drawbacks compared with the current implementation I think.

oyvindeide:
It fails now because the linting isn´t passing, but can leave that until you chime in what you think before I spend more time.

oyvindeide:
I cleaned this up and removed everything related to linting, so if you agree with this change @markusdregi I propose we merge this, then you can finish up your linting PR. To run linting with `tox`, add this to `tox.ini`:
```
[testenv:linting]
description = Invoke pylint to lint the code
deps =
    pylint
    {[testenv]deps}
commands = pylint semeio
```
and the `linting` command to `envlist`, as well as this to `travis.yml`:
```
    - python: 3.7
      env: TOXENV=linting
```

oyvindeide:
test this please

PR #84, opened 2020-04-14
Porting of old csv_export1

oysteoh:


markusdregi:
I think we should port the tests also...

oysteoh:
> I think we should port the tests also...

that is on my agenda, will update the pr as soon as i have those running :)

oysteoh:
closing pr for now, not an easy way to get this to fly on python3. we'll try other approaches first

PR #86, opened 2020-04-21
[WIP] Add a reporter for messaging

### LINKED ISSUE #82: Use higher level interface for data/messages to/from jobs
The content in this issue is a complete copy of the content from https://github.com/equinor/semeio/pull/66
We've agreed that this is a good approach and fits nicely with the roadmap of ert.

--------------------------------------------------

**Suggested `Reporter`**:

Principles:

Data flowing out of workflows

- into other workflows 
- to the user

Arguably fixes #45 #43 #42

lars-petter-hauge:
Put the PR up so it's easier to discuss.

This is merely a dumb implementation which will require iterations. It does solve the first problem though.

I will add couple comments around as well

First iteration of #82

### DIFF
--- a/semeio/jobs/correlated_observations_scaling/scaled_matrix.py
+++ b/semeio/jobs/correlated_observations_scaling/scaled_matrix.py
@@ -3,15 +3,17 @@ from copy import deepcopy
 import numpy as np
 
 from semeio.jobs.correlated_observations_scaling.exceptions import EmptyDatasetException
+from semeio.reporter import Reporter
 
 
 class DataMatrix(object):
-    def __init__(self, input_data):
+    def __init__(self, input_data, reporter):
         """
             Takes input data in the form of a Pandas multi index dataframe with observations,
         standard deviation and simulated values. Assumes observations are
         prepended with _OBS and standard deviation with _STD.
         """
+        self.reporter = reporter
         self.data = input_data
         if input_data.shape[1] == 0:
             raise EmptyDatasetException("Empty dataset")
@@ -52,9 +54,13 @@ class DataMatrix(object):
         )
         nr_observations = data_matrix.shape[1]
 
-        print("Scaling factor calculated from {}".format(events.keys))
+        self.reporter.report(
+            value="Scaling factor calculated from {}".format(events.keys)
+        )
 
-        return self._calculate_scaling_factor(nr_observations, nr_components)
+        return self._calculate_scaling_factor(
+            nr_observations, nr_components, self.reporter
+        )
 
     def _get_data(self):
         return self.data[~self.data.index.isin(["OBS", "STD"])]
@@ -86,7 +92,7 @@ class DataMatrix(object):
         return len([1 for i in variance_ratio[:-1] if i < threshold]) + 1, s
 
     @staticmethod
-    def _calculate_scaling_factor(nr_observations, nr_components):
+    def _calculate_scaling_factor(nr_observations, nr_components, reporter):
         """
         Calculates a observation scaling factor which is:
             sqrt(nr_obs / pc)
@@ -95,8 +101,8 @@ class DataMatrix(object):
             pc is the number of primary components from PCA analysis
                 below a user threshold
         """
-        print(
-            "Calculation scaling factor, nr of primary components: {:d}, number of observations: {:d}".format(
+        reporter.report(
+            value="Calculation scaling factor, nr of primary components: {:d}, number of observations: {:d}".format(
                 nr_components, nr_observations
             )
         )
--- a/semeio/reporter/reporter.py
+++ b/semeio/reporter/reporter.py
@@ -1,0 +1,32 @@
+class Reporter:
+    def __init__(self, name):
+        self.name = name
+
+    def _get_suffix_for_value(self, value):
+
+        if isinstance(value, (str, float, int)):
+            return "txt"
+
+        valid_types = ",".join(["string", "float", "int"])
+
+        raise TypeError(
+            "The type of value <{}> must be one of: {} was {}".format(
+                value, valid_types, type(value)
+            )
+        )
+
+    def report(self, key=None, value=None, write=False):
+        if key is None and write:
+            raise ValueError("key must be specified if write is True, was None")
+
+        if write:
+            filename = "{job_name}_{key}.{suffix}".format(
+                job_name=self.name,
+                key=str(key),
+                suffix=self._get_suffix_for_value(value),
+            )
+
+            with open(filename, "w") as fh:
+                fh.write(str(value))
+        else:
+            print(value)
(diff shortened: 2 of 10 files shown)


semeio/jobs/correlated_observations_scaling/scaled_matrix.py
```diff
@@ -3,15 +3,17 @@
 import numpy as np
 
 from semeio.jobs.correlated_observations_scaling.exceptions import EmptyDatasetException
+from semeio.reporter import Reporter
 
 
 class DataMatrix(object):
-    def __init__(self, input_data):
+    def __init__(self, input_data, reporter):
```
lars-petter-hauge, line 10: I think it is silly that a `DataMatrix` should be initialized with a `reporter`. I wonder if `Reporter` should be implemented like `logging`?

semeio/reporter/reporter.py
```diff
@@ -0,0 +1,32 @@
+class Reporter:
+    def __init__(self, name):
+        self.name = name
+
+    def _get_suffix_for_value(self, value):
```
lars-petter-hauge, line 5: Not so certain this is the best way of doing it..

The idea is that we could be able to separate write implementations for `dict`, `numpy`, `pd.DataFrame` etc. But it very quickly becomes a mess.. and really gold plating

semeio/jobs/correlated_observations_scaling/scaled_matrix.py
markusdregi: I think that is a good observation ;)

markusdregi: I would say that if `DataMatrix` wants to post structured data, then it should not be allowed to (and we would have to pass the `reporter` as you've done). Then I think we should redesign the code.

However, for general logging statements it could be an idea to register a listener to the Python logger and hence fetch all logging statements and pass them on via the reporter 🤷

lars-petter-hauge:
I did some basic testing when looking at this. The way we're executing external workflows is to call `subprocess.popen` (https://github.com/equinor/libres/blob/2681278831b64b9775b57b155badae9f131ff865/python/res/job_queue/external_ert_script.py#L19)

**Observation:** whatever is added to the `logger` module will always be caught in the `stderr` part of `Popen.communicate`. 

Example:
`runnable_script.py`:
```
import logging

logging.basicConfig(level=logging.INFO)
logger = logging.getLogger("logger_name")

if __name__ == '__main__':
    print("testing print")
    logger.warning("logged warning")
    logger.info("logged info")
    logger.debug("logged debug")
```

execute with `subprocess.Popen` (similar to what `libres` does)

```
from subprocess import Popen, PIPE

command = ["python", "runnable_script.py"]

job = Popen(command,
            stdout = PIPE,
            stderr = PIPE)

stdoutdata, stderrdata = job.communicate()

# <<< stdoutdata
# >>> b'testing print\n'
# <<< stderrdata
# >>> b'WARNING:logger_name:logged warning\nINFO:logger_name:logged info\n'
```

I found that behaviour first of a bit weird, but second this means there's likely a bug in how ert displays the stdout from external workflows.

Having the `runnable_script.py` exit with `sys.exit(1)` does not alter the result in any way.

Including logging to `logger.error()` appends the result on the `stderrdata` (not surprising..)

semeio/reporter/reporter.py
```diff
@@ -0,0 +1,32 @@
...
+
+        valid_types = ",".join(["string", "float", "int"])
+
+        raise TypeError(
+            "The type of value <{}> must be one of: {} was {}".format(
+                value, valid_types, type(value)
+            )
+        )
+
+    def report(self, key=None, value=None, write=False):
```
markusdregi, line 18: I think deducing data types will get really hairy. I also don't think it should be on the caller to toggle a `write`-flag.

I suggest that in the first iteration we support:
```python
Reporter.report(self, value: str)
Reporter.publish(self, key: str, value: json)
```

Then we immediately wrap the Python logger (with log-level `info`) around `Reporter.report` and let all `report` calls be appended in a file on disk with name `<reporter_namespace>/report`.

Second, we use the `publish` method to persist structured data as json. We can later extend with `publish_matrix` etc, but I suggest that the default structured data format is json. Each of these calls will in the first iteration induce a dump of the value to `<reporter_namespace>published/<key>.json`.

We need to find a reasonable base-folder for the data dumping. One option is to inject `ENSPATH` into the job description files and use `<enspath>/reporter` as base. This is at least a location the user is expecting output to.

markusdregi: We can extend upon this later, but this will be sufficient for our needs now 👍

lars-petter-hauge: > I think deducing data types will get really hairy. 

I agree, I also thought this was a poor decision.

> I also don't think it should be on the caller to toggle a write-flag.

Unsure.. As long as everything is written to disk as suggested, then of course it doesn't matter. But if we get in a scenario where it would be beneficial to only store some data _somewhere_, then I think it should be up to the caller to decide if the option is available.

Just note that if we decide to go with write-to-disk as default, we would have to keep that as default unless we want to break the API..

lars-petter-hauge:
> However, for general logging statements it could be an idea to register a listener to the Python logger and hence fetch all logging statements and pass them on via the reporter :shrug: 

I can't see how we are to register a listener to the python logger. But I do have a suggestion though.

Facilitate a log handler from ert. In the case where the application already uses `logging` from python, the only necessary thing to do would be to import the handler from ert and add it to the logger at the highest level. This will be very little invasive. If we want the user to instead import a `Reporter` from ert, and change all their calls to `Reporter.report`, that would demand a much larger change in their code. 

On our side we defined how the handler emit in the different cases of `gui` and `cli`. 

Haven't made a POC for this, and I haven't used handlers as much, so not sure how well it works just yet.

This would only apply for `stdout` type messages, and not structured data I believe. The data type that gets into the handler is already pure a bytes type.

semeio/reporter/reporter.py
markusdregi: > Unsure.. As long as everything is written to disk as suggested, then of course it doesn't matter. But if we get in a scenario where it would be beneficial to only store some data somewhere, then I think it should be up to the caller to decide if the option is available.

But the point is that the job does not know whether the data is important or not. Because it is dependent on the setting for which it runs in. Hence, the job should only serve the data and the option of keeping it or not should be configured further up in the call stack (preferably to configuration file).

markusdregi: > Just note that if we decide to go with write-to-disk as default, we would have to keep that as default unless we want to break the API..

That is an experimental feature that we will break as soon as we can. When this data goes into the database we will immediately stop dumping then to files...

markusdregi:
> I can't see how we are to register a listener to the python logger. But I do have a suggestion though.

By using the `logging.addHandler`?

lars-petter-hauge:
Right, the developers of the workflows has to use the `logging.addHandler`. If that was your suggestion then I'm completely on board. And we'll provide the `Handler`.

I wondered if you suggested that we somehow injected a listener actively from our side, without any changes from the workflow developers point of view.

markusdregi:
Well, I was hoping that we could add the handler as a utility and that the workflow developer only had to use the logging module. It might be that we need to persist the default logger name in the workflow, but it would be preferable if we instead could set the default logger to be a specific logger...

PR #87, opened 2020-04-23
Add simple report function

dotfloat:
This PR provides a `report(key, val)` function that creates a file with `key` as basename and appropriate extension for the type of `val`. If `val` is a numpy array then it is written to cwd as `{key}.csv` using pandas. If `val` is representable as JSON, then a `{key}.json` is created.

This is a dead stupid change, but it can serve as foundation for future work. I believe having two APIs for reporting is necessary:
1. Internal Python jobs that use `ErtScript` would need to have their own `Reporter` object with added context about the workflow (ie. unique identifier). This object must be created and managed by libres.
2. External Python jobs can `from semeio.reporter import report` and then use it as is. This would behave like `logging.info`, where it is initialised with sensible defaults, but you can override the behaviour. For example, on first `report` invocation, a connection to ERT will be attempted, but unit tests may change the report handler and mock the connection.

I do not think this PR belongs in `semeio`, however. It should be in its own repository or libres.

---
## Logger
POSIX mandates only two output streams `stdout` and `stderr`, thus Python's logging stuff isn't visible from the outside world. For external Python workflow we should provide a convenience function that installs the necessary handlers that `report` stuff to ERT. Thus, there can be three types of external workflows:
1. Compliant: ERT sees both `stdout`, `stderr` and a nicely formatted `logging`.
2. Non-compliant: ERT sees only `stdout`, `stderr`. Python's `logging` writes to `stdout` by default.
3. Very non-compliant: Python's `logging` was told to write to `stdout.log`. ERT sees nothing. We tell our users to please stop.

The logger stream can be smarter than a simple byte stream. We could, for example, mandate each message is a single UTF-8 line, with a timestamp and log level. Thus it'd be possible to show dates for each line and highlight warnings in the GUI.

I've prototyped streaming data via ZeroMQ and similar reporter, and wouldn't be difficult to do with a bit more infrastructure:
https://github.com/equinor/ert-msgbus/blob/fc867325e1acb607094890e1cea0e9285dead64c/ert_msgbus/service/reporter.py#L9
https://github.com/equinor/ert-msgbus/blob/fc867325e1acb607094890e1cea0e9285dead64c/ert_msgbus/service/message.py#L27

---
## Yet another call to use UNIX sockets instead of TCP in the future for external workflows
I'll beat this dead horse some more. However the communication channel will look like, I think it's important that our first attempt will be to use UNIX sockets instead of trying to do TCP/HTTP:
1. It is trivial to add security so that only the current user's processes can use this channel. No need for authentication, SSL or crypto.
2. SSH can forward UNIX sockets: https://medium.com/@dperny/forwarding-the-docker-socket-over-ssh-e6567cfab160
3. It is more performant than TCP since it avoids the entire network stack

ZeroMQ supports both in-process, unix and tcp sockets. If we choose to go with ZeroMQ then this is trivial. But if we choose to write our own, I think UNIX is the correct decision.

### DIFF
--- a/semeio/reporter/reporter.py
+++ b/semeio/reporter/reporter.py
@@ -1,0 +1,29 @@
+from .serialize_numpy import serialize_numpy
+from .serialize_json import serialize_json
+
+
+class Reporter(object):
+    _serializers = [serialize_numpy, serialize_json]
+
+    def __init__(self):
+        pass
+
+    def report(self, key, val):
+        if not isinstance(key, str):
+            raise TypeError("Report key must be of type str")
+        if key == "" or "/" in key:
+            raise ValueError("Invalid report key")
+
+        for ser in self._serializers:
+            if ser(key, val):
+                return
+        raise TypeError(
+            "Reporter cannot serialize values of type '{}'".format(type(val).__name__)
+        )
+
+
+_reporter = Reporter()
+
+
+def report(key, val):
+    _reporter.report(key, val)
--- a/semeio/reporter/reporter.py
+++ b/semeio/reporter/reporter.py
@@ -1,9 +1,12 @@
-from .serialize_numpy import serialize_numpy
-from .serialize_json import serialize_json
+import re
+
+from semeio.reporter.serialize_numpy import serialize_numpy
+from semeio.reporter.serialize_json import serialize_json
 
 
 class Reporter(object):
     _serializers = [serialize_numpy, serialize_json]
+    _valid_key = re.compile("^[0-9a-zA-Z-_+.]+$")
 
     def __init__(self):
         pass
@@ -11,8 +14,8 @@ class Reporter(object):
     def report(self, key, val):
         if not isinstance(key, str):
             raise TypeError("Report key must be of type str")
-        if key == "" or "/" in key:
-            raise ValueError("Invalid report key")
+        if key.match(self._valid_key):
+            raise ValueError("Report key {} malformed: May only contain alphanumerics, '_', '-', '+', '.'", repr(key))
 
         for ser in self._serializers:
             if ser(key, val):
--- a/semeio/reporter/serialize_json.py
+++ b/semeio/reporter/serialize_json.py
@@ -1,0 +1,10 @@
+import json
+
+
+def serialize_json(key, val):
+    try:
+        with open("{}.json".format(key), "w") as f:
+            json.dump(val, f)
+    except:
+        return False
+    return True
--- a/semeio/reporter/serialize_json.py
+++ b/semeio/reporter/serialize_json.py
@@ -5,6 +5,10 @@ def serialize_json(key, val):
     try:
         with open("{}.json".format(key), "w") as f:
             json.dump(val, f)
-    except:
+    except OverflowError:
+        # Happens when there is a cycle
+        return False
+    except TypeError:
+        # Unserialisable type
         return False
     return True
--- a/semeio/reporter/__init__.py
+++ b/semeio/reporter/__init__.py
@@ -1,1 +1,1 @@
-from .reporter import report
+from semeio.reporter.reporter import report
--- a/semeio/reporter/__init__.py
+++ b/semeio/reporter/__init__.py
@@ -1,0 +1,1 @@
+from .reporter import report
--- a/semeio/reporter/serialize_numpy.py
+++ b/semeio/reporter/serialize_numpy.py
@@ -1,0 +1,10 @@
+import numpy as np
+import pandas as pd
+
+
+def serialize_numpy(key, val):
+    if not isinstance(val, np.ndarray):
+        return False
+
+    pd.DataFrame(val).to_csv("{}.csv".format(key), header=False, index=False)
+    return True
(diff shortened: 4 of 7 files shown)


semeio/reporter/reporter.py
```diff
@@ -0,0 +1,29 @@
...
+class Reporter(object):
+    _serializers = [serialize_numpy, serialize_json]
+
+    def __init__(self):
+        pass
+
+    def report(self, key, val):
+        if not isinstance(key, str):
+            raise TypeError("Report key must be of type str")
+        if key == "" or "/" in key:
```
jondequinor, line 14: It's not certain that the user will know what happened here, if some workflow author has a buggy workflow reporting with either of these keys. So maybe include in the exception what key was faulty.

Also, I really think we should constrain it to alphanumeric characters and some select others (`-_.`). Non-printable characters and a bunch of others are not going to bode well cometh the cloud.

semeio/reporter/serialize_json.py
```diff
@@ -0,0 +1,10 @@
+import json
+
+
+def serialize_json(key, val):
+    try:
+        with open("{}.json".format(key), "w") as f:
+            json.dump(val, f)
+    except:
```
jondequinor, line 8: This is bad for a number of reasons, but mainly to separate os errors (disk full, lacking permissions) and application errors (non-serializable things like dates etc).

dotfloat:
I agree with everything. My thoughts with splitting the API is that external workflows will only have 1 reporter and so having it be a global object is fine, but for internal we may want to have multiple. Thus we need to have a way to pass a reporter object. Thus the API normally is `report` while for internals we would provide a `self.report` or pass along the `self.reporter` object.

If our internal workflows run in serial then we can update the global instance. If we know that each internal workflow runs in its own thread then we can use thread-local global variable. However, I don't think it's wise to make external workflows have to explicitly create a reporter object, since they will never use more than one.

I really like the allure of external workflows needing only

```python
from semeio.reporter import report

report("foo", "bar")
```
I don't know if it's possible to replicate this in internal workflows sanely.

markusdregi:
The notion of an internal and external workflow does not make sense in the future. No workflows should have access to `enkf_main` or any other internal objects, all of them should function on streams of data. Most users would then like to get some utility for dealing with these streams and hence I think a wrapper for external scripts will add a lot of value. And in that case, we can make a reporter instance also available to them.

semeio/reporter/reporter.py
```diff
@@ -0,0 +1,29 @@
...
+from .serialize_json import serialize_json
+
+
+class Reporter(object):
+    _serializers = [serialize_numpy, serialize_json]
+
+    def __init__(self):
+        pass
+
+    def report(self, key, val):
```
markusdregi, line 11: Are we we don't want the user to be explicit about the encoding they want? A single string is valid json for instance. As mentioned in #86 I think a `report(self, val:str)` which takes a report statement and aggregates it and a `publish(self, key:str, val:jsonifiable-object)` that stores data as json. Later, we can extend by `publish_matrix(self, key:str, matrix)` and so forth.

dotfloat: I see little value in making more public-facing functions when we can easily detect and handle the type ourselves in an unambiguous way. You are correct in that a `str` is JSON-serializable and produces a JSON, but I don't see how that is a problem.

#86 uses reporting of string data in lieu of logging, which I thought we agreed was better left to Python's `logging` module.

dotfloat:
When I say "internal", I'm not talking about the availability of an EnKFMain object, but rather that the internal workflow code is run inside the same process as ERT. If today we have two ways to call `report`, one for SCOUT and one for everyone else, I think it's fair. In the future I want there to only be one way to `report` in Python, and that is to :

```python
from semeio.reporter import report

report("foo", "bar")
```

But I don't see how this is feasable today. Making everyone else do:
```python
from semeio.reporter import Reporter

def main():
  reporter = Reporter()  # There will only be 1 instance of this class ever
  start_workflow(reporter)
```
just so that we avoid distinguishing between internal and external is bad imo.

markusdregi:
> but rather that the internal workflow code is run inside the same process as ERT

But in the future even that will not be true. You should at least not depend on it. The users will benefit from some assistance to fetch the right data stream also, hence I think utility will exist and then we might as well do some utility on the reporter as well ;)

semeio/reporter/serialize_json.py
dotfloat: Note that the JSON module uses very generic errors. `TypeError` is thrown when the type is not JSON-serialisable, `OverflowError` is thrown when there's a cycle in the input.

One possibility for finer exception handling is to overload with our own `JSONEncoder.default` function that throws something other than `TypeError`, however, the docs state that:

> If specified, default should be a function that gets called for objects that can’t otherwise be serialized. It should return a JSON encodable version of the object or raise a TypeError. If not specified, TypeError is raised.

It seems that the doc is pretty strict about it this being a `TypeError`, and I'm not sure if we should break this contract.

markusdregi:
What is the status on this one? :)

dotfloat:
POC, I'd say. I can refactor it fairly quickly if we like the API as is (changing to `report_json`, `report_matrix`)

markusdregi:
If you don't mind I can probably highjack it and finish it off ;) According to Zenhub you have quite a bit on your plate...

dotfloat:
Perfectly fine with me

markusdregi:
Replaced by #99 for now...

Issue #91, opened 2020-04-28
[GENDATA_RFT] Also export data to CSV


berland:
There is a need (webviz) to be able to export the data from GENDATA_RFT not only to a bunch of individual files, but also to a CSV file.

PR #92, opened 2020-04-29
GENDATA_RFT exporting data to CSV

berland:
First step towards solving #91 

Is this something that eventually would get merged?


ertomatic:

markusdregi:
> Is this something that eventually would get merged?

I have no problems with that at least 🤷 I would however suggest that you split the flake-suggestion out in a separate PR.

oyvindeide:
> I have no problems with that at least 🤷 I would however suggest that you split the flake-suggestion out in a separate PR.

Don`t think we`ll accept changes to setup.cfg as that will be deleted because it\`s use is deprecated, feel free to add it to `tox.ini` instead.

semeio/jobs/rft/trajectory.py
```diff
@@ -68,12 +88,45 @@ def __getitem__(self, i):
...
+            )
+            dframe = pd.concat(
+                [dframe.drop("grid_ijk", axis="columns"), ijk_df], axis="columns", sort=False
+            )
+        if "measured_depth" in dframe:
+            dframe = dframe.sort_values("measured_depth")
+        elif "true_vertical_depth" in dframe:
+            dframe = dframe.sort_values("true_vertical_depth")
+        return dframe
+
```
oyvindeide, line 114: Your PR looks pretty good, but there is quite a bit of logic added here with no tests, could you add that please?

markusdregi:
> I have no problems with that at least 🤷

This comment was directed towards outputting the data as csv. I think we should strive to make the data in the csv output a super set of the current output such that one does not end up in a situation where one wants to read both.

semeio/jobs/rft/trajectory.py
berland: Tests added

oyvindeide:
Hi, how is your progress on this, is it ready for review?

berland:
I think it can be reviewed yes.  I suspect there should be some refactoring to get the aggegration of all data in a function, instead of relying on an exported CSV file, but that should maybe be done together with a similar refactor of aggregation of the txt data export (so perhaps left for a future issue).

Not  sure about the default filename.

oyvindeide:
Sounds good, I´ll have a look at it later (and feel free to change the name from WIP then ;)

semeio/jobs/rft/gendata_rft.py
```diff
@@ -62,8 +63,15 @@ def _populate_trajectory_points(
...
 
-def run(well_times, trajectories, ecl_grid, ecl_rft, zonemap=None):
-
+def run(
+    well_times,
+    trajectories,
+    ecl_grid,
+    ecl_rft,
+    zonemap=None,
+    csvfile="_gendata_rft.csv",
```
oyvindeide, line 72: Shouldn´t this be the other way around? By that I mean, you are now setting it ON as default, shouldn´t it be opt-in instead of opt-out?, this should be `None` as default?

semeio/jobs/scripts/gendata_rft.py
```diff
@@ -82,6 +82,18 @@ def _build_parser():
...
 mapping will be validated and if it is invalid, the pressure is set to -1 and the
 inactive value to 0
+""",
+    )
+    parser.add_argument(
+        "-c",
+        "--csvfile",
+        required=False,
+        type=str,
+        default="gendata_rft.csv",
```
oyvindeide, line 92: Same as above, should this be defaulted to ON?

semeio/jobs/rft/gendata_rft.py
```diff
@@ -80,6 +88,12 @@ def run(well_times, trajectories, ecl_grid, ecl_rft, zonemap=None):
 
             _write_points(trajectory_points, fname)
             _write_inactive_info(trajectory_points, fname, zonemap)
+            dframes.append(
+                trajectory_points.to_dataframe().assign(well=well, time=time)
+            )
+
+    if csvfile is not None:
```
oyvindeide, line 95: Currently this doesn´t make that much sense as it is quite hard for a user to opt-out of this, as both the function and `argparse` default to something different than None.

semeio/jobs/rft/trajectory.py
```diff
@@ -88,6 +93,41 @@ def __getitem__(self, i):
...
+        Returns:
+            pd.DataFrame
+        """
+        dframe = pd.DataFrame(data=[vars(point) for point in self.trajectory_points])
+        dframe["is_active"] = [point.is_active() for point in self.trajectory_points]
+        dframe["inactive_info"] = [
+            point.inactive_info() for point in self.trajectory_points
+        ]
+        if "grid_ijk" in dframe:
```
oyvindeide, line 110: Might be good to split this onto a couple of helper functions? Would make it easier to read.

tests/jobs/rft/test_trajectory.py
```diff
@@ -93,7 +97,7 @@ def initdir(tmpdir):
 @pytest.mark.parametrize(
     "fname", ["valid_trajectories.txt", "valid_trajectories_with_comments.txt"]
 )
-def test_load(initdir, fname):
+def test_load_and_to_dataframe(initdir, fname):
```
oyvindeide, line 100: This test currently asserts alot of different things, is it possible to split it more into unit tests? That makes it much easier to read, and also to debug if something should start failing. Related to my comment above about helper functions.

tests/jobs/rft/test_trajectory.py
```diff
@@ -93,7 +97,7 @@ def initdir(tmpdir):
...
     expected_utmxs = [0, 4]
     trajectory = Trajectory.load_from_file(fname)
```
oyvindeide, line 102: This could be mocked for most of the tests so you don´t have to load from file each time.

tests/jobs/rft/test_trajectory.py
```diff
@@ -102,6 +106,29 @@ def test_load(initdir, fname):
...
+    assert len(dframe) == 2
+    # Check that we have the input columns:
+    assert {"utm_x", "utm_y", "measured_depth", "true_vertical_depth", "zone"}.issubset(
+        set(dframe.columns)
+    )
+    # grid_ijk is a temp column, not to be present in output
+    assert "grid_ijk" not in dframe
+
+    # pressure should not be there when there is no data for it:
+    assert "pressure" not in dframe
```
oyvindeide, line 120: This could probably be it´s own test with one test for `pressure` in dframe, and one without. That would help me in knowing when to expect `pressure`, and when not to expect it.

semeio/jobs/rft/gendata_rft.py
berland: I think I want this ON by default, enabling a transition to a situation where the dumped CSV file is the authoritative source for these simulation results (and not the scattered text files).

lars-petter-hauge: All defaults should be handled in the `argparse` section. In that case the user can be informed of defaults by running `app_name --help`. If we add the argparse to documentation it will be shown there as well. I suppose if we change the default in `argparse` it will always override the one added here as well, making it irrelevant to the code.

semeio/jobs/scripts/gendata_rft.py
lars-petter-hauge: Given the statement in the above comment from author that it should be on by default - should it both output all the separate files and an aggregated version by default? I would also find it to be more sensible to not do it by default. And in the way it is setup now it is pretty much impossible to turn it off, as the `type=str` will fail if one provides a `None` argument..

semeio/jobs/rft/gendata_rft.py
berland: Yes, I will remove it from  `jobs/rft/gendata_rft.py` and inject it as a default parameter in `jobs/config_jobs/GENDATA_RFT`

lars-petter-hauge: Ah, yes, I hate that part.. Having to add it to the `JOB_DESCRIPTION` file, where none of that information is propagated to the user.. I really disliked it when we did it with `ZONEMAP` as well. The `JOB_DESCRIPTION` file doesn't handle optional parameters very well, does it.

Wouldn't need it in the `argparse` either then.

lars-petter-hauge: Here is how the optional parameter is handled for `ZoneMap` https://github.com/equinor/semeio/blob/138d1b666c4efbe033a9dc63c739736247d3560e/semeio/jobs/rft/zonemap.py#L21
I believe that is how it was handled before we did the refactoring as well.

berland: Mitigate it by syncing up JOB_DESCRIPTION and argparse behaviour is perhaps the best we can do for now. Noone (i.e.  ert 2.x users)  is using the argparse interface directly anyway except developers.

semeio/jobs/scripts/gendata_rft.py
berland: Keeping this as is, to be in sync with JOB_DESCRIPTION

semeio/jobs/rft/gendata_rft.py
berland: Changing the `run()` default to `None`, letting it be overridden by both argparse default and `JOB_DESCRIPTION`.

lars-petter-hauge: > Noone (i.e. ert 2.x users) is using the argparse interface directly anyway except developers.

But we will likely build the documentation on top of the `argparse`, and thus it will be shown indirectly to the user (this has been done for several of the jobs in everest, and will likely also be done for ert jobs). So it is good that the `argparse` instance is well-made.

oyvindeide: > Mitigate it by syncing up JOB_DESCRIPTION and argparse behaviour is perhaps the best we can do for now. Noone (i.e. ert 2.x users) is using the argparse interface directly anyway except developers.

Then it would be good to add a test to keep the two in sync.

berland: How? There is not test infrastructure in the repository to test ert on ert config files (?)

oyvindeide: > How? There is not test infrastructure in the repository to test ert on ert config files (?)

Not really, we might consider adding that, I was thinking just a quick and dirty comparison by pulling it from argparse and comparing with the `JOB_DESCRIPTION`.

berland: Done.

semeio/jobs/rft/trajectory.py
berland: `grid_ijk` refactored into its own tuple-splitting-function

tests/jobs/rft/test_trajectory.py
berland: The test with `pressure` is in `test_gendata_rft.py`. In `test_trajectory.py` there is only clean trajectories. Clearified in docstrings and comments.

tests/jobs/rft/test_trajectory.py
berland: Split into a separate test for the dataframe structure only.

oyvindeide:
If you rebase on master your tests should turn green again.

semeio/jobs/rft/gendata_rft.py
```diff
@@ -43,19 +63,32 @@ def _populate_trajectory_points(
...
             well, time, trajectories[well], ecl_grid, ecl_rft, zonemap
         )
+
         if trajectory_points:
             fname = "RFT_{}_{}".format(well, report_step)
 
             _write_points(trajectory_points, fname)
             _write_inactive_info(trajectory_points, fname, zonemap)
+            dframes.append(
+                trajectory_points.to_dataframe().assign(well=well, time=time)
```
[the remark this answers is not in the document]
berland: slightly rewritten.

berland:
> Sorry for the late review, feel free to poke me earlier should it happen again. Looking good now, two minor comments, and one a bit larger one, is it possible to address this comment:
> 
> > This comment was directed towards outputting the data as csv. I think we should strive to make the data in the csv output a super set of the current output such that one does not end up in a situation where one wants to read both.
> 
> in a test? Know there a a lot of files to check, and do not want custom parsing of them, so might not really be possible.

I suggest then I rewrite the `_write_*` functions to obtain their data from the DataFrame and not the TrajectoryPoints instances.

tests/jobs/rft/test_gendata_rft.py
```diff
@@ -68,6 +70,49 @@ def input_data(tmpdir):
...
+    assert "zone" in dframe
+    assert "i" in dframe
+    assert "j" in dframe
+    assert "k" in dframe
+    assert "pressure" in dframe
+    assert "valid_zone" in dframe
+    assert "is_active" in dframe
+    assert "inactive_info" in dframe
+    assert "well" in dframe
+    assert "time" in dframe
```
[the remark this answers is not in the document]
berland: done.

oyvindeide:
> I suggest then I rewrite the `_write_*` functions to obtain their data from the DataFrame and not the TrajectoryPoints instances.

I like that idea, there should be some tests asserting the old output, so should hopefully be ok to refactor. Er you ok with doing it in this PR?

berland:
> I like that idea, there should be some tests asserting the old output, so should hopefully be ok to refactor. Er you ok with doing it in this PR?

Yes. It did actually uncover a bug (some missing info) in the dataframe code.

semeio/jobs/rft/gendata_rft.py
```diff
@@ -1,27 +1,86 @@
 import logging
+import pandas as pd
 
 logger = logging.getLogger(__name__)
 
 
-def _write_points(points, fname):
-    with open(fname, "w+") as fh:
-        fh.write("\n".join([str(p.get_pressure()) for p in points]))
+def _write_active_pressure_points(trajectory_df, fname):
```
oyvindeide, line 7: Can you split this back into three functions? Makes it slightly easier to read.

semeio/jobs/rft/gendata_rft.py
```diff
@@ -1,27 +1,86 @@
...
+            fh.write("\n".join(["-1" * len(trajectory_df)]))
+
     logger.info("Forward model script gendata_rft.py: Wrote file {}".format(fname))
 
+    # Write a file with "1" pr row if a point is active, "0" if not
     with open(fname + "_active", "w+") as fh:
-        fh.write("\n".join([str(int(p.is_active())) for p in points]))
-
-
-def _write_inactive_info(points, fname, zonemap=None):
```
oyvindeide, line 15: Is `zonemap` no longer needed here? Is it just that it is handled elsewhere?

oyvindeide:
Jenkins: (re)test this please

semeio/jobs/rft/gendata_rft.py
berland: It was two `def`s, which should have been one or three. I can change to three.

berland: Second  thoughts: This function is only for making the GEN_DATA information, and makes sense being grouped, the files should not be written individually.

berland:
The GEN_DATA part of it would be nice  to deprecate, but should probably not be done within Ert 2.x. Refactoring  the GEN_DATA output to be based on the dataframe is also a little bit risky. If there are bugs, it will break history matching workflows.

semeio/jobs/rft/gendata_rft.py
berland: It is now handled on dataframe construction.

semeio/jobs/rft/gendata_rft.py
```diff
@@ -1,27 +1,98 @@
...
-            "\n".join([p.inactive_info(zonemap) for p in points if not p.is_active()])
+            "\n".join(
+                trajectory_df.sort_values("order")["is_active"]
+                .astype(int)
+                .astype(str)
+                .values
+            )
         )
 
+    # Write a file with textual information for each pressure point that has been
```
oyvindeide, line 58: Do you mean contextual?

oyvindeide:
Checked the `jenkins` build, and it is not failing because of this PR, but because of `configsuite` out of sync across different repos (we are testing against `komodo`).

semeio/jobs/rft/gendata_rft.py
berland: Not really. Rephrased.

berland:
(everything is to be squashed when ready)

berland:
Squashed.

oyvindeide:
Thanks for the PR and for bearing out with me :)

PR #95, opened 2020-05-04
Black all code, and resolve all flake8 issues

berland:


ertomatic:

markusdregi:
I find the black-commit a bit surprising as the test suite is already running black. Did black suggest these or is it flake that is complaining?

berland:
These trailing commas seem to have been a feature of an older black. 19.3b0 (/usr/bin/black on my local ubuntu) will inject these, while 19.10b0 from my virtenv will remove them.

oyvindeide:
> These trailing commas seem to have been a feature of an older black. 19.3b0 (/usr/bin/black on my local ubuntu) will inject these, while 19.10b0 from my virtenv will remove them.

Tests are running `black==19.10b0`, which was passing on commits prior to this change.

berland:
Ah. I think I got it, it is black 19.3b0 that removes the trailing commas, while 19.10b0 does not care whether the comma is there or not.

markusdregi:
Then I suggest that you update your `black` @berland ;)

tox.ini
```diff
@@ -20,6 +21,13 @@ deps =
 commands =
     black --check semeio tests setup.py
 
+[testenv:flake8]
+description = Check PEP8 using flake8
+deps =
+    flake8
+commands =
+    flake8 semeio tests setup.py
```
oyvindeide, line 29: I´m fine with having `flake8` in the test suite, but _think_ we should do `python -m flake8 ...`, that goes for the other tests as well, but that is for another issue.

oyvindeide:
Thanks, looks good now, just one more comment (which I should have made yesterday but forgot), can you also add it to `.travis.yml` like this:
```
    - python: 3.7
        env: TOXENV=flake8
```

PR #98, opened 2020-05-12
Ignore order of columns in csv-export2 testing

oysteoh:


berland:
I think this is the line that fails: https://github.com/equinor/semeio/blob/df8233e3d120d5ad8f178508dec5abac46305413/tests/jobs/csv_export2/test_integration.py#L113

PR #99, opened 2020-05-12
Reporter

### LINKED ISSUE #82: Use higher level interface for data/messages to/from jobs

markusdregi:
Closes #82

Remaining work:
- [x] Fix Jenkins failure
- [x] Write integration test having ERT execute a SemeioScript


ManInFez:
What happens if two reporters are reporting to the same file?

tests/communication/test_semeio_script.py
```diff
@@ -0,0 +1,204 @@
...
+    ert = Mock(
+        resConfig=Mock(
+            return_value=Mock(model_config=Mock(getEnspath=Mock(return_value=enspath)))
+        )
+    )
+
+    msg = "My logging msg"
+
+    class MySuperScript(SemeioScript):
+        def run(self, *args):
```
ManInFez, line 56: shouldnt this also execute the run method of SemeioScript? `super(MySuperScript, self).run() or something?

semeio/communication/semeio_script.py
```diff
@@ -0,0 +1,65 @@
...
+    def _format_record(self, log_record):
+        log_fmt = "{log_level} [{log_time}]: {log_message}"
+        return log_fmt.format(
+            log_level=log_record.levelname,
+            log_time=datetime.datetime.fromtimestamp(log_record.created),
+            log_message=log_record.message,
+        )
+
+
+class SemeioScript(ErtScript):  # pylint: disable=too-few-public-methods
```
ManInFez, line 43: I thinks this class deserves a bit of documentation, as it is quite convoluted with the internal function, report handler, and log handler context

markusdregi:
> What happens if two reporters are reporting to the same file?

This first iteration is by no means thread safe.

tests/communication/test_semeio_script.py
markusdregi: Well, should and could and all of that 🤷 I just wanted to put myself in-between an implementation of an `ErtScript` and `ERT` itself in such a manner that any workflow could start inheriting `SemeioScript` instead of `ErtScript` and would continue to function. Ert will call the `run`-method, while `ErtScript` does in fact not implement `run`. So, there is an expectancy that the child class will implement it. And no child class currently expects there to be a `super.run`-method. Hence, I injected myself in-between by wrapping the `run`-method of the child in the `__init__`.

semeio/communication/semeio_script.py
markusdregi: I agree 👍

semeio/communication/semeio_script.py
```diff
@@ -0,0 +1,65 @@
...
+        base_dir = ert.resConfig().model_config.getEnspath()
+        base_dir = os.path.realpath(base_dir)
+        self._output_dir = os.path.join(base_dir, "reports", type(self).__name__,)
+        self._reporter = FileReporter(self._output_dir)
+
+        # pylint: disable=access-member-before-definition
+        self._real_run = self.run
+
+        def run_with_handler(self, *args):
+            log = logging.getLogger("")
```
dotfloat, line 56: Seems like you're modifying the global logger. I would be very surprised as a user if I was implementing a subclass of `SemeioScript` and had such side-effects.

markusdregi:
Jenkins: Retest this please

ManInFez:
Test this please!

semeio/communication/semeio_script.py
markusdregi: Made an issue here: https://github.com/equinor/semeio/issues/104

markusdregi:
Figured out the Jenkins error in the end. It was the default log level that was different. Bumped all tests to use `error` to ensure that they get through...

semeio/communication/semeio_script.py
markusdregi: Wrote some documentation and split __init__ into some smaller functions

markusdregi:
I've now also made an integration test. However, there is a caveat; it requires ERT to be installed. Hence, as of now it will not run on Travis, but only on Jenkins. We can leave it like this, which is my suggestion as of now. Or we can install ERT on Travis as well or rewrite the integration test to be towards libres. To me the libres `WorkflowRunner` should not be of concern for `semeio`. We don't really want to know how ERT executes workflows. And for installing ERT on Travis, we could, but I suggest we wait until libres and hence ERT is pip-installable. If you agree, I'll add it to the list in https://github.com/equinor/semeio/issues/101

semeio/communication/reporter.py
```diff
@@ -0,0 +1,37 @@
...
+    def _prepare_output_file(self, namespace):
+        if not os.path.exists(self._output_dir):
+            os.makedirs(self._output_dir)
+
+        if not os.path.isdir(self._output_dir):
+            msg = "Expected output_dir to be a directory ({})"
+            raise ValueError(msg.format(self._output_dir))
+
+        if os.path.sep in namespace:
+            err_msg = "Namespace contains path separators ({}) "
```
ManInFez, line 34: trailing whitespace

semeio/communication/reporter.py
```diff
@@ -0,0 +1,37 @@
...
+        output_file = self._prepare_output_file(namespace) + ".json"
+
+        with open(output_file, "w") as f:
+            json.dump(data, f)
+
+    def publish_msg(self, namespace, msg):
+        fmt = "{}\n"
+        output_file = self._prepare_output_file(namespace)
+        with open(output_file, "a") as f:
+            f.write(fmt.format(msg))
```
ManInFez, line 23: should message be stripped for trailing new line, in order to prevent empty lines?

semeio/communication/reporter.py
```diff
@@ -0,0 +1,37 @@
...
+        self._output_dir = output_dir
+
+        if not os.path.isabs(output_dir):
+            err_fmt = "Expected output_dir to be an absolute path, received '{}'"
+            raise ValueError(err_fmt.format(output_dir))
+
+    def publish(self, namespace, data):
+        output_file = self._prepare_output_file(namespace) + ".json"
+
+        with open(output_file, "w") as f:
```
ManInFez, line 16: For me, it is unclear as to why we are overwriting the data in a given namespace?

semeio/communication/reporter.py
markusdregi: Thank you 👍

semeio/communication/reporter.py
markusdregi: I'm into just giving people what they're asking for. This is unstructured output anyway. If you want trailing spaces, you can have them? 🤷

semeio/communication/reporter.py
markusdregi: Well, mostly due to the lack of a better alternative. Any other alternative is prone to give highly unexpected results. Note that as we are basing the location on `enspath`, users typically use `CASE_NAME` as part of this and hence would have separate locations for different cases. Also, note that the persistence of data is highly experimental and will not reside of disk for long and hence any attempt to pick up the data in the other end is at your own risk 😬

markusdregi:
I’m not sure which stdout you mean; the one of the workflow or of the client? Right now they are the same, but that is a coincidence more than a feature.

But, as a consequence of this, if the workflow wants to propagate a message to the user it should use the logger. It will persist to disk via the reporter, but also to standard out of the workflow (and also the client). Hence, logging will currently give you output in the terminal.

When that is said I must stress that this PR in no way focused on the machinery storing the messages, nor a client for consuming them. It is a crude implementation dumping to disk; while offering an API that can serve as a basis on top  of a proper backend later on. My hope is that this data soon will reside within new storage ☺️

PR #100, opened 2020-05-13
Skip install for formatting

oyvindeide:


oyvindeide:
> 🐎

Thanks! You are super fast! Do you still agree with my `fixup`?

PR #102, opened 2020-05-13
Improve error message to user from DESIGN_KW

berland:
Ref. error reported on slack ert-users.

This will only improve the error message.

### DIFF
--- a/semeio/jobs/design_kw/design_kw.py
+++ b/semeio/jobs/design_kw/design_kw.py
@@ -79,15 +79,44 @@ def is_comment(line):
 
 
 def extract_key_value(parameters):
-    # Extract all key, value pairs from parameters file
-    # If key already exists, raise Exception
+    """Parses a list of strings, looking for key-value pairs pr. line
+    separated by whitespace, into a dictionary.
 
+    Spaces in either key or in value are not supported.
+
+    Args:
+        parameters (list of str)
+
+    Returns:
+        dict, with the keys and values parsed.
+
+    Raises:
+        ValueError, with error messages and all unparseable lines.
+    """
     res = {}
+    errors = []
     for line in parameters:
-        key, value = line.split()
+        line_parts = line.split()
+        if not line_parts:
+            continue
+        if len(line_parts) == 1:
+            errors += ["No value found in line {}".format(line)]
+            continue
+        if len(line_parts) > 2:
+            errors += ["Too many values found in line {}".format(line)]
+            continue
+        key, value = line_parts
         if key in res:
-            raise SystemExit(
-                "{} is defined multiple times in parameters file".format(key)
-            )
+            errors += ["{} is defined multiple times".format(key)]
+            continue
         res[key] = value
+    if errors:
+        # Gather errors, and add warnings that should only be
+        # given once (and not for every occurence)
+        error_str = "\n".join(errors)
+        if '"' in error_str:
+            error_str += "\nQuotes are not supported."
+        if "Too many values" in error_str:
+            error_str += "\nSpaces in values are not supported."
+        raise ValueError(error_str)
     return res
--- a/tests/jobs/design_kw/test_design_kw.py
+++ b/tests/jobs/design_kw/test_design_kw.py
@@ -29,9 +29,31 @@ def test_extract_key_value_ok():
     }
 
 
+def test_extract_key_value_whitespace_error():
+    """Test that user friendly error message is displayed
+    when an error occurs"""
+    data = ['"FOO BAR COM " 2']
+    with pytest.raises(ValueError, match="FOO BAR COM"):
+        design_kw.extract_key_value(data)
+    with pytest.raises(ValueError, match="Quotes"):
+        design_kw.extract_key_value(data)
+
+    data = ["SENTENCE this is a string"]
+    with pytest.raises(ValueError, match="SENTENCE"):
+        design_kw.extract_key_value(data)
+    with pytest.raises(ValueError, match="Spaces in values are not supported"):
+        design_kw.extract_key_value(data)
+
+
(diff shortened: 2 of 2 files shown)


berland:
Previous error is:

```
  <File "/prog/res/komodo/2020.05.00-py36/root/lib/python3.6/site-packages/semeio/jobs/scripts/design_kw.py", line 53, in <module>
    main(sys.argv[1:])
  File "/prog/res/komodo/2020.05.00-py36/root/lib/python3.6/site-packages/semeio/jobs/scripts/design_kw.py", line 48, in main
    log_level=parsed_args.log_level,
  File "/prog/res/komodo/2020.05.00-py36/root/lib/python3.6/site-packages/semeio/jobs/design_kw/design_kw.py", line 30, in run
    key_vals = extract_key_value(parameters)
  File "/prog/res/komodo/2020.05.00-py36/root/lib/python3.6/site-packages/semeio/jobs/design_kw/design_kw.py", line 91, in extract_key_value
    key, value = line.split()
ValueError: too many values to unpack (expected 2)>
```

ertomatic:

oyvindeide:
Thanks for the patch, seems like a good addition. One question though, it seems like the py2 behavior is to strip quotes, are any other characters stripped in the py2 version that should be handled spesifically?

Jenkins: test this please

berland:
Not sure why this was not a problem with py2, the semeio code would at least give the same error in py2. Perhaps not semeio code was in use, or the design matrix generator behaves differently.

fmu-tools is also being fixed to strip spaces from stray spaces in excel input, to avoid this happening.

oyvindeide:
> Not sure why this was not a problem with py2, the semeio code would at least give the same error in py2

These jobs are only used in py3, py2 jobs are coming from `ert-statoil`.

tests/jobs/design_kw/test_design_kw.py
```diff
@@ -29,6 +29,17 @@ def test_extract_key_value_ok():
     }
 
 
+def test_extract_key_value_whitespace_error():
+    """Test that user friendly error message is displayed
+    when an error occurs"""
+    data = ['"FOO BAR COM " 2']
+    try:
```
oyvindeide, line 36: I prefer using: 
`with pytest.raises(ValueError) as error_msg:` in tests, though they do the same under the hood, it makes it easier to read (I think).

berland: Solved when I found the `match=` argument to `pytest.raises()`. If not, it is difficult to do asserts on the message.

berland:
I agree that parsing the entire `parameters.txt` before giving errors is even better.

semeio/jobs/design_kw/design_kw.py
```diff
@@ -83,11 +83,22 @@ def extract_key_value(parameters):
...
+    unparsable_lines = []
     for line in parameters:
-        key, value = line.split()
+        try:
+            key, value = line.split()
+        except ValueError:
+            unparsable_lines += [line]
+            continue
         if key in res:
             raise SystemExit(
```
oyvindeide, line 94: I think this should also be included in your list of unparsable_items, and instead of just adding the line to the list, you could add the whole error message, that makes it easier to expand it later should we want to add more validation. Then you can just loop over that list in the end if it exists before raising.

berland: Done

semeio/jobs/design_kw/design_kw.py
```diff
@@ -79,15 +79,50 @@ def is_comment(line):
...
+        except ValueError as e_msg:
+            if "not enough values to unpack" in str(
+                e_msg
+            ) or "need more than 1 value to unpack" in str(e_msg):
+                # (py2 and py3 differ in emitted error message)
+                errors += ["No value found in line {}".format(line)]
+            elif "too many values to unpack" in str(e_msg):
+                errors += ["Spaces in values are not supported"]
+            else:
+                errors += [str(e_msg)]
```
oyvindeide, line 111: I am not sure parsing the error message is the best approach here, it would for example be possible to use `len(line.split())` in the error message and say for example `Expected 2, found: {} in line: {}).format(len(line.split()), line)` or something of that nature.

semeio/jobs/design_kw/design_kw.py
```diff
@@ -79,15 +79,50 @@ def is_comment(line):
...
-            )
+            errors += ["{} is defined multiple times".format(key)]
+            continue
         res[key] = value
+    if unparsable_lines or errors:
+        error_str = ""
+        if unparsable_lines:
+            error_str = "Could not parse these lines:\n"
+            error_str += "\n".join(unparsable_lines)
+        if '"' in error_str:
```
oyvindeide, line 123: Do you think this kind of error is the most likely to pop up as you are special casing for it? If you think it is likely that it will happen again I´m fine with special casing it. But if the error from above is enough to clue to user in, it might not be needed?

berland: I  anticipate that users might try to add quotes if they need strings with spaces as values in  `parameters.txt`, so as long as it is not supported, we can give an (overly?) helpful message.

semeio/jobs/design_kw/design_kw.py
berland: Rewritten to catch the error before  the `ValueError` occurs. I have kept different errors for `len() == 1` and `len() > 2`, I  think it  benefits  users.

berland: This change also introduced  ignoring empty lines. The code in the master branch will error on empty lines. Not sure if that is an error, or if it is something we should ignore. Comments are already ignored, so it can make sense to  ignore empty lines also.

PR #103, opened 2020-05-13
Add pre commit

oyvindeide:


lars-petter-hauge:
Suppose not all changes in `tox.ini` should be in the `fix formatting` commit (i.e. the `pre-commit` command is added there)

lars-petter-hauge:
I've tested this locally, and I will keep using it for some time to see if I like it. Especially important how long it takes to run (good that it only runs on the diff though!).

With this merge we in fact force flake8, pylint and even no-whitespace on all PR's. We are, if not anything else, strict :)

In favor of adding this to see how it is. If we like, perfect, otherwise it's not much of a hassel to remove.

lars-petter-hauge:
Rebase and we’re good to go?

markusdregi:
Do we really want to add pre-commit hooks to the repo? Is not what happens between me and my branch a personal matter? 😂

oyvindeide:
> Do we really want to add pre-commit hooks to the repo? Is not what happens between me and my branch a personal matter? 😂

The actual hooks aren´t added, you have to install `pre-commit` locally to set them up, so it is opt-in. It´s quite standard on many popular repos. The good thing is that it groups all the formatting tests, and the day I implemented it, I had a look at PRs in `semeio` and noticed that many had all regular tests passing, but was failing on of the formatting tests. I´ve done the same many times and this will alleviate that annoyance if you chose to use it.

I´m happy to have a discussion though if you feel strongly against it.

oyvindeide:
> Rebase and we’re good to go? 

Yes and no, just want to check with @markusdregi next week regarding the `fixup` before that.

markusdregi:
First of all, let me just make clear that you are the maintainers of the repository and hence have the deciding vote on this 👍 At least until the team riots 🏴‍☠️ 

> it is opt-in

Then I personally have nothing against it. I just believe that we should be very careful dictating tooling and process before a PR is flagged as `ready for review`...

oyvindeide:
> Then I personally have nothing against it. I just believe that we should be very careful dictating tooling and process before a PR is flagged as ready for review...

Agreed, and I know @lars-petter-hauge feels the same way as that was the first thing he asked me.

While I have you here @markusdregi, what do you think about 6085cff83d3da26c79e7005aabda950b61933bc8, should we ignore the errors, as there really isn´t much we can do about it?

markusdregi:
It is a bit disappointing to be honest if every time you make a `SemeioScript` you'll have to disable a `pylint` warning 🤔

markusdregi:
@oyvindeide: Which version of `pylint` are you using? I thought that I'd run `pylint` manually on both the tests and semeio folder, and it seems plausible. Because I'm getting no warnings on my machine with `2.5.2`...

oyvindeide:
> @oyvindeide: Which version of pylint are you using? I thought that I'd run pylint manually on both the tests and semeio folder, and it seems plausible. Because I'm getting no warnings on my machine with 2.5.2...
 
Can reproduce with `2.5.2` on master, but you have to set this: 
`ignore-patterns=tests`
to
`ignore-patterns=`
in `pylintrc`

oyvindeide:
>  It is a bit disappointing to be honest if every time you make a SemeioScript you'll have to disable a pylint warning 🤔

We can disable it globally though.

markusdregi:
You're right @oyvindeide! I suggest we add it to the global list... I'll try to come up with a way to avoid it 👍

oyvindeide:
> You're right @oyvindeide! I suggest we add it to the global list... I'll try to come up with a way to avoid it 👍

Sounds good, I´ll fix the warning here then and create an issue to investigate.

lars-petter-hauge:
Although we specifically stated on the command line call that `pylint` should go through the `tests` folde:

PR #113, opened 2020-05-26
Add INSERT_NOSIM and REMOVE_NOSIM jobs

sondreso:
Resolves https://github.com/equinor/semeio/pull/113


sondreso:
Inputs on how this could be tested is appreciated :slightly_smiling_face:

oyvindeide:
A really ugly integration test by executing it with `libres`? 🤷

sondreso:
I was hoping to avoid that, but it might be to much to ask :see_no_evil:

oyvindeide:
We don´t have anything for the other `JOB_DESCRIPTION` files, and these are not special in that regards. If you come up with something that works reasonably well we should implement it on the others as well. That being said we should find another way to do this asap.

sondreso:
The easiest is maybe to crate a ERT setup which includes all of the jobs, and then run that as one very big integration test?

oyvindeide:
We should perhaps have some test fixture for this, resolving the job, preferably in ert, but guess it has to be libres. It is one of the most important APIs we have.

sondreso:
Updated with integration tests

tests/jobs/nosim/test_nosim.py
```diff
@@ -0,0 +1,70 @@
...
+    ):
+        p = subprocess.Popen(
+            ["ert", "test_run", "nosim.ert", "--verbose"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            env=os.environ.copy(),
+        )
+        p.wait()
+
+    with open("nosim/real_0/iter_0/TEST.DATA") as fh:
```
oyvindeide, line 69: I can´t get this to pass locally on my machine, any idea why? Running osx, example output:
`fh.read()`:
```
RUNSPEC
NOSIM
-- and more RUNSPEC
```
Expected:
```
RUNSPEC
-- and more RUNSPEC
```

sondreso: Have you tried running the sed command manually to see if you get the expected results? There are slight differences in how sed works on mac and linux, but I'm surprised if it's this big of a difference!

sondreso: Apparently:
> . -i works with GNU sed, recent versions of NetBSD, OpenBSD sed, but not OS X's sed. 

https://unix.stackexchange.com/a/131940

sondreso: Do we care about this? It's an copy of the current NOSIM job, so neither will work on OSX.

oyvindeide: Looks like the alternative is some funky rewrite, I support just keeping it as is.

PR #124, opened 2020-06-05
Update tests and description

lars-petter-hauge:
I've shamefully copied content from https://github.com/equinor/semeio/pull/77 and added it to the description of the schema. Furthermore the schema should then either be added using the generate rst function, or using the sphinx extension from config suite.

The tests have been slightly restructured as well, one test removed (`invalid_too_short_index_list`, as it did not test what was expected, outdated), and added one more for required keys.

Currently one test is failing because there are multiple errors when adding a negative value in the index list. The second error, which shouldn't be there, is:
`"'Will go through the input and try to convert to list of int' failed on input '[-1, 2, 3]' with error 'invalid literal for int() with base 10: '''"`

Not clear how to handle the last error yet.

### DIFF
--- a/semeio/jobs/correlated_observations_scaling/job_config.py
+++ b/semeio/jobs/correlated_observations_scaling/job_config.py
@@ -44,10 +44,13 @@ def _realize_list(input_string):
     """
     real_list = []
     for elem in input_string.split(","):
+        if elem.startswith("-"):
+            raise ValueError(
+                "Elements can not be negative, neither singletons nor in range,"
+                " got: {}".format(elem)
+            )
         bounds = elem.split("-")
         if len(bounds) == 1:
-            if "-" in elem:
-                raise ValueError("Did not expect '-' in singleton")
             real_list.append(int(elem))
         elif len(bounds) == 2:
             if elem.count("-") != 1:
--- a/semeio/jobs/correlated_observations_scaling/job_config.py
+++ b/semeio/jobs/correlated_observations_scaling/job_config.py
@@ -84,6 +84,82 @@ def _min_max_value(value):
     return 0.0 < value < 1.0
 
 
+_CALCULATE_KEYS_DESCRIPTION = """
+The keys that will be used for calculating a scaling factor and update all
+data points within said keys. Include the indexes under "index" in order to
+update selected data points. Use UPDATE_KEYS if you would like the scaling
+to be applied to other keys than the ones listed under CALCULATE_KEYS.
+The configuration accepts a list of CALCULATE_KEYS, facilitating update on
+multiple groups for a single run, instead of having to run multiple times.
+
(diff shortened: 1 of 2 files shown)


lars-petter-hauge:
I can split out the
```
if elem.startswith("-"):
    raise ValueError("Do no accept negative elements")
```
in a single commit..

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -44,10 +44,10 @@ def _realize_list(input_string):
     """
     real_list = []
     for elem in input_string.split(","):
+        if elem.startswith("-"):
+            raise ValueError("Do no accept negative elements")
```
oyvindeide, line 48: Might want to reword this message a bit.

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -93,67 +169,36 @@ def build_schema():
...
-                            }
-                        },
-                    },
+                    "keys": _KEYS_SCHEMA,
                     "threshold": {
                         MK.Required: False,
                         MK.Type: types.Number,
                         MK.ElementValidators: (_min_max_value,),
+                        MK.Description: "Instructs the job to ignore principal"
+                        " components where the summative variance is smaller"
```
oyvindeide, line 180: Perhaps use a different word than summative? Not very commonly used.

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -93,67 +169,36 @@ def build_schema():
...
+                        " than this value.",
+                    },
+                    "std_cutoff": {
+                        MK.Required: False,
+                        MK.Type: types.Number,
+                        MK.Description: "Filters out any observation that has"
```
oyvindeide, line 186: Is this accurate enough? Is filtered out if the ensemble std is outside that value.

semeio/jobs/correlated_observations_scaling/job_config.py
lars-petter-hauge: Rephrased. Also note that there will be two errors thrown, and this is only from the transformation. The error from validation will still be output.

semeio/jobs/correlated_observations_scaling/job_config.py
lars-petter-hauge: changed with `cummulative`

semeio/jobs/correlated_observations_scaling/job_config.py
lars-petter-hauge: I think we should use the same descriptions for both this and the misfit preprocessor job. I'll copy those over here.

lars-petter-hauge:
If you're happy with this I'll rebase :)

PR #125, opened 2020-06-05
Updated configsuite

lars-petter-hauge:
The one that bugs me a bit is the change in user feedback when `MK.Required` is not used:
From: `'Missing key: CALCULATE_KEYS'`
To: `Section length must be greater than 0 is false on input '()'`

I could have made context validators, but issues are:
 - has to be fairly hacky to facilitate every section that will be validated
 - made unique context validator for each container that must be validated.

### DIFF
--- a/semeio/jobs/correlated_observations_scaling/job_config.py
+++ b/semeio/jobs/correlated_observations_scaling/job_config.py
@@ -87,6 +87,11 @@ def _min_max_value(value):
     return 0.0 < value < 1.0
 
 
+@configsuite.validator_msg("keys must be provided for CALCULATE_KEYS")
+def _CALCULATE_KEYS_key_not_empty_list(content):
+    return len(content) > 0
+
+
 _CALCULATE_KEYS_DESCRIPTION = """
 The keys that will be used for calculating a scaling factor and update all
 data points within said keys. Include the indexes under "index" in order to
@@ -130,21 +135,19 @@ on "FOPR", but only update the scaling on indices "50-100".
 """
 
 _KEYS_SCHEMA = {
-    MK.Required: True,
+    MK.ElementValidators: (_CALCULATE_KEYS_key_not_empty_list,),
     MK.Type: types.List,
     MK.Content: {
         MK.Item: {
             MK.Type: types.NamedDict,
             MK.Content: {
                 "key": {
-                    MK.Required: True,
                     MK.Type: types.String,
                     MK.Description: "Name of the key. An asterisk is accepted"
                     " as a suffix to expand all matching keywords (e.g."
                     " WOPR* will include WOPR:OP1)",
                 },
                 "index": {
-                    MK.Required: False,
                     MK.Type: types.List,
                     MK.LayerTransformation: _to_int_list,
                     MK.Description: "Indexes matching the data points relevant"
@@ -170,13 +173,11 @@ def build_schema():
         MK.LayerTransformation: _expand_input,
         MK.Content: {
             "CALCULATE_KEYS": {
-                MK.Required: True,
                 MK.Type: types.NamedDict,
                 MK.Description: _CALCULATE_KEYS_DESCRIPTION,
                 MK.Content: {
                     "keys": _KEYS_SCHEMA,
                     "threshold": {
-                        MK.Required: False,
                         MK.Type: types.Number,
                         MK.ElementValidators: (_min_max_value,),
                         MK.Description: "Threshold used when computing primary"
@@ -184,7 +185,6 @@ def build_schema():
                         MK.Default: 0.95,
                     },
                     "std_cutoff": {
-                        MK.Required: False,
                         MK.Type: types.Number,
                         MK.Description: "A lower bound on the ensemble standard"
                         " deviation. All data points with insufficient variation"
@@ -192,7 +192,6 @@ def build_schema():
                         MK.Default: 0.000001,
                     },
                     "alpha": {
-                        MK.Required: False,
                         MK.Type: types.Number,
                         MK.Description: "Scalar controlling the allowed distance"
                         " between ensemble mean and observation. In particular,"
@@ -204,7 +203,6 @@ def build_schema():
                 },
             },
             "UPDATE_KEYS": {
-                MK.Required: False,
                 MK.Type: types.NamedDict,
                 MK.Description: _UPDATE_KEYS_DESCRIPTION,
                 MK.Content: {"keys": _KEYS_SCHEMA},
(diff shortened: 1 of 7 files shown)


semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -84,76 +84,123 @@ def _min_max_value(value):
...
     return {
         MK.Type: types.NamedDict,
         MK.Description: "Keys and index lists from all scaled keys",
         MK.LayerTransformation: _expand_input,
         MK.Content: {
             "CALCULATE_KEYS": {
-                MK.Required: True,
                 MK.Type: types.NamedDict,
+                MK.Description: _CALCULATE_KEYS_DESCRIPTION,
+                MK.ElementValidators: (_is_not_empty,),
```
markusdregi, line 175: I don't think the `_is_not_empty` validator does you much good on a Named dict?

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -84,76 +84,123 @@ def _min_max_value(value):
...
-                                        },
-                                    },
-                                },
-                            }
-                        },
-                    },
+                    "keys": _KEYS_SCHEMA,
                     "threshold": {
-                        MK.Required: False,
+                        MK.AllowNone: True,
```
markusdregi, line 179: Does it make sense to allow the `threshold` to be `None`? Should we not instead provide a default value? I think there already is a layer that currently keeps the default values...

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -84,76 +84,123 @@ def _min_max_value(value):
...
                         MK.Type: types.Number,
                         MK.ElementValidators: (_min_max_value,),
+                        MK.Description: "Instructs the job to ignore principal"
+                        " components where the summative variance is smaller"
+                        " than this value.",
+                    },
+                    "std_cutoff": {
+                        MK.AllowNone: True,
```
markusdregi, line 187: Default instead?

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -84,76 +84,123 @@ def _min_max_value(value):
...
+                        MK.Type: types.Number,
+                        MK.Description: "Filters out any observation that has"
+                        " a standard deviation above this value.",
+                    },
+                    "alpha": {
+                        MK.AllowNone: True,
```
markusdregi, line 193: Default instead?

markusdregi:
> From: 'Missing key: CALCULATE_KEYS'
To: Section length must be greater than 0 is false on input '()'

How about if you write the validator message as `keys must be provided` and then add the key-path of the error to the print? Currently that will reside in the message: `keys must be provided is false on input '()'`, that you can combine with the key-path `CALCULATE_KEYS` as you like. With https://github.com/equinor/configsuite/issues/147 you could even customise the message format to avoid the `is false on input '()'` as it does not give you much in this scenario.

semeio/jobs/correlated_observations_scaling/job_config.py
lars-petter-hauge: I actually did provide a Default when I added the description in https://github.com/equinor/semeio/pull/124, but as `MK.Default` was not accessible for `0.5.3` I removed it again, just for the sake of that PR. I'll definitively add it instead - and then we don't need `MK.AllowNone` of course.

semeio/jobs/correlated_observations_scaling/job_config.py
lars-petter-hauge: Quite true - removed!

semeio/jobs/correlated_observations_scaling/job_config.py
lars-petter-hauge: yes

semeio/jobs/correlated_observations_scaling/job_config.py
lars-petter-hauge: yes

lars-petter-hauge:
> How about if you write the validator message as `keys must be provided `and...

As it now is only used for a single element validation, I've gone with specifying the message completely.

I don't disagree including the key path though, but I suppose that will be added on a global basis and not only for this specific event.

lars-petter-hauge:
Fairly happy with this one, but the configsuite version is not updated from `0.5.3`. Suppose`tox` is not using the content of `install_requires` from `setup.py`. Fine by me, just need to figure out where it should be

.travis.yml
```diff
@@ -47,14 +47,15 @@ install:
     - git clone --branch $LIBECL_VERSION --depth 1 https://github.com/equinor/libecl
     - bash ert/.build_install.sh libecl
     - bash ert/.build_install.sh libres
-    - export PYTHONPATH=$INSTALL_DIR/lib/python$TRAVIS_PYTHON_VERSION/dist-packages:$PYTHONPATH
-    - export PYTHONPATH=$INSTALL_DIR/lib/python$TRAVIS_PYTHON_VERSION/site-packages:$PYTHONPATH
+    - export PYTHONPATH=$PYTHONPATH:$INSTALL_DIR/lib/python$TRAVIS_PYTHON_VERSION/dist-packages
```
markusdregi, line 50: Why?

semeio/jobs/correlated_observations_scaling/job_config.py
```diff
@@ -87,6 +87,11 @@ def _min_max_value(value):
     return 0.0 < value < 1.0
 
 
+@configsuite.validator_msg("keys must be provided for CALCULATE_KEYS")
+def _is_not_empty(content):
```
markusdregi, line 91: This is not a general `_is_not_empty` anymore. Please change the name of the validator :)

markusdregi:
This is starting to look good 👍 Address my two comments and fix the tests - and then we are good to go 🤷

.travis.yml
lars-petter-hauge: The issue I have is that libres has a hard dependency for configsuite `0.5.3` (or had - I've removed it now as it wasn't a real dependency). All my last 5-6 commits are just trying to find a way to update configsuite to `0.6`, and that doesn't seem to be as easy as I thought.

I could just use the master branch of libres instead of using the one in `.LIBRES_VERSION` from `ert`, but I wanted to see if I could find a more sane way, as this isn't really an edge case.

lars-petter-hauge:
Just want to make things run at least. So now I've specified master version of libres, instead of using the one tagged in ert.. I don't really like this solution, and if anyone have a good idea on how to make sure that `ConfigSuite 0.6` is used during the test, I'll be happy to use it!

markusdregi:
I think we should make a new tag in libres and update the one used in ERT. I'll get to it.

Right now you have the problem of inconstant dependency requirements and that is a real problem that cannot be bypassed (although it is not real this time...).

lars-petter-hauge:
In the `update test_scaling_job I've only moved stuff around and included verification of what `get_index_lists` and `get_calc_keys` returns.

PR #137, opened 2020-08-10
Persist all reported data

### LINKED ISSUE #120: [COS] Reporter overwriting previous values
The [COS] job can typically be run several times back to back, and currently the reported values are overwritten each time, should be append perhaps?

markusdregi:
Resolves #120 

This is a possible solution to the reporter overwriting previously published data, by keeping all reports of the same namespace and name within a list. As there are no guarantees on the output format of the data, and the data is to be moved to storage over time, I believe this is a non-controversial change. It is for sure not the perfect solution for the problem, but I think it is adequate until we start learning about consuming this data from a readers perspective.

### DIFF
--- a/semeio/communication/reporter.py
+++ b/semeio/communication/reporter.py
@@ -16,8 +16,16 @@ class FileReporter(object):
 
     def publish(self, namespace, data):
         output_file = self._prepare_output_file(namespace) + ".json"
+
+        if os.path.exists(output_file):
+            with open(output_file) as f:
+                all_data = json.load(f)
+        else:
+            all_data = []
+
+        all_data.append(data)
         with open(output_file, "w") as f:
-            json.dump(data, f)
+            json.dump(all_data, f)
 
     def publish_msg(self, namespace, msg):
         fmt = "{}\n"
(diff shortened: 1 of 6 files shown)


oyvindeide:
This looks like a pragmatic solution to the first problem. Though I am a bit worried about the files becoming a bit hard to read as the users often have several runs in the same location and there is no context in the data. Should we add a logged warning that the file is appended?

markusdregi:
I agree that the proposed solution is not optimal. However, I'm not sure if it is nicely solvable as of now. Because we want the possibility of logging multiple times, but we do run into danger and confusion when runs are reran and reports are being merged. And currently we do not have a clear way of separating these two scenarios.

Since the consumption of the output data is experimental at best, I suggest we keep it as it for now and later implement a loader that allows us to store some meta data or that we just moves immediately to storing reports in storage. Then, the reports can be tied to the evaluation of a particular pipeline and hence the confusion should go away 🤷 

The logged warning will almost always happen, so I'm not convinced that will be of much help. We could also look into having the misfit preprocessor clear the reports at each new run. That would probably avoid most of the confusing cases...

PR #140, opened 2020-08-14
Get alpha and std_cutoff from ert

oyvindeide:


oyvindeide:
@markusdregi I have chosen a slightly different approach that the issue suggested, and put the behavior in the job itself. The main reason for that is that it is possible to run the job from three different places, ACO, spearman and the job itself. I felt it would be confusing behavior for them to act differently. The second reason is that it really should not be defaulted in the job itself because the default should be the behavior of `ert`, and by only putting it in AOC we would keep a separate default in the job itself.I have no problem changing that behavior though if you feel strongly for implementing it in AOC only. But then I would suggest we remove the possibility of calling the job from the spearman job, which maybe we should do anyway as it has been surpassed by AOC?

oysteoh (approved):

PR #142, opened 2020-08-21
Delete deprecated

oyvindeide:
This is never called from anywhere and can safely be deleted


oysteoh (approved):

oysteoh:
What is the "default" job which has failed?

oyvindeide:
> What is the "default" job which has failed?

I´ve no idea, it suddenly appeared on a bunch of jobs, I suspect @jondequinor?

PR #144, opened 2020-08-24
Fix crash when no well data is asked for

berland:
gendata_rft will crash (in pandas.concat()) if the list of trajectory points is empty.

### DIFF
--- a/semeio/jobs/rft/gendata_rft.py
+++ b/semeio/jobs/rft/gendata_rft.py
@@ -133,6 +133,10 @@ def run(
     csvfile=None,
 ):
     dframes = []
+
+    if not well_times:
+        raise ValueError("No RFT data requested")
+
     for well, time, report_step in well_times:
         logger.debug(
             "Collecting RFT for well: {} at date: {}, report step: {}".format(
@@ -161,5 +165,5 @@ def run(
     if csvfile is not None:
         pd.concat(dframes, ignore_index=True, sort=False).to_csv(csvfile, index=None)
 
-    with open("GENDATA_RFT.OK", "w") as fh:
-        fh.write("GENDATA RFT completed OK")
+    if len(dframes) < len(well_times):
+        raise ValueError("Failed to extract requested RFT data")
--- a/semeio/jobs/scripts/gendata_rft.py
+++ b/semeio/jobs/scripts/gendata_rft.py
@@ -128,16 +128,20 @@ def main_entry_point(args=None):
 
     logger.info("All files loaded\nRetrieving RFT data...")
 
-    gendata_rft.run(
-        well_times=options.well_and_time_file,
-        trajectories=trajectories,
-        ecl_grid=options.eclbase[0],
-        ecl_rft=options.eclbase[1],
-        zonemap=options.zonemap,
-        csvfile=options.csvfile,
-    )
-
-    logger.info("Completed!")
+    try:
+        gendata_rft.run(
+            well_times=options.well_and_time_file,
+            trajectories=trajectories,
+            ecl_grid=options.eclbase[0],
+            ecl_rft=options.eclbase[1],
+            zonemap=options.zonemap,
+            csvfile=options.csvfile,
+        )
+        with open("GENDATA_RFT.OK", "w") as fh:
+            fh.write("GENDATA RFT completed OK")
+        logger.info("Completed!")
+    except ValueError as exception:
+        logger.error("Failed with error message: {}".format(exception))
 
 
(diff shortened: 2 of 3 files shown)


ertomatic:

berland:
(This crash has not been reported/observed in the wild, only in local testing)

oyvindeide:
The changes looks fine, and my question is due to not knowing this job that well. I see that `trajectory_path` is a required argument, but does not this change make it non-required?

berland:
No, this will just handle a file being empty.

oyvindeide:
It will handle an empty wells file? Is that not a symptom of something else being wrong? The reason for asking is that now the user can set `--csvfile`, but not actually get a file out, with no explanation of why.

Should we not rather enforce non-empty wells and trajectory files? Or if they are not required, make them not required?

berland:
I think this also covers cases where there is something inconsistent between coordinates and grids, so that even if the file is non-empty, the list `trajectory_points` become empty. The original code tests for this in line 142, and skips without warning in this case. The PR warns in this case, and ensures that the code will not crash (this crash did not occur prior to the csvfile feature being merged).

oyvindeide:
But isnt this another, and real problem, that we should account for: 
> I think this also covers cases where there is something inconsistent between coordinates and grids, so that even if the file is non-empty, the list trajectory_points become empty. The original code tests for this in line 142, and skips without warning in this case.

I must say that I am for crashing if something is not configured properly, instead of silently carrying on. So maybe this should have been a crash in the original code?

berland:
Code changed, now will not write OK file in errored scenarios.

semeio/jobs/rft/gendata_rft.py
```diff
@@ -157,9 +160,12 @@ def run(
 
             # Aggregate dataframe for all wells and report steps.
             dframes.append(trajectory_df)
+        else:
+            failed = True
 
-    if csvfile is not None:
+    if csvfile is not None and dframes:
```
oyvindeide, line 166: How about wrapping this in a `try/except` and inject some logging if it fails, so it is easier for the user to debug?

semeio/jobs/rft/gendata_rft.py
```diff
@@ -157,9 +160,12 @@ def run(
...
         pd.concat(dframes, ignore_index=True, sort=False).to_csv(csvfile, index=None)
 
-    with open("GENDATA_RFT.OK", "w") as fh:
-        fh.write("GENDATA RFT completed OK")
+    if dframes and not failed:
```
oyvindeide, line 169: Isnt `if dframe` abduntant? If that check were to ever fail, the job will also be `failed = True`?

Would also be good to add some logging here perhaps, as to why the job has failed? If we can do that, can we also be more specific about why/where it fails?

One option here is to move the writing of the `OK` file one level up, to `main_entry_point`, then this part of the code can just raise errors. Then the function call can be wrapped in a try/except and handle the errors accordingly.

berland: The `if dframe` is needed there to cover the scenario where `well_times` is empty.

berland:
Code updated after review.

semeio/jobs/rft/gendata_rft.py
berland: Fixed.

semeio/jobs/rft/gendata_rft.py
berland: Moved OK file writing up one level.

semeio/jobs/rft/gendata_rft.py
```diff
@@ -158,8 +162,7 @@ def run(
             # Aggregate dataframe for all wells and report steps.
             dframes.append(trajectory_df)
 
-    if csvfile is not None:
-        pd.concat(dframes, ignore_index=True, sort=False).to_csv(csvfile, index=None)
+    pd.concat(dframes, ignore_index=True, sort=False).to_csv(csvfile, index=None)
```
oyvindeide, line 165: If no `csv` file is given, doesnt this write to `stdout`? If so, that fills the `stdout` files with some unwanted text.

oyvindeide (commented):
I think this looks really good now, just one comment about the `csvfile`.

semeio/jobs/rft/gendata_rft.py
berland: Yes. Should perhaps reinstate that `if`.

oyvindeide (approved):
LGTM!

Issue #149, opened 2020-08-27
[CSV_EXPORT2] Does not create parent directories for output file


berland:
csv_export2.py should create any necessary parent directories needed for its output file path. Its predecessor did (I think).

berland:
This is the example workflow configuration from the Reek dataset:
```
EXPORT_RUNPATH 0-24 | 0-0
CSV_EXPORT2 <RUNPATH_FILE> <CONFIG_PATH>/../output/results/<CASE_DIR>/spotfire/<CASE_DIR>.csv monthly FOPT* FGPT* FWPT* ROIP RGIP WOPT* WOPR*
```
which the directory typically does not exist prior to running the workflow as CASE_DIR may change for almost every run.

The forward model `MAKE_FOLDER` should not be used for this, as it would run on every node, and forward models should not touch `CONFIG_PATH`. Is there at all any make folder options for workflows?

berland:
Closed pr. discussion in #150

PR #150, opened 2020-08-27
CSV export to non-existing directory

berland:
Solves #149

### DIFF
--- a/semeio/jobs/csv_export2/csv_export2.py
+++ b/semeio/jobs/csv_export2/csv_export2.py
@@ -1,4 +1,4 @@
-import pathlib
+import os
 
 import pandas as pd
 
@@ -16,5 +16,8 @@ def csv_exporter(runpathfile, time_index, outputfile, column_keys=None):
     summary = ensemble_set.load_smry(time_index=time_index, column_keys=column_keys)
     parameters = ensemble_set.parameters
     summary_parameters = pd.merge(summary, parameters)
-    pathlib.Path(outputfile).parent.mkdir(parents=True, exist_ok=True)
+
+    output_dir = os.path.dirname(outputfile)
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
     summary_parameters.to_csv(outputfile, index=False)
--- a/semeio/jobs/csv_export2/csv_export2.py
+++ b/semeio/jobs/csv_export2/csv_export2.py
@@ -1,4 +1,7 @@
+import pathlib
+
 import pandas as pd
+
 from fmu import ensemble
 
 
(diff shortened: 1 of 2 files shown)


ertomatic:

oysteoh:
There should be a squash of the commits at least - other than that It looks fine ! But - we had a small discussion on slack which didn't conclude I think - @markusdregi - should we make directories or not?

oyvindeide:
The consensus was we should not make directories. The main argument was that making directories could mask different issues. A better solution would be to catch the error and provide good feedback. If there are strong counter arguments though, we are all ears.

A second point is I am against relying on `ert-statoil` test data, at least as the only means of testing something, and new tests should be created using open data.

markusdregi:
The conclusion yesterday was as @oyvindeide mentions. But the discussion can be reopened if you are opinionated ;) However, what I would like to do is for us to agree and then I'll write up some `FMU`-dev conventions. I think the users would benefit greatly from consistent behaviour in these cases...

A thought I had yesterday evening was whether we should make a workflow that creates the necessary folder structure at the beginning of the run?

```
CREATE_FOLDER_STRUCTURE ../input/folder_structure.yml --root=..
```

```yaml
# ../input/folder_structure

input:
  some_data:
  the_secret_sauce:
    ingredient1:
    ingredient2:
config:
output:
  exports:
  storage:
```

Could this be a solution?

oyvindeide:
I've seen that discussion before somewhere? Yammer perhaps? Were you a part of that, and remember the outcome? I've no strong opinion on the matter.

berland:
I added some details on the Reek configuration in the issue, #149. Since it is in Reek and used in the course, we should expect this to be widespread, and it does not work currently with CSV_EXPORT2.

Is there a clean way today to create a folder for a workflow? I am not aware.

markusdregi:
I think we have two options:
- implement the above job and install it both as a workflow and a forward model job to create a batch of folders. Could be useful and reduce the need for 20 steps in the forward model to create nothing but folders 🤷 
- reevaluate our decision and decide that workflows and jobs in general should all create the necessary folders to write their output.

Personally I'm leaning towards the first...

berland:
What should be our recommendation for workaround today? MAKE_FOLDER run as a FORWARD_MODEL?

markusdregi:
I can implement an MVP of the above `CREATE_FOLDER_STRUCTURE` if we believe it is a good idea. I think @eivindsm patched the HM tutorial by changing the path to something that already existed...

berland:
> A second point is I would like to avoid relying on `ert-statoil` test data, at least as the only means of testing something, and new tests should be created using open data.

Shifted this problem to #152

PR #151, opened 2020-08-27
Blacken with black 20.8b0

berland:
Going from black 19.10b0 to 20.8b0, black will modify the source code.

Could potentially pin the black version in CI setup.

### DIFF
--- a/semeio/communication/semeio_script.py
+++ b/semeio/communication/semeio_script.py
@@ -71,7 +71,11 @@ class SemeioScript(ErtScript):  # pylint: disable=too-few-public-methods
     def _get_output_dir(self):
         base_dir = self.ert().resConfig().model_config.getEnspath()
         base_dir = os.path.realpath(base_dir)
-        return os.path.join(base_dir, "reports", type(self).__name__,)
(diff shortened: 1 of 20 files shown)


ertomatic:

oyvindeide:
Jenkins: test this please

oyvindeide:
This looks great, thanks for doing it, do you know why it fails though? Black says one file would be reformatted.

berland:
Failed CI was due to missing local rebase.

Another thing, when `pre-commit` is installing black on CI, I think we are getting the `stable` branch straight from github. Should semeio be more in control of which black version to run?

oyvindeide:
> Failed CI was due to missing local rebase.
> 
> Another thing, when `pre-commit` is installing black on CI, I think we are getting the `stable` branch straight from github. Should semeio be more in control of which black version to run?

I am not sure what I think, it is quite annoying when this happens, but at the same time, we get a reminder to update it. I considered asking you to pin it in this PR, but decided against it.

oyvindeide (approved):
LGTM! Thanks for updating!

PR #153, opened 2020-08-28
First version - ensure folder structure

markusdregi:
Example of a script that can be used to setup folders, both in the forward model and for workflows.

```yml
# example_config.yml
a:
  b:
  c:
    d:
e:
f:
```

**Forward model usage:**
In the forward model the job can be used as a replacement for multiple calls to `MAKE_DIRECTORY`. By using it as:

```
FORWARD_MODEL ENSURE_FOLDER_STRUCTURE(<CONFIG>=<CONFIG_PATH>/example_config.yml)
```

Will create the folders `a`, `e` and `f` in the runpath, and in addition create `b` and `c` inside `a`, and last `d` inside `c`.

**Workflow usage:**
The job can also be used as a workflow to create necessary folders that are expected to exist later on in the run.

The user will then have to create a workflow, i.e. `INIT_EXPERIMENT` which they can hook in their setup.
```
ENSURE_FOLDER_STRUCTURE <CONFIG_PATH>/example_config.yml -r=<CONFIG_PATH>/..
```

The folder structure described will in this setup be created one level down from the configuration file location. Notice the `-r` (short for `--root`) that allows you to change the root for where the folders are to be created. It is defaulted to cwd, which is why it will create it directly in the runpath in the forward model case.

Also, notice that it will not touch any existing files, only create the missing directories (if there are any missing). Hence, the name `ensure...`.

### DIFF
--- a/semeio/jobs/ensure_folder_structure/__init__.py
+++ b/semeio/jobs/ensure_folder_structure/__init__.py
@@ -1,0 +1,1 @@
+from semeio.jobs.ensure_folder_structure.job import ensure_folder_structure as run
--- a/semeio/jobs/ensure_folder_structure/job.py
+++ b/semeio/jobs/ensure_folder_structure/job.py
@@ -1,0 +1,35 @@
+import logging
+import os
+
+
+def _validate_path_element(path_elem):
+    if os.path.isabs(path_elem):
+        raise ValueError(
+            "Configured structure cannot contain absolute elements."
+            "Found: {}".format(path_elem)
+        )
+    if ".." in path_elem:
+        raise ValueError(
+            "Path elements should not contain '..'."
+            "Found: {}".format(path_elem)
+        )
+
+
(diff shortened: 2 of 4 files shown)


berland:
Possible name shortening: `ENSURE_FOLDER_STRUCTURE` -> `ENSURE_FOLDERS`

berland:
I am unsure whether users will see a requirement to put stuff into yet another file as a simplification. Awaiting an ERT 3 config file which I suppose could easier inline the folder structure, would it also be possible to supply say a comma-separated list of directories to ensure (so the job would either accept a yaml file or a string (or both)).

berland:
> ...  would it also be possible to supply say a comma-separated list of directories to ensure (so the job would either accept a yaml file or a string (or both)).

No, this belongs to a feature request for the current `MAKE_DIRECTORY`

markusdregi:
Closing due to unfit solution

Issue #106, opened 2020-05-15
[GENDATA_RFT] Extracts saturations


berland:
Gendata_rft can be extended to also extract saturations from the same `.RFT` file, this functionality has existed in a fork `gendata_rftsat` of the original `gendata_rft`, and is in use by several assets.

Similar request in https://github.com/equinor/ecl2df/issues/41

berland:
https://github.com/equinor/subscript/issues/54

oyvindeide:
Closed in #154

PR #154, opened 2020-08-31
Output saturations from gendata_rft

berland:
Solves #106 

~This is WIP, no tests added yet.~

The Norne binary output files in the test dataset do not have the saturations included, so water and gas saturations will not be extracted for them - thus a new dataset must be prepared (Reek from ecl2df's test data works).  Comments to the PR is wanted before such an extra dataset is made.

This PR is based on the difference between the files

https://github.com/equinor/ert-statoil/blob/master/etc/ERT/Scripts/gendata_rft.py
https://github.com/equinor/ert-statoil/blob/master/etc/ERT/Scripts/gendata_rftsat.py

Norne dataset has "ecl_plt_cell" while Reek gived "ecl_rft_cell", when referring to: 
https://github.com/equinor/libecl/blob/f1443476bc784f5b1d052e33cf1ea3235c41a00a/python/ecl/rft/ecl_rft_cell.py

"ecl2csv rft" however returns saturations from the Norne dataset, for whatever reason.


ertomatic:

oyvindeide:
Sorry for responding a bit late. I will been a bit busy, but will respond as soon as possible.

semeio/jobs/rft/trajectory.py
```diff
@@ -65,11 +68,34 @@ def get_pressure(self):
...
+
+    def get_sgas(self):
+        if self.is_active():
+            return self.sgas
+        return -1
+
+    def get_soil(self):
+        if self.is_active():
+            return self.soil
+        return -1
```
oyvindeide, line 84: Wont rule out that I am blind, but are these used anywhere?

semeio/jobs/rft/gendata_rft.py
```diff
@@ -151,14 +161,16 @@ def run(
...
-            _write_gen_data_files(trajectory_df, "RFT_{}_{}".format(well, report_step))
+            _write_gen_data_files(trajectory_df, well, report_step)
 
             # Aggregate dataframe for all wells and report steps.
             dframes.append(trajectory_df)
+        else:
+            logger.warning("No trajectory points loaded")
 
-    if csvfile is not None:
+    if csvfile is not None and dframes:
```
oyvindeide, line 173: Can you rebase this PR on master?

semeio/jobs/rft/trajectory.py
berland: Not currently in semeio, was just added so the API was analogous as `get_pressure()` (which is used once in semeio, in a test). There are user scripts around that might end up using this API, unless we can make then redundant.

berland:
Rebased. Note that Reek datafiles are in the commit, but not used yet.

semeio/jobs/rft/trajectory.py
berland: Removed. We should redirect users to the CSV output.

oyvindeide:
Having a look now, are the reek files ok to have publicly (presumably as they are from ecl2df)?

berland:
> Having a look now, are the reek files ok to have publicly (presumably as they are from ecl2df)?

Yes, these are already published with an open license.

tests/jobs/rft/test_gendata_rft.py
```diff
@@ -117,12 +137,32 @@ def test_gendata_rft_csv(tmpdir, input_data):
     assert (~dframe[~dframe["is_active"]]["inactive_info"].isnull()).all()
 
 
-def test_gendata_rft_directory(tmpdir, input_data):
+def test_gendata_rft_csv_reek(tmpdir, reek_data):
```
oyvindeide, line 140: Good with an integration test, but would like to see an integration test of `update_simdata_from_rft` as well.

oyvindeide (commented):
Changes look good, some minor comments.

semeio/jobs/rft/trajectory.py
```diff
@@ -65,11 +67,17 @@ def get_pressure(self):
...
         if self.grid_ijk:
             rftcell = rft.ijkget(self.grid_ijk)
             if rftcell:
                 self.pressure = rftcell.pressure
+                if rftcell.TYPE_NAME == "ecl_rft_cell":
+                    self.swat = rftcell.swat
+                    self.sgas = rftcell.sgas
+                else:
+                    self.swat = None
+                    self.sgas = None
```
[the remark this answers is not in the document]
berland: Fixed.

semeio/jobs/rft/trajectory.py
```diff
@@ -126,6 +134,9 @@ def to_dataframe(self, zonemap=None):
         elif "true_vertical_depth" in dframe:
             dframe = dframe.sort_values("true_vertical_depth")
 
+        if "swat" in dframe and "sgas" in dframe and "soil" not in dframe:
```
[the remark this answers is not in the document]
berland: Moved to  `update_simdata_from_rft`. I think I put it in the pandas code to be more robust  with regards to two-phase Eclipse runs  (only oil  and water, no gas).  But I am not sure if gendata_rft will work at all for two-phase (maybe  `sgas`  becomes `None`, maybe an exception is raised). That belongs in another issue.

tests/jobs/rft/test_gendata_rft.py
berland: Wouldn't that be a unit test? I skipped it because there were none such already. Is there enough difference between the integration test and such a unit test to warrant having that extra test layer?

oyvindeide: I meant unit test, was just about to correct myself now ;) I also noticed there was no unit test, and would be good to have one, as it makes debugging easier.

semeio/jobs/rft/trajectory.py
```diff
@@ -65,11 +67,17 @@ def get_pressure(self):
             return self.pressure
         return -1
 
-    def update_pressure_from_rft(self, rft):
+    def update_simdata_from_rft(self, rft):
         if self.grid_ijk:
             rftcell = rft.ijkget(self.grid_ijk)
             if rftcell:
                 self.pressure = rftcell.pressure
+                if rftcell.TYPE_NAME == "ecl_rft_cell":
```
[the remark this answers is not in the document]
berland: Rewritten.

semeio/jobs/rft/gendata_rft.py
```diff
@@ -24,24 +24,48 @@ def _write_gen_data_files(trajectory_df, fname):
...
-    _write_inactive_info(fname + "_inactive_info", trajectory_df)
 
-
-def _write_pressure(fname, trajectory_df):
+    _write_simdata(
+        os.path.join(directory, "RFT_{}_{}".format(well, report_step)),
+        "pressure",
+        trajectory_df,
+    )
+    if "swat" in trajectory_df:
```
oyvindeide, line 37: You are making this test both inside and outside `_write_simdata`. I get why, but is it needed? The flip side is that without this check here, there api is more predictable (instead of outputting a variable number of files).

semeio/jobs/rft/gendata_rft.py
```diff
@@ -158,17 +182,16 @@ def run(
...
-            )
+            _write_gen_data_files(trajectory_df, outputdirectory, well, report_step)
 
             # Aggregate dataframe for all wells and report steps.
             dframes.append(trajectory_df)
+        else:
+            logger.warning("No trajectory points loaded")
 
-    if csvfile is not None:
+    if csvfile is not None and dframes:
```
oyvindeide, line 194: Should we warn if the user has specified `csvfile` but one can not be provided? Or alternatively, move this below the check on line 174?

berland: Yes, it should be moved down.

berland: Or, maybe not. When the ValueError hits two lines down, it would be nice to have the CSV file for debugging.

tests/jobs/rft/test_gendata_rft.py
berland: Added in `test_trajectory_rftfile.py`.

berland:
@anders-kiaer, are you able to have a look?

berland:
Unanswered questions:
* Should SWAT-files we written even if there is no saturations available, as we do for pressure? (with -1). For usage with AHM, that is maybe how it must be done?
* I changed SATW and SATG to SWAT and SGAS, compared to the original gendata_rftsat code. I guess this only has to link with the filenames specified in GEN_DATA in ert config, so it should be fine to avoid non-conforming SATW terms.

semeio/jobs/rft/gendata_rft.py
berland: `and dframes` is now  reinstated, and a new test warrants its presence.

berland:
Will squash eventually, but keeping individual commits until review is finished.

semeio/jobs/rft/gendata_rft.py
berland: `_write_simdata` is now predictable (or it emits error messages), and `_write_gen_data_files` outputs a variable number of files, but always the pressure file.

berland:
Should I squash?

I think we can leave this as is, and get it into Komodo. 

We should make more interactive tests with full AHM setups to see if this (saturation matching) actually works as intended, when in a Komodo release.

rnyb:
> Should I squash?
> 
> I think we can leave this as is, and get it into Komodo.
> 
> We should make more interactive tests with full AHM setups to see if this (saturation matching) actually works as intended, when in a Komodo release.

Yes, just go for it

EivindBakken:
Sorry for being a bit offline - too many other tasks to dive into the important details. I notice the request on Slack for testing: It would be good to get Daniel MV on it: Smørbukk Sør is where this was developed and part of Daniels job towards Åsgard is to brush up their set-up to make it work & according to the latest standards. And I believe he has the skills to dive into it. What other fields are using saturation observations, @rnyb ? I proposed it for Snorre last week, but not sure who is going to to the model update (a fit-for-purpose update with improved representation of gas flooding in a segment that is the target of well from a potential new template). Maybe someone with experience about the method could assist them while also testing?

semeio/jobs/rft/trajectory.py
```diff
@@ -61,15 +101,38 @@ def inactive_info(self, zonemap=None):
...
+        set upfront.
+
+        Args:
+            rftfile (EclRFTFile)
+        """
         if self.grid_ijk:
-            rftcell = rft.ijkget(self.grid_ijk)
-            if rftcell:
+            rftcell = rftfile.ijkget(self.grid_ijk)
+            if isinstance(rftcell, EclPLTCell):
```
oyvindeide, line 124: Can you change this back to `if rftcell` and just do `if` on `EclRFTCell` to avoid repeating?

tests/jobs/rft/conftest.py
```diff
@@ -0,0 +1,102 @@
...
+def get_ecl_base_reek():
+    return ECL_BASE_REEK
+
+
+def get_expected_results_path_norne():
+    return EXPECTED_RESULTS_PATH_NORNE
+
+
+def get_mock_data_content_norne():
+    return MOCK_DATA_CONTENT_NORNE
```
oyvindeide, line 40: Return a copy to be on the safe side?

semeio/jobs/rft/gendata_rft.py
```diff
@@ -158,18 +183,19 @@ def run(
...
-                trajectory_df,
-                os.path.join(outputdirectory, "RFT_{}_{}".format(well, report_step)),
-            )
+            _write_gen_data_files(trajectory_df, outputdirectory, well, report_step)
 
             # Aggregate dataframe for all wells and report steps.
             dframes.append(trajectory_df)
+        else:
+            logger.warning(
+                "No trajectory points for well {} at date: {} found".format(well, time)
```
oyvindeide, line 194: Should be changed to `logger.error`? Because if this happens, the script will throw a `ValueError`

semeio/jobs/rft/gendata_rft.py
```diff
@@ -158,18 +183,19 @@ def run(
...
+            )
 
-    if csvfile is not None:
+    if csvfile is not None and dframes:
         pd.concat(dframes, ignore_index=True, sort=False).to_csv(csvfile, index=None)
 
     if len(dframes) < len(well_times):
-        raise ValueError("Failed to extract requested RFT data")
+        raise ValueError("Failed to extract all requested RFT data")
```
oyvindeide, line 201: Should we catch these in the entry point and exit more gracefully? Still crash, but with a message instead of a traceback.

semeio/jobs/rft/trajectory.py
```diff
@@ -25,14 +35,31 @@ def __init__(self, utm_x, utm_y, measured_depth, true_vertical_depth, zone=None)
...
         self.zone = zone
-        self.grid_ijk = None
+        self.grid_ijk = None  # tuple
         self.pressure = None
+        self.swat = None
+        self.sgas = None
+        self.soil = None
         self.valid_zone = False
 
     def set_ijk(self, point):
```
oyvindeide, line 45: Slightly outside the scope, but should this be a `property` with validation on the `setter`? Can be a separate issue  as you are just updating the doc string.

oyvindeide (commented):
Looking good, some minor comments (some slightly outside the scope of this PR, possible future issues)

semeio/jobs/rft/trajectory.py
berland: Fixed.

tests/jobs/rft/conftest.py
berland: Fixed.

semeio/jobs/rft/gendata_rft.py
berland: Fixed.

semeio/jobs/rft/gendata_rft.py
berland: Good idea. Changed it.

oyvindeide:
Looks good, can you update the failing tests, and it should be good to go. Need to swap `ValueError` for `SystemExit` and check the log for the error instead of the error itself. And `test_gendata_rft_csv_reek` is failing for some reason?

semeio/jobs/rft/trajectory.py
berland: I think I will leave this for later.

berland:
Squashed and ready.

oyvindeide (approved):
LGTM!

PR #155, opened 2020-09-02
Drop py2

oyvindeide:


oyvindeide:
The Jenkins failure is expected as that runs a matrix including `py2`. I suggest ignoring that for now as when `py2` support is dropped, that problem will fix itself.

test_requirements.txt
```diff
@@ -1,7 +1,2 @@
-pytest==5.4.3; python_version>="3.6"  # Fixed in https://github.com/pytest-dev/pytest/pull/7565
-pytest; python_version<="2.7"
-mock
-zipp<2; python_version<='2.7'
-configparser<5; python_version<='2.7'
```
lars-petter-hauge, line 5: Why where these in place at all? `mock` isn't used? Configparser only for py2?

tests/communication/test_integration.py
```diff
@@ -7,10 +6,7 @@
 import sys
 from semeio.communication import SEMEIOSCRIPT_LOG_FILE
 
-if sys.version_info >= (3, 3):
-    which = shutil.which
-else:
-    which = distutils.spawn.find_executable
+which = shutil.which
```
lars-petter-hauge, line 9: no need to assign the function anymore, just use `shutil.which` where it is supposed to be used.

tests/jobs/misfit_preprocessor/test_integration.py
```diff
@@ -48,7 +44,7 @@ def test_misfit_preprocessor_main_entry_point_gen_data(monkeypatch):
     # call_args is a call object, which itself is a tuple of args and kwargs.
     # In this case, we want args, and the first element of the arguments, which
     # again is a tuple containing the configuration which is a list of configs.
-    assert len(run_mock.call_args[0][0]) == 47, "wrong number of clusters"
+    assert len(list(run_mock.call_args)[0][0]) == 47, "wrong number of clusters"
```
lars-petter-hauge, line 47: Don't understand why we all of a sudden needs `list`, but okay..? :P

lars-petter-hauge:
Single comment that should be addressed, other than that this is good to go!

Aaaah, very nice to get rid of python2 from the pipeline! Thank you!

tests/jobs/misfit_preprocessor/test_integration.py
oyvindeide: A pylint issue, could opt to ignore the issue instead if you would prefer? `unsubscriptable` error

test_requirements.txt
oyvindeide: To be honest, no, I dont know. I suspect they are remnants from the creation of this package.

tests/jobs/misfit_preprocessor/test_integration.py
lars-petter-hauge: I would almost argue that it's better to include the `ignore unsubscriptable` - as that will highlight all workarounds necessary with pylint, instead of hiding it in the code.

But I'll leave it up to you to decide

lars-petter-hauge (approved):

PR #156, opened 2020-09-03
Add documentation to black_box_job

### LINKED ISSUE #109: Write documentation of black box job
#70 introduced a workflow for clustering and scaling correlated observations. This workflow needs documentation. A good starting point is probably to run it through the documentation generator that will come in configsuite 0.6.

oyvindeide:
Closes #109

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ setup(
         "ert": [
             "semeio_jobs = semeio.hook_implementations.jobs",
             "SpearmanCorrelation = semeio.jobs.scripts.spearman_correlation",
+            "MisfitPreprocessor = semeio.jobs.scripts.misfit_preprocessor",
         ]
     },
     license="GPL-3.0",
--- a/tests/hook_implementations/test_hook_implementations.py
+++ b/tests/hook_implementations/test_hook_implementations.py
@@ -2,13 +2,17 @@
 
 import os
 import semeio.hook_implementations.jobs
-from semeio.jobs.scripts import spearman_correlation
+from semeio.jobs.scripts import spearman_correlation, misfit_preprocessor
 from ert_shared.plugins.plugin_manager import ErtPluginManager
 
 
 def test_hook_implementations():
     pm = ErtPluginManager(
-        plugins=[semeio.hook_implementations.jobs, spearman_correlation]
+        plugins=[
+            semeio.hook_implementations.jobs,
+            spearman_correlation,
+            misfit_preprocessor,
+        ]
     )
 
     expected_jobs = {
--- a/semeio/jobs/config_workflow_jobs/MISFIT_PREPROCESSOR
+++ b/semeio/jobs/config_workflow_jobs/MISFIT_PREPROCESSOR
@@ -1,2 +1,0 @@
-INTERNAL    True
-SCRIPT      ../scripts/misfit_preprocessor.py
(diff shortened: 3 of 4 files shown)


jondequinor:
How does black_box_job and misfit_preprocessor fit together?

Title needs some work, seems you are migrating misfti_prep to the new plugin system, not adding docs. That's one side-effect.

setup.py
```diff
@@ -25,7 +25,12 @@ def package_files(directory):
     url="https://github.com/equinor/semeio",
     description="Jobs and workflow jobs for Ert.",
     packages=find_packages(include=["semeio*"], exclude=["semeio.jobs.scripts"]),
-    entry_points={"ert": ["semeio_jobs = semeio.hook_implementations.jobs"]},
+    entry_points={
+        "ert": [
+            "semeio_jobs = semeio.hook_implementations.jobs",
+            "MisfitPreprocessor = semeio.jobs.scripts.misfit_preprocessor",
```
hnformentin, line 31: Why do we need to have this line here? Isn't it redundant with `scripts = glob("semeio/jobs/scripts/[!__]*.py")`?

oyvindeide: We probably could, but I am not convinced we should do it implicitly. The main drawbacks in my view is having it explicitly allows anyone to quickly tell how many entry points we have, and where they are. The second is that by `glob`ing, we have to keep all the jobs in the same place. I would prefer we dont any more, and put everything related to one workflow in a separate module. That makes it much easier to navigate.

oyvindeide:
> How does black_box_job and misfit_preprocessor fit together?

The black box job is the default settings of the misfit_preprocessor. 

> Title needs some work, seems you are migrating misfti_prep to the new plugin system, not adding docs. That's one side-effect.

Sort of agreed (though the side effect is the main goal), I will expand on the commit text.

setup.py
hnformentin: I understand, do you think we could let explicit the other jobs declared in the glob, to be consistent overall? 
Maybe in a different commit, but same PR?

hnformentin:
Worked with @lars-petter-hauge @kvashchuka and @oyvindeide to review this PR.
We pulled the PR, built the documentation, verified the output and made tests before commenting.

setup.py
oyvindeide: I am not quite sure I understand what you mean? Do you mean the forward model jobs? The other workflows have to add their own entry point.

jondequinor (approved):
LGTM!

setup.py
hnformentin: I see. As we are introducing a new way to define entry points, I thought it would follow with the modification of the other entry points (so all have the same standard).

oyvindeide: I think we should do that to forward model jobs as well, but there is no system for that in `ert`, yet. For now we will do it for all the workflow jobs, but they will add their own entry point as they are updated from the old to the new spec.

tests/hook_implementations/test_hook_implementations.py
```diff
@@ -35,7 +38,6 @@ def test_hook_implementations():
     installable_workflow_jobs = pm.get_installable_workflow_jobs()
     for wf_name, wf_location in expected_workflow_jobs.items():
```
hnformentin, line 39: Maybe review the wf_location test.

PR #157, opened 2020-09-03
Document spearman correlation through plugin system

### LINKED ISSUE #10: Document spearman correlation job
Document the sc job via argparse. This setup can be the same as for spinningjenny, but the inclusion into ERT should be dynamic. I.e. the docs of spearman job should not be directly referred to anywhere in the ERT doc source code.

oyvindeide:
Closes #10

### DIFF
--- a/tests/hook_implementations/test_hook_implementations.py
+++ b/tests/hook_implementations/test_hook_implementations.py
@@ -1,9 +1,15 @@
+# pylint: disable=missing-docstring
+
+import os
 import semeio.hook_implementations.jobs
+from semeio.jobs.scripts import spearman_correlation
 from ert_shared.plugins.plugin_manager import ErtPluginManager
 
 
 def test_hook_implementations():
-    pm = ErtPluginManager(plugins=[semeio.hook_implementations.jobs])
+    pm = ErtPluginManager(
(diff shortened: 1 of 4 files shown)


jondequinor:
How will this document SC?

oyvindeide:
Can see that the PR title and commit message was a bit unclear, so I have changed that. Regarding docs, this was the issue is was working from:

> A good start is to document the job using the argparse and configsuite schema.

Not to say that we should not add more documentation in the future, but for now I think this is enough, especially considering most users will not call this job directly, but will call it through the misfit preprocessor job.

jondequinor (approved):
Lgtm!

tests/hook_implementations/test_hook_implementations.py
```diff
@@ -30,7 +33,6 @@ def test_hook_implementations():
     installable_workflow_jobs = pm.get_installable_workflow_jobs()
     for wf_name, wf_location in expected_workflow_jobs.items():
         assert wf_name in installable_workflow_jobs
```
hnformentin, line 35: wf_location may be unnecessary.

oyvindeide: Good point, because the exact path will now vary between runs (and is the domain of `ert`, not semeio), I've changed to asserting that it is a file instead.

PR #160, opened 2020-09-03
WIP: misfit_preprocessor job

### LINKED ISSUE #115: New clustering option in `misfit_preprocessor`
The `misfit_preprocessor` should expose the option of first running COS on the specified set of observations, then use the degrees of freedom as a cluster size parameter to the spearman clustering and then use those clusters again in COS.

This should be the default option if nothing else is specified.

xjules:
We should rename the current scaling job to PCAJob, while the actual ScalingJob should provide for observation scaling only.

There are still tests missing for missfit job!

Resolves #115


oyvindeide:
Superseded by #177, closing this one.

PR #161, opened 2020-09-03
Filter out records not of this thread

### LINKED ISSUE #104: Evaluate the logging of SemeioScript
In #99 `SemeioScript` adds a handler to the root logger to ensure that log statements is propagated via the reporter to the appropriate place. However, currently this most likely have the side effect that log statements from the ERT application will end up also in the workflow log. This should be avoided. At the same time we want log-statements from the workflow and the libraries it is utilising to end up in the same place.

We can probably use the thread id to make a filter that can be appended to the handler. But hopefully there is a more robust and elegant solution.

jondequinor:
Fixes #104 

The approach is described in the issue.

### DIFF
--- a/semeio/communication/semeio_script.py
+++ b/semeio/communication/semeio_script.py
@@ -1,7 +1,8 @@
 import datetime
 import logging
-from logging.handlers import BufferingHandler
 import os
+import threading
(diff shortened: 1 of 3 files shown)


sondreso (approved):
LGTM!

Issue #59, opened 2020-03-09
DESIGN_KW and GEN_KW working together


berland:
There is a syntax conflict in `parameters.txt` when parameters are variably set using design matrices (DESIGN2PARAMS) and GEN_KW.  For the Reek FMU template project, this has led to inclusion of a separate forward model (workaround: PARAMETERS_4DESIGN)/parameters_4designkw.py) that modifies `parameters.txt` by removing all prefixes in front of a colon pr. line 

Forward models used in Reek should all be considered included in semeio (or some repo) as this project is copied manually by users.

I am not very comfortable adding an extra forward model to workaround this incompatibility in a template project and we should seek a better solution. 

A possibility is that DESIGN_KW is changed so when a users wants to replace `<SOMEPARAMETER>`, it will give a hit in lines like `SOMEGROUP:SOMEPARAMETER 100` in `parameters.txt`. 

The PR for the pyscal forward model effectively does this using the function `rm_genkw_prefix()` 
https://github.com/equinor/semeio/pull/53/files#diff-4226388281c1739d36a044379a3fab60R171

Same discussions earlier:
https://www.yammer.com/statoil.com/#/Threads/show?threadId=1222922541
https://github.com/equinor/ert/issues/275
https://github.com/equinor/ert/issues/599

berland:
@jcrivenaes @rnyb

berland:
```
#!/usr/bin/env python
"""
Fix for handling multiple global variable files (ipl and yml) when
using GEN_KW update parameters.txt by appending a repeat of
unc. params in a DESIGN_KW friendly format with updated
parameters.txt file, DESIGN_KW can now be used to fill values in
multiple global variable files

Roger Nybo : Jan 2019

"""

from __future__ import print_function

# -----------------------------------
# paths are relative to ert <runpath>

filename = "parameters.txt"  # input file

# -----------------------------------
# Main
# -----------------------------------

print("---------------------------------------")
print(
    "Read parameters.txt file and append all uncertainty "
    + "parameters to file with GEN_KW friendly format"
)

with open(filename, "r") as myfile:
    all_lines = myfile.readlines()

try:
    infile_append = open(filename, "a")
    print("OK:" + filename + " file was found.\n")
except:
    print("-----\nNot OK: file " + filename + " was not found\n-----")

for line in all_lines:
    if not line.startswith("LOG") and ":" in line:
        x = line.split(":")
        infile_append.write(x[1])

infile_append.close()

print(
[...]

PR #163, opened 2020-09-04
Support missing namespace in parameter names in design_kw

berland:
First attempt at solving #59 

Check first if the behaviour imposed by the added tests is what we want, then assess whether the code achieving it was sufficiently elegant (perhaps it was not..)

### DIFF
--- a/semeio/jobs/design_kw/design_kw.py
+++ b/semeio/jobs/design_kw/design_kw.py
@@ -22,10 +22,13 @@ def run(
     _logger.setLevel(log_level)
 
     valid = True
+
     with open(parameters_file_name) as parameters_file:
         parameters = parameters_file.readlines()
 
     key_vals = extract_key_value(parameters)
+
+    key_vals.update(rm_genkw_prefix(key_vals))
 
     with open(template_file_name, "r") as template_file:
         template = template_file.readlines()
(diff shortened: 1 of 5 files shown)


ertomatic:

tests/jobs/design_kw/test_design_kw.py
```diff
@@ -78,6 +78,69 @@ def test_extract_key_value_dup_key():
...
+    "str_or_dict, expected",
+    [
+        ("foo", "foo"),
+        ("foo:bar", "bar"),
+        ("foo:bar:com", "bar:com"),
+        ("", ""),
+        ({"foo:bar": "value"}, {"bar": "value"}),
+        ({"foo:bar": "value", "foo:com": "value2"}, {"bar": "value", "com": "value2"}),
+        ({}, {}),
+        # Duplicates are dropped, a warning is logged.
```
markusdregi, line 91: I don't think any warning is logged here...

semeio/jobs/design_kw/design_kw.py
```diff
@@ -116,3 +119,53 @@ def extract_key_value(parameters):
     if errors:
         raise ValueError("\n".join(errors))
     return res
+
+
+def rm_genkw_prefix(str_or_dict, ignoreprefixes="LOG10_"):
```
markusdregi, line 124: Unless I've misunderstood the type of `str_or_dict` will always be a dict on the first call, and then the string support is used internally on the keys afterwards. I think the implementation would be much cleaner if we utilised this assumption and hid the string-support within the function such that `rm_genkw_prefix` only accepts dicts of depth one.

markusdregi (commented):
I think this is a fair approach and the tests seems adequate 👍 I would like to see an added test for logging if that is the expected behaviour and in addition I think the implementation of `rm_genkw_prefix` could be simplified a bit.

I'm supporting this direction

tests/jobs/design_kw/test_design_kw.py
berland: Added extra duplicates warning.

semeio/jobs/design_kw/design_kw.py
berland: New  implementation written - it got a lot better with only dict as input.

semeio/jobs/design_kw/design_kw.py
```diff
@@ -116,3 +119,55 @@ def extract_key_value(parameters):
     if errors:
         raise ValueError("\n".join(errors))
     return res
+
+
+def rm_genkw_prefix(paramsdict, ignoreprefixes="LOG10_"):
+    """Remove anything before the first colon in a string, or remove
```
markusdregi, line 125: The doc string still need an update, besides that it looks good 👍 Nice clean up of `rm_genkw_prefix` ;)

berland: Updated.

markusdregi (approved):
🚀

Issue #166, opened 2020-09-08
[design2params]  TRUE is mapped to True


berland:
A design matrix cell value with the string `TRUE`, is mapped into `True` in  `parameters.txt`.

This conflicts with the Py2 behaviour, and crashes some FMU workflows.

Credit: ilor

berland:
The culprit is  `pd.read_excel()` which converts to dtype `bool` when possible.

oyvindeide:
Solved in #167

PR #167, opened 2020-09-08
Fix TRUE/True handling in design2params

berland:
Solves #166


ertomatic:

oyvindeide (commented):
Changes looks good, but wouldn´t the test make more sense as a unit test of `_read_excel`? Then you dont need as much setup for it.

berland:
Sure, this justs reflects that the failing test was written before the cause of the error :)

berland:
On second thought,  I think I want it as a test at this level - to avoid errors from all the other lines of code that potentially touches the dataframe.

berland:
Other tests fail though: The change in this PR incurs that "1" in floating point columns in Excel are written as "1" in parameters.txt, while current code writes "1.0000...",  because Pandas has identified the column as floats.

berland:
Looking at the precursor code in ert-statoil, I find the following code:
```
       if str(cell_value) != "": 
            if worksheet.cell_type(datarow, current_column) == xlrd.XL_CELL_NUMBER:
                if cell_value.is_integer():
                    cell_value = int(cell_value)
            parameterfile.write(str(headername) + " " + str(cell_value) + "\n")
```
which I think would write a "1" in an Excel cell as "1" in `parameters.txt`.

oyvindeide:
> Looking at the precursor code in ert-statoil, I find the following code:
> 
> ```
>        if str(cell_value) != "": 
>             if worksheet.cell_type(datarow, current_column) == xlrd.XL_CELL_NUMBER:
>                 if cell_value.is_integer():
>                     cell_value = int(cell_value)
>             parameterfile.write(str(headername) + " " + str(cell_value) + "\n")
> ```
> 
> which I think would write a "1" in an Excel cell as "1" in `parameters.txt`.

I dont think I am going out on a limb when I assume that is not tested?

berland:
Few tests in previous version.. But it is perhaps what py2 users are running (and have adapted their workflows to)?

Consider breaking this into 2020.09.

berland:
The `dtype=str` change in this PR does not affect the floating point behaviour in the new parametrized test `test_single_cell_value()`, only the True/False behaviour.

berland:
The changes reflected in `refdesignmatrix.txt` can be revealed by the added `test_pair_cell_value` where one can see that this PR now changes certain "0.0" to "0" in `parameters.txt`

oysteoh (changes requested):
Changes look good to me - and good coverage with tests documenting the behavior the change introduces. Good work 👍 

The jenkins error i cannot investigate right now - having trouble with my vpn; do you know if it is a valid error or just jenkins which does not build correct at the moment?

Please squash changes and then i will approve ! 👍

berland:
Here is the same commit directly on top of 0.5.4 (this bug is a Py3 upgrade blocker for at least one FMU project):

https://github.com/berland/semeio/tree/patch-release-0.5.5

oysteoh (approved):

PR #168, opened 2020-09-08
Give examples of cos usage through plugin system

### LINKED ISSUE #9: Document scaling job
Document the scaling job via argparse. This setup can be the same as for spinningjenny, but the inclusion into ERT should be dynamic. I.e. the docs of ScalingJob should not be directly referred to anywhere in the ERT doc source code.

oyvindeide:
Most of what was suggesten in: https://github.com/equinor/semeio/pull/77 has already been implemented in https://github.com/equinor/semeio/pull/124 This PR takes the examples from #77 and passes them to `ert` through the plugin system.

Closes #9

### DIFF
--- a/semeio/workflows/correlated_observations_scaling/cos.py
+++ b/semeio/workflows/correlated_observations_scaling/cos.py
@@ -12,6 +12,121 @@ from semeio.workflows.correlated_observations_scaling import ScalingJob
 from semeio.workflows.correlated_observations_scaling.obs_utils import keys_with_data
 
 
+_DESCRIPTION = """
+    Correlated_observation_scaling is an ERT workflow job that does Principal
+    Component Analysis (PCA) scaling of observations in ERT. The job accepts a
+    configuration as the only argument.
+
+    A valid job configuration is a YAML file in one of three formats:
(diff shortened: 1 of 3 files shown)


semeio/workflows/correlated_observations_scaling/cos.py
```diff
@@ -12,6 +12,117 @@
...
+
+Correlated Observations Scaling Configuration
+=============================================
+{}
+""".format(
+    configsuite.docs.generate(job_config.build_schema())
+)
+
+
+_EXAMPLES = """
```
lars-petter-hauge, line 38: I mean.. Not sure if it is an overkill.. But if we split the examples out into individual strings, could we then create tests in order to ensure that the examples are in fact valid?

oyvindeide: I should be able to check that they are valid configs, but the context validation will be missing. Might be overkill, but you never know, good to have forced up to date docs.

lars-petter-hauge (approved):

jondequinor (approved):
:heartbeat:

Issue #136, opened 2020-08-07
Reports end up as a case


markusdregi:
The reporter puts the reports in a folder called `reports` within `ENSPATH`. After a run, a new case named `reports` shows up, which is not exactly as it is supposed to be. Let us see if we can either change the name (like making it a hidden folder) to avoid storage picking it up. If not, we'll have to find a different place to put the reports...

Steps to reproduce:
- Fetch the poly case
- Load and hook `MISFIT_PREPROCESSOR`
- Run ES
- Inspect the cases in the GUI

lars-petter-hauge:
Verified that renaming to `.reports` will mitigate the bug. Unsure if that's the optimal solution though. If the content of `.reports` eventually will only go to the new ert-storage, then I am satisfied with the temporary solution (although I am not sure if there are other consequences of adding anything to the `storage` folder).

sondreso:
Resolved by https://github.com/equinor/semeio/pull/169

PR #169, opened 2020-09-09
Move report folder

lars-petter-hauge:
Solves #136 

The report folder is added to the storage folder of ert. This comes with
the issue that it is regarded as an actual case in the eyes of ERT. ~By
hiding the folder we avoid the issue.~

### DIFF
--- a/semeio/communication/semeio_script.py
+++ b/semeio/communication/semeio_script.py
@@ -3,8 +3,11 @@ import logging
 import os
 import threading
 from logging.handlers import BufferingHandler
+from pathlib import Path
 from types import MethodType
+
 from res.enkf import ErtScript
+
 from semeio.communication.reporter import FileReporter
 
 SEMEIOSCRIPT_LOG_FILE = "workflow-log.txt"
@@ -72,11 +75,14 @@ class SemeioScript(ErtScript):  # pylint: disable=too-few-public-methods
         self.run = MethodType(run_with_handler, self)
 
     def _get_output_dir(self):
-        base_dir = self.ert().resConfig().model_config.getEnspath()
+        res_config = self.ert().resConfig()
+        base_dir = res_config.config_path
         base_dir = os.path.realpath(base_dir)
+        sub_dir = Path(res_config.user_config_file).stem
         return os.path.join(
             base_dir,
             "reports",
+            sub_dir,
             type(self).__name__,
         )
 
--- a/tests/communication/unit/test_semeio_script.py
+++ b/tests/communication/unit/test_semeio_script.py
@@ -9,14 +9,18 @@ from semeio.communication import SEMEIOSCRIPT_LOG_FILE, SemeioScript
 from unittest.mock import Mock
 
 
+def _ert_mock(config_path="config_path", user_config_file="config_file"):
+    resconfig_mock = Mock()
+    resconfig_mock.config_path = config_path
+    resconfig_mock.user_config_file = user_config_file
+
+    return Mock(resConfig=Mock(return_value=resconfig_mock))
+
+
 def test_semeio_script_publish(tmpdir):
     tmpdir.chdir()
-    enspath = "enspath"
-    ert = Mock(
-        resConfig=Mock(
-            return_value=Mock(model_config=Mock(getEnspath=Mock(return_value=enspath)))
-        )
-    )
+    config_path, user_config_file = "config_path", "config_file"
+    ert = _ert_mock(config_path, user_config_file)
 
     namespace = "arbitrary_data"
     data = [1, 2, 3, 4, "This is a very important string"]
@@ -29,7 +33,11 @@ def test_semeio_script_publish(tmpdir):
     my_super_script.run()
 
     expected_outputfile = os.path.join(
-        enspath, "reports", MySuperScript.__name__, namespace + ".json"
+        config_path,
+        "reports",
+        user_config_file,
+        MySuperScript.__name__,
+        namespace + ".json",
     )
 
     with open(expected_outputfile) as f:
@@ -40,12 +48,8 @@ def test_semeio_script_publish(tmpdir):
 
 def test_semeio_script_logging(tmpdir):
     tmpdir.chdir()
-    enspath = "enspath"
-    ert = Mock(
-        resConfig=Mock(
-            return_value=Mock(model_config=Mock(getEnspath=Mock(return_value=enspath)))
-        )
-    )
+    config_path, user_config_file = "config_path", "config_file"
+    ert = _ert_mock(config_path, user_config_file)
 
     msg = "My logging msg"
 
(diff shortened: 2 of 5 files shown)


oyvindeide:
Is this really a good solution though? This is where the jobs put their output, we are not exactly making it accessible when doing it like this.

sondreso:
I am also skeptical to this! I am fairly certain the case is still available in the ERT "backend", and worried that it might cause weird bugs if we (by accident) try to load it sometime. 

> We are not exactly making it accessible when doing it like this.

Should these files be accessible though? Isn't the plan to try to move away from dependencies on specific directory structures?

oyvindeide:
> Should these files be accessible though? Isn't the plan to try to move away from dependencies on specific directory structures?

That is the plan, yes, but the reason this writes files is so we have an interim solution. So they should be semi accessible at least (not hidden).

lars-petter-hauge:
I also am sceptical, as mentioned in the issue. Would it be better to dump it next to the configfile though? I don't like that either to be honest, as users typically have multiple config files in the same folder, thus we'll be adding a bunch of logs from different configuration runs into the same folder.

sondreso:
> That is the plan, yes, but the reason this writes files is so we have an interim solution. So they should be semi accessible at least (not hidden).

When we talk about accessible, do we mean to the users or to some tooling that we are building? 

> Would it be better to dump it next to the configfile though? I don't like that either to be honest, as users typically have multiple config files in the same folder, thus we'll be adding a bunch of logs from different configuration runs into the same folder.

We could create subfolders for each config name in some `.report` folder next to the config (or hash of the config name, depending on the answer to the question above).

oyvindeide:
> When we talk about accessible, do we mean to the users or to some tooling that we are building?

To the user, which is the reason this was created in the first place, we should not communicate through files.

lars-petter-hauge:
So `<CONFIG_PATH>/reports/<CONFIG_NAME>/`?

sondreso:
> So `<CONFIG_PATH>/reports/<CONFIG_NAME>/`?

That was my suggestion yes, but I'm far from convinced that it's the best!

lars-petter-hauge:
I'd say it's better than what we have now though, which is definitively screwing up storage :P

lars-petter-hauge:
Most of the changes are related to the tests, and I have to say my mocking abilities are not the best :)

If there's a better way of creating that mock object feel free to shout out!

semeio/communication/semeio_script.py
```diff
@@ -72,11 +75,14 @@ def run_with_handler(self, *args, **kwargs):
         self.run = MethodType(run_with_handler, self)
 
     def _get_output_dir(self):
-        base_dir = self.ert().resConfig().model_config.getEnspath()
+        res_config = self.ert().resConfig()
+        base_dir = res_config.config_path
```
oyvindeide, line 79: This could probably be added to the `libres facade`, that requires a PR in `ert`, but think it is worth it, especially if we think this will be an option in the new API.

tests/communication/unit/test_semeio_script.py
```diff
@@ -6,18 +6,27 @@
...
 
-from unittest.mock import Mock
+from unittest.mock import Mock, PropertyMock
 
+def _ert_mock(config_path="config_path", user_config_file="config_file"):
+    resconfig_mock = Mock()
+    config_path_prop_mock = PropertyMock(return_value=config_path)
+    user_config_prop_mock = PropertyMock(return_value=user_config_file)
+    type(resconfig_mock).config_path = config_path_prop_mock
+    type(resconfig_mock).user_config_file = user_config_prop_mock
```
oyvindeide, line 16: Do you intend to do this:
```suggestion
    resconfig_mock = Mock()
    resconfig_mock.config_path = config_path
    resconfig_mock.user_config_file = user_config_file

    return Mock(
        resConfig=Mock(
            return_value=resconfig_mock
        )
    )
```

lars-petter-hauge: I was given the impression from the docs that one should use `PropertyMock` instead of modifying the `__set__`'er.

But it didn't look particularly nice when one has to use the `type` in order to set the `PropertyMock` on the `Mock` itself (which again, is from the documentation)..

lars-petter-hauge: I'll change it though to the proposed example.

oyvindeide: Depends what you want to do - do you want to assert on the property? Then yes, definitely use `PropertyMock`

lars-petter-hauge: As in, assert that it has been called? Not doing it for now at least.

lars-petter-hauge:
We are at the "one black version against another" again it seems :P 

But worse, the black version in travis makes the lines too long for pylint. So I can't really use the same version og black by the looks of it..

lars-petter-hauge:
Alright, gotten as far as I should I suppose.

semeio/communication/semeio_script.py
lars-petter-hauge: I am slightly hesitant to add more to the `libres_facade` because we're using it in this PR. I do understand you that the argument is it useful enough to be part of the new API, and as such it's not "just because we need it here". But to be honest, I don't know the future of the `libres_facade`, nor how the API will evolve.

Unless there's a strong reason I say we should leave it. Possibly create an issue in ERT that would discuss around the `libres_facade` and argue that at least we've found a reason to include this option.

oyvindeide: I would be in the favor of at least creating an issue, because part of the point of `libres_facade` is to highlight what functions we are actually using?

oyvindeide:
It works like this, but I was wondering, is there any discussion in the original implementation regarding the location of the reports, or the rationale for putting it in `storage`?

lars-petter-hauge:
I don't think the new storage is at the point where it allows us to dump arbitrary data just yet. As of to date I believe the only thing it is possible to add to the storage outside of what ert itself does, is comments to observations.

But allowing forward models and workflows to post data themselves is on the agenda I believe (either directly, or through a messaging system).

semeio/communication/semeio_script.py
lars-petter-hauge: Agreed

lars-petter-hauge: Or, now looking at the facade I see that it all has to do with internal memory related information. What we are using here is completely disk related, and disk relation is something we want to avoid. I would argue that it would be against the future implementation of libres to add any functionality related to disk.

What we instead want is to use the new storage for all of this once that is ready, and completely avoid anything to disk. And instead, if users need it, have a solution for them to export the data to disk and then have them explicitly state the location.

oyvindeide:
> I don't think the new storage is at the point where it allows us to dump arbitrary data just yet. As of to date I believe the only thing it is possible to add to the storage outside of what ert itself does, is comments to observations.
> 
> But allowing forward models and workflows to post data themselves is on the agenda I believe (either directly, or through a messaging system).

No, and was not quite what I meant, was wondering if there was a discussion when this was originally implemented as to the location of the files that could be relevant.

oyvindeide (approved):
LGTM!

PR #171, opened 2020-09-10
Move workflows

oyvindeide:


jondequinor (changes requested):
I think this should be in the job itself, so that the py27 job doesn't start at all.

oyvindeide:
> I think this should be in the job itself, so that the py27 job doesn't start at all.

Agreed, I created #172 as it is unrelated to this PR and removed the commit.

jondequinor (approved):

PR #173, opened 2020-09-10
Refactor scaling factor calculation

### LINKED ISSUE #143: Unused code and function
This is never used:

https://github.com/equinor/semeio/blob/2d7263f7cd0291550004b29dc9949f2d300ca256/semeio/jobs/correlated_observations_scaling/job.py#L40

As such we should probably delete it and clean up the code so we don`t have duplicate ways of doing the same thing.

oyvindeide:
Closes #143

### DIFF
--- a/semeio/workflows/correlated_observations_scaling/job.py
+++ b/semeio/workflows/correlated_observations_scaling/job.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
 import configsuite
-import numpy as np
 from semeio.workflows.correlated_observations_scaling import job_config
 from semeio.workflows.correlated_observations_scaling.exceptions import ValidationError
 from semeio.workflows.correlated_observations_scaling.obs_utils import (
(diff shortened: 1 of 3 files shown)


semeio/workflows/correlated_observations_scaling/scaled_matrix.py
```diff
@@ -61,9 +57,6 @@ def get_scaling_factor(self, events):
                 below a user threshold
         """
         data_matrix = self.get_data_matrix()
-        nr_components, singular_values = self.get_nr_primary_components(
-            threshold=events.threshold
-        )
         nr_observations = data_matrix.shape[1]
 
         logging.info("Scaling factor calculated from {}".format(events.keys))
```
lars-petter-hauge, line 62: It is a bit weird that the only reason for including `events` in the function is to add this very general log statement. Suppose it could've been moved to outside of the function call and rather logged on the other side?

oyvindeide: You raise a valid point, a bit besides the point of the issue, but I will fix it.

semeio/workflows/correlated_observations_scaling/job.py
```diff
@@ -50,18 +49,15 @@ def scale(self, measured_data):
         matrix = DataMatrix(measured_data.data)
         matrix.normalize_by_std()
 
-        scale_factor = matrix.get_scaling_factor(config.CALCULATE_KEYS)
         events = config.CALCULATE_KEYS
-        data_matrix = matrix.get_data_matrix()
         nr_components, singular_values = matrix.get_nr_primary_components(
```
lars-petter-hauge, line 53: This is more of a question than anything else. Does it make sense to have the `DataMatrix` instance store the `nr_components`? And in that case it is not necessary to keep something that is strictly related to the instance and feed it back again. If it wasn't for the `self._reporter.publish` of `singular_values` I would think that the `get_scaling_factor` could've checked if the `nr_components` was calculated, and if not, calculate them..

Not a big concern though, this PR already improves the code flow.

oyvindeide: Dont think that makes sense, because then the `DataMatrix`is more locked to doing only this kind of scaling. I think it makes more sense then to move the `get_scaling_factor` to a stand-alone function in that case.

oyvindeide: Turns out it was already logged on the other side.

lars-petter-hauge: Let's not include it to the `DataMatrix` at least then.

lars-petter-hauge (approved):
Improves flow and removes unused code.

Issue #174, opened 2020-09-11
[GENDATA_RFT]: Add optional directory for outputted files.


berland:
User request to be able to have all the files written to a specified directory, and not directly in RUNPATH.

Number of files written by `gendata_rft` is 1 and up to theoretically hundreds, which clutter up.

This could be implemented as an option with default `.`. 

The job should probably the assume the directory is there  already.

(assuming that `GENDATA` is happy loading files from relative paths)

@rnyb, @erlid

markusdregi:
This is the function that write the files: https://github.com/equinor/semeio/blob/1dcc19e515b94ac0c9225c100d7e46f86beb248d/semeio/jobs/rft/gendata_rft.py#L7

I support the alternative of defaulting to `.` 👍

PR #176, opened 2020-09-15
Add output directory option to gendata_rft

berland:
Solves #174


ertomatic:

semeio/jobs/scripts/gendata_rft.py
```diff
@@ -100,6 +100,19 @@ def _build_parser():
...
     )
+    parser.add_argument(
+        "-d",
+        "--directory",
+        required=False,
+        type=str,
+        default=".",
+        help=(
+            "Output directory used for outputting resulting files "
+            "(except for the csvfile output). "
```
markusdregi, line 111: Is there a reason for not also putting the csv file in the output directory?

berland: Not a very good reason, but the proposed feature is there to not clutter up the RUNPATH by tens of files, for which the CSV export is not to blame.

The CSV output file has its own dedicated argument, and there the user can specify an output directory already.

If output directory is set, and CSV output filename is an absolute path, how to proceed would be questionable.

markusdregi: I see your point! :) In general I think it would be cleaner if a script dumps all of its output files into the output directory when applicable. I leave the decision to you ;)

Also, maybe `--output-directory` and `-o` is more intuitive command line options?

berland: Yes, I had `-o` first, will change back to that. Little bit long argument in ERT config, and  I don't want to shorten  to  OUTPUT  (to mitigate confusion with the CSV output)

berland:
Updated `--directory` -> `--outputdirectory`

markusdregi (approved):
Looks good 👍

PR #177, opened 2020-09-15
Add auto cluster

oyvindeide:


oyvindeide:
Rebased on #178, so a few of the commits will go away.

tests/jobs/misfit_preprocessor/test_misfit_preprocessor.py
```diff
@@ -160,18 +165,19 @@ def test_misfit_preprocessor_n_polynomials(num_polynomials):
     )
     measured_data = MockedMeasuredData(observations, simulated)
 
-    config = {}
+    config = {"clustering": {"method": method}, "scaling": {"threshold": 0.99}}
```
lars-petter-hauge, line 168: Should we keep the spearman_correlation at the same threshold as default? And rather include it for the auto_scale only?

tests/jobs/misfit_preprocessor/test_misfit_preprocessor.py
```diff
@@ -223,7 +230,10 @@ def test_misfit_preprocessor_configuration_errors():
 
     config = {
         "unknown_key": [],
-        "clustering": {"spearman_correlation": {"fcluster": {"threshold": 1.0}}},
+        "clustering": {
+            "method": "spearman_correlation",
```
lars-petter-hauge, line 234: So we are changing the schema to require a `method` option. Could we then rename the `spearman_correlation` below to be `clustering_options`? Or does it perhaps make sense to keep it as is..

Looking at this from the user perspective I think it would make sense to use `clustering_options` instead. Not sure about the complexity that would bring for the validation part though, as it is likely easier to create individual schema's with their own validation, rather than doing a "if method x, validate content based on x requirements"

tests/jobs/misfit_preprocessor/test_misfit_preprocessor.py
oyvindeide: I agree with the second comment, as to the first point, it does not really impact that test at all, because that job is only looking at the output of the clustering job, and the scaling has no impact on that. 

If you think it makes it more readable though, I can change it, but the reason I did not is to make it more precise what we are testing. But could be that I failed at that.

tests/jobs/misfit_preprocessor/test_misfit_preprocessor.py
oyvindeide: We are not adding `method`to the `schema`, we are just changing the default option of `method` from  `spearman_correlation` to `auto_scale`. I do agree that it would be good to change the names of the schema, but was thinking we could have a discussion with the users regarding the names and the schema before doing that.

tests/jobs/misfit_preprocessor/test_misfit_preprocessor.py
lars-petter-hauge: I would say it makes it less readable, but we'll also not be changing as much. But get your first point and I suppose it's not an issue.

tests/jobs/misfit_preprocessor/test_misfit_preprocessor.py
lars-petter-hauge: Alright, if this does not introduce any changes to the schema I agree that we leave it as is and do the discussion later. That was what you mentioned on the meeting as well I suppose.

Agreed

lars-petter-hauge (approved):
Feel free to merge when tests pass! GJ

PR #178, opened 2020-09-15
Refactor pca

oyvindeide:


semeio/workflows/correlated_observations_scaling/job_config.py
```diff
@@ -171,48 +180,119 @@ def _CALCULATE_KEYS_key_not_empty_list(content):
...
-                        MK.Type: types.Number,
-                        MK.Description: "Scalar controlling the allowed distance"
-                        "between ensemble mean and observation. In particular, "
-                        "if: `abs(observed_value - ensemble_mean) > "
-                        "alpha * (ensenmble_std + observed_std)` the data point "
-                        "will be dropped. The value is defaulted to the value used "
-                        "in the ert run, but if a value is given in the config, "
-                        "that value will be used",
-                    },
+_CORRELATED_OBSERVATIONS_SCHEMA = {
```
oyvindeide, line 183: No changes here, this is just formatting (and changed from function to variable.

lars-petter-hauge:
Could you please provide reasoning behind the refactoring, adding it to the git commit message.

oyvindeide:
> Could you please provide reasoning behind the refactoring, adding it to the git commit message.

Yes, will do that.

semeio/workflows/correlated_observations_scaling/cos.py
```diff
@@ -147,13 +150,36 @@ def run(self, job_config):
             facade.get_alpha(), facade.get_std_cutoff()
         )
         for config in user_config:
-            job = ScalingJob(
-                obs_keys, obs, obs_with_data, config, self.reporter, default_values
+            config = ObsCorrConfig(config, obs_keys, default_values)
```
lars-petter-hauge, line 153: I would avoid overwriting the for loop variable.

semeio/workflows/correlated_observations_scaling/job_config.py
```diff
@@ -171,48 +180,119 @@ def _CALCULATE_KEYS_key_not_empty_list(content):
...
+        "UPDATE_KEYS": {
+            MK.Type: types.NamedDict,
+            MK.Description: _UPDATE_KEYS_DESCRIPTION,
+            MK.Content: {"keys": _KEYS_SCHEMA},
+        },
+    },
+}
+
+
+class ObsCorrConfig:
```
lars-petter-hauge, line 229: To be honest, the need to create a class for this instead of using the `ConfigSuite` object directly, means something is off with the original class.. We're not doing anything more than adding the validation (which could've been together with where the `SCHEMA` is located), and kind of hiding the initialisation of the configsuite object. Finally we provide just a handful of the configurations through additional functions.. 

We're at creating "yet another configuration class".. But this PR does improve the code flow, as all of this logic was just hidden within the job itself. Now we at least have the class completely without any business logic (apart from validation).

tests/jobs/correlated_observations_scaling/unit/test_scaling_config.py
```diff
@@ -74,17 +70,16 @@ def test_config_value_overwrites_default(input_key):
         "CALCULATE_KEYS": {"std_cutoff": 1.0e-6, "alpha": 3},
         "UPDATE_KEYS": {},
     }
-    reporter = FileReporter(os.path.realpath(os.getcwd()))
-    job = ScalingJob(["a_key"], [], ["a_key"], user_config_dict, reporter, default_vals)
+    job = ObsCorrConfig(user_config_dict, [], default_vals)
```
lars-petter-hauge, line 73: If we want to be picky, it's no longer a `job` instance :P

tests/jobs/correlated_observations_scaling/unit/test_scaling_config.py
```diff
@@ -57,12 +55,10 @@ def test_valid_job(
...
         "UPDATE_KEYS": {},
     }
-    reporter = FileReporter(os.path.realpath(os.getcwd()))
-    job = ScalingJob(
-        obs_keys, [], obs_with_data, user_config_dict, reporter, default_vals
-    )
-    assert job.get_index_lists() == scaling_job_content["get_index_lists"]
-    assert job.get_calc_keys() == scaling_job_content["get_calc_keys"]
+
+    config = ObsCorrConfig(user_config_dict, obs_keys, default_vals)
```
lars-petter-hauge, line 59: `test_valid_configuration` I guess. This just makes more sense as well, not having to initiate a `FileReporter` and the entire `ScalingJob` when we're only asserting parts of the configuration

semeio/workflows/correlated_observations_scaling/job_config.py
lars-petter-hauge: But was there a specific reason for doing this change? From function to variable?

lars-petter-hauge:
Had a couple comments but nothing much

Cheers for improving the structure!

Suppose you want to rebase squash and all, and I can approve :)

semeio/workflows/correlated_observations_scaling/job_config.py
oyvindeide: So would you add it as some context to the `ConfigSuite` object? I guess that might be possible, but not very readable. Would you prefer I do that?

tests/jobs/correlated_observations_scaling/unit/test_scaling_config.py
oyvindeide: Picky is good, confusing language is bad ;)

semeio/workflows/correlated_observations_scaling/job_config.py
oyvindeide: As to getting values straight from the config object, that would be possible, the main problem is that the job might have a slight off schema, and then I am thinking of `std_cutoff` and `alfa`, that should perhaps be moved up a level, instead of being under `CALCULATE_KEYS`, but that is more of a breaking change. That would make passing configuration more natural (`config.CALULATE_KEYS` to one job and `config.FILTERING` to the other for example)

semeio/workflows/correlated_observations_scaling/job_config.py
oyvindeide: I saw that the author of `ConfigSuite` chose that style, so I conformed.

semeio/workflows/correlated_observations_scaling/job_config.py
lars-petter-hauge: I don't think we should add anything to the `ConfigSuite` object, I think we should keep it as is, I was just pointing out that we're kind of treating the symptoms here. No need to introduce breaking change because we're not happy with the `class` design :)

semeio/workflows/correlated_observations_scaling/job_config.py
lars-petter-hauge: Alright, I don't have any objections really. As long as we don't get a mix of one over the other between jobs now? I think there are more jobs in semeio that use `ConfigSuite`, and I think we should honour our own structure as well.

oyvindeide: I changed it to be internally consistent, because `misfit_preprocessor` is using dict.

oyvindeide:
Rebased

lars-petter-hauge (approved):

PR #181, opened 2020-09-17
Set fixed tag for libres data

ManInFez:
When libres data changes, semeio might break. Fixed to specific tag. should be updated on demand by semeio

### DIFF
--- a/ci/jenkins/testkomodo.sh
+++ b/ci/jenkins/testkomodo.sh
@@ -2,7 +2,8 @@
 copy_test_files () {
     cp -r $CI_SOURCE_ROOT/tests $CI_TEST_ROOT/tests
     TMP_CLONE_DIR=$(mktemp -d)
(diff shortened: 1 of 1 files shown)


lars-petter-hauge:
Could I ask why this is a problem now? Have there been an update to the libres test data? (glanced over and couldn't see it)

With that said, all of the "libres_test_data" stuff is really annoying and should be removed..

ManInFez:
Yes, in libres master, the test data has renamed the old data containing custom_kw

oyvindeide:
> With that said, all of the "libres_test_data" stuff is really annoying and should be removed..

I second that opinion, but think we must keep it around until there is a good way to `POST` something to the `ert` data api (it is needed for updating observations). 

Also, we are also depending on other things in `libres`, namely `ErtScript`, so this has an impact there as well.

jondequinor (approved):

PR #192, opened 2020-09-22
Do garbage collection after tests to avoid storage pollution in root

### LINKED ISSUE #72: Tests leaks storage folder in root
After running the tests in a clean semeio tree I get a `storage` folder in the root folder. I suspect one of the tests using snake oil to be the sinner as `storage` contains a single folder named `snake_oil`.

oyvindeide:
Closes #72

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import os
 import gc
 import pytest
 from tests import legacy_test_data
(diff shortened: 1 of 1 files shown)


markusdregi:
So basically this PR moves cwd into a separate folder for the run of the test right? Any reason that folder needs to be shared?

oyvindeide:
> So basically this PR moves cwd into a separate folder for the run of the test right? Any reason that folder needs to be shared?

It has scope `session`, so it happens once before running the tests, and does not return before the tests have completed, which is the reason for the shared folder. All the tests creating `storage` are already running in their own `tmpdir`, the problem here lies in the tear down of the tests. 

As far as I could see, there were two cases occurring, if running all the tests just adding `gc.collect` was enough from preventing getting `storage` in the wrong place. If running a subset then it was necessary to move `cwd` before the tests and restore after they were completed.

markusdregi:
> It has scope session

I missed that part 🤷

markusdregi (approved):
It looks good and it also appears to succeed at its promise on my machine 👏

PR #196, opened 2020-09-24
Move csv_export to workflows

oyvindeide:


markusdregi:
The 2 suffix was intentional ;) I'll get back with more in some minutes...

oyvindeide:
> The 2 suffix was intentional ;) I'll get back with more in some minutes...

I know, and the job is still registered (and run) as `CSV_EXPORT2`

markusdregi:
The motivation behind the `2`-suffix was to make it clear that it was an iteration of `CSV_EXPORT1` not to be confused with `CSV_EXPORT`. Another point was that we would like to implement a csv export where were we don't necessarily are trying to replicate `CSV_EXPORT1` behaviour, but do an evaluation of what the users actually want and back that implementation by the data API. The idea was that we would then call that export `CSV_EXPORT` and start facing out `CSV_EXPORT2`.

The discovery of `csv_export2.py` not being executable came from a user that wanted to run the script directly in the terminal. With this in mind I think we should consider keeping the naming of the jobs and scripts consistent, so that we, future us and the users are able to navigate between old `CSV_EXPORT`, `CSV_EXPORT1`, `CSV_EXPORT2`, and future `CSV_EXPORT` 😂

oyvindeide:
I just don´t like having the 2 suffix when there is no 1 in the repo 😂 But I conceed (though I would like to change the internal names and just expose the end-points as `2`).

markusdregi (approved):
Saving `CSV_EXPORT` for later. A nice clean-up though :)

Issue #197, opened 2020-09-24
[DESIGN2PARAMS] Silently ignores existing data in parameters.txt


berland:
design2params will ignore values in the design matrix if it already exists in parameters.txt. This is the correct behaviour, but a warning should be emitted for the keys that are ignored.

Namespaces should be respected, so if `SOMENAMESPACE:FOO` exists in `parameters.txt`, there should not be emitted a warning if the design matrix contains `FOO` (ambiguities from that will be handled by  `design_kw`).

PR #199, opened 2020-09-24
Warn if keys exist in parameters.txt

berland:


berland:
Solves #197

oysteoh (approved):
Seems to be working as expected - and has tests - so good job! 👍 👍

PR #201, opened 2020-09-25
Add configuration error output to ots job

oyvindeide:


semeio/jobs/scripts/overburden_timeshift.py
```diff
@@ -38,14 +39,13 @@ def _get_args_parser():
     return parser
 
 
-def main_entry_point(args=None):
+def main_entry_point():
```
lars-petter-hauge, line 42: If we want to test the `main_entry_point` without modifying `sys.args` at runtime (and without using `subprocess`), we want to allow the `main_entry_point` to accept arguments I believe.

semeio/jobs/scripts/overburden_timeshift.py
```diff
@@ -38,14 +39,13 @@ def _get_args_parser():
...
     parser = _get_args_parser()
-    options = parser.parse_args(args)
+    options = parser.parse_args()
     if options.help_params:
         print(generate_rst_doc())
     else:
-        ots_run(options.config)
-
-
-if __name__ == "__main__":
```
lars-petter-hauge, line 50: Removing the `if __name__...` I would agree on!

semeio/jobs/scripts/overburden_timeshift.py
oyvindeide: Dont want to adapt production code to adhere to tests, can always mock `sys.args`?

semeio/jobs/overburden_timeshift/ots.py
```diff
@@ -14,6 +14,7 @@
 from semeio.jobs.overburden_timeshift.ots_vel_surface import OTSVelSurface
 from semeio.jobs.overburden_timeshift.ots_res_surface import OTSResSurface
 from semeio.jobs.overburden_timeshift.ots_config import build_schema
+from semeio._exceptions.exceptions import ConfigurationError
```
xjules, line 17: The other jobs should use it too I reckon

tests/jobs/overburden_timeshift/test_ots_config.py
```diff
@@ -0,0 +1,68 @@
...
+        },
+    }
+    with tmpdir.as_cwd():
+        with open("ots_config.yml", "w") as f:
+            yaml.dump(conf, f, default_flow_style=False)
+        ots_load_params("ots_config.yml")
+
+
+def test_invalid_config(tmpdir, monkeypatch):
+    dates = []
```
xjules, line 41: context_mock can go away and it should trigger the ConfigurationError anyway

semeio/jobs/scripts/overburden_timeshift.py
oyvindeide:
```
import argparse
import sys
def test_argparse(monkeypatch):
    monkeypatch.setattr(sys, "argv", ["path", "-c", "config_file"])
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "-c", "--config"
    )
    options = parser.parse_args()
```

tests/jobs/overburden_timeshift/test_ots_config.py
oyvindeide: It also triggers `util_abort`

semeio/jobs/scripts/overburden_timeshift.py
lars-petter-hauge: Alright, that's a reasonable way of doing it, I think. And of course we're only mocking specifically sysargv, and not anything else

tests/jobs/overburden_timeshift/test_ots_config.py
lars-petter-hauge: Then there might be another issue? Why is util abort triggered? (He asked, well knowing the answer can be awkward ;) )

oyvindeide: There is another issue, related to reading an ecl file which isnt ecl I think. My plan is to create an issue once this is merged (so I can reference the code). The configuration should in general have better test coverage.

semeio/jobs/overburden_timeshift/ots.py
oyvindeide: If you agree that this is the right approach, then yes I agree, we could do this for the rest of the scripts as well.

tests/jobs/overburden_timeshift/test_ots_config.py
lars-petter-hauge: That shouldn't throw a util abort, but a more general error which can be caught (although that error is thrown for a lot of things..). Util abort is rather thrown if the ecl file is corrupt, and ecl can't recover.. but for later I guess

lars-petter-hauge (approved):
Alright, lgtm

semeio/jobs/scripts/overburden_timeshift.py
oyvindeide: I will move the console scripts to another PR

PR #209, opened 2020-09-29
Add observations filtering option to MPP

### LINKED ISSUE #205: User configured observations list not used in MisfitPreprocessor
The MisfitPreprocessor has the option of giving a subset of observations to work on, but that is never passed on to the job. The configuration is ready for this option, but currently all observations is read:

https://github.com/equinor/semeio/blob/66f6ae542c79a2c7ad9b6cadb93a2f39fa66cabe/semeio/workflows/misfit_preprocessor/config.py#L413

This option should be added to the MisfitPreprocessor, that probably requires us to do the filtering earlier, so only the necessary MeasuredData is returned, or the dataframe must be filtered after the fact. There is some functionality in correlated observations scaling for this exact procedure, so that could be moved to a shared location.

oyvindeide:
Closes #205

### DIFF
--- a/semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
+++ b/semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
@@ -24,7 +24,7 @@ class MisfitPreprocessorJob(SemeioScript):  # pylint: disable=too-few-public-met
         config = assemble_config(config_record, observations)
         config = config.snapshot
 
-        measured_record = _load_measured_record(facade, observations)
+        measured_record = _load_measured_record(facade, config.observations)
         scaling_configs = misfit_preprocessor.run(
             **{
                 "config": config,
--- a/semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
+++ b/semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
@@ -6,6 +6,7 @@ from ert_shared.libres_facade import LibresFacade
 from ert_shared.plugins.plugin_manager import hook_implementation
 from semeio.communication import SemeioScript
 
+from semeio.workflows.misfit_preprocessor.config import assemble_config
 from semeio.workflows import misfit_preprocessor
 from semeio.workflows.correlated_observations_scaling.cos import (
     CorrelatedObservationsScalingJob,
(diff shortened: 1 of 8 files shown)


lars-petter-hauge:
Please add a description, or specify an issues that this PR solves.

semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
```diff
@@ -17,11 +18,16 @@
...
+        config = config.snapshot
+
+        measured_record = _load_measured_record(facade, observations)
         scaling_configs = misfit_preprocessor.run(
             **{
-                "misfit_preprocessor_config": config_record,
+                "config": config,
                 "measured_data": measured_record,
                 "reporter": self.reporter,
             }
```
lars-petter-hauge, line 33: Could I ask why we're not providing the arguments as normal? We're here pretty much just creating a dict just to unpack it again with `**`?

`scaling_configs = misfit_preprocessor.run(config=config,..` should do the same thing?

lars-petter-hauge:
Could you add some reasoning behind the refactoring. How does this improve code flow? Does this in any way help (or worse, deviate) the way misfitpreprocessor and spearman_correlation runs? Looks like there could be some more reuse of code for those two, but not sure if that is within the scope of the PR

It's just easier to review if the author has pitched the idea instead of having to "guess" why (or worse, misinterpret the reason)

semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
oyvindeide: They should give the same thing. I did not implement this originally, so can not say why it was done, would you like me to change it?

lars-petter-hauge (approved):
Looks good!

semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
lars-petter-hauge: I was a bit quick approving, so it's already done. Not a critical thing

markusdregi: Sorry for being late to the dance @lars-petter-hauge and @oyvindeide. It was implemented as this due to my writing on records and future data flow in the workflows in other semeio issues at the same time. It was useful for the discussion, but should it leak into production code? Perhaps not!

Feel free to change it to more standard argument passing at any time if you feel like it 👍

Issue #152, opened 2020-08-28
Use open data for testing


berland:
References to ert_statoil should be removed in test code, e.g. for `csv_export2.py`

An open alternative is: https://github.com/equinor/webviz-subsurface-testdata

oyvindeide:
Closed in https://github.com/equinor/semeio/pull/220

PR #220, opened 2020-10-06
Mock a test dataset for csv_export2 using Norne data

berland:
Refactor verification code to use Pandas for CSV comparison.
This includes a marginally stricter requirement on the header - now avoids theoretical false positives but keeping column order arbitrary.


berland:
Solves #152

oysteoh (approved):

PR #223, opened 2020-10-07
Add docs about recommended hook

### LINKED ISSUE #58: Evaluate strategies for getting the scaling job to function with ES MDA
**The initially expected problem of mixing scaling factors seems to not be an issue**
~~Today both the scaling job and the ES MDA algorithm are writing to the same scaling constant in storage. In particular, the algorithm overwrites the scaling from the scaling job. This should be resolved. Investigate if a "quick fix" where we either change the operator or introduce a second scaling constant in storage is possible. Or possibly something completely different?~~

**Run scaling job only once**
The ACO workflow should only be executed after the first iteration to ensure algorithmic stability. This can probably be resolved as a hack (for now) by passing the iteration number to the workflow under the hood. We need to check whether the scaling factors survive from one iteration to the next.

oyvindeide:
Closes #58, blocked by: https://github.com/equinor/ert/pull/1044

### DIFF
--- a/semeio/workflows/misfit_preprocessor/config.py
+++ b/semeio/workflows/misfit_preprocessor/config.py
@@ -381,6 +381,12 @@ _SCHEMA = {
         "to be an almost out-of-the-box solution, where the user can tweak some "
         "parameters to get reasonable clusterings. The rest will be configured "
         "according to whatever is at the time considered best practices."
+        "The recommended practice for running this workflow is though the "
+        "ert workflow hooks. The hook needs to run after the simulations, as "
+        "simulated data is needed for the workflow. The relevant hooks are "
+        "POST_SIMULATION, PRE_FIRST_ANAYLSIS and PRE_ANALYSIS. The recommended "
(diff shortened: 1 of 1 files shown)


sondreso (approved):

semeio/workflows/misfit_preprocessor/config.py
```diff
@@ -381,6 +381,12 @@ def _realize_filters(observation_keys, context):
         "to be an almost out-of-the-box solution, where the user can tweak some "
         "parameters to get reasonable clusterings. The rest will be configured "
         "according to whatever is at the time considered best practices."
+        "The recommended practice for running this workflow is though the "
+        "ert workflow hooks. The hook needs to run after the simulations, as "
+        "simulated data is needed for the workflow. The relevant hooks are "
+        "POST_SIMULATION, PRE_FIRST_ANAYLSIS and PRE_ANALYSIS. The recommended "
```
berland, line 387: typo: ANAYLSIS

PR #226, opened 2020-10-08
Replace surface

### LINKED ISSUE #213: [OTS] Maps exported to irap ascii, most users use irap binary
There is no option in `ecl` to write files to irap binary, only ascii. To improve this we can change to `xtgeo` which allows for irap ascii and binary.

oyvindeide:
Closes #213

### DIFF
--- a/semeio/jobs/overburden_timeshift/ots.py
+++ b/semeio/jobs/overburden_timeshift/ots.py
@@ -1,13 +1,16 @@
-import xtgeo
 import logging
-from collections import namedtuple
 import os.path
+from collections import namedtuple
+from itertools import product
 from datetime import datetime as dt
 from pathlib import Path
 
+
+import configsuite
+import yaml
+import xtgeo
 import numpy as np
 from scipy.interpolate import CloughTocher2DInterpolator
-
 from ecl.grid import EclGrid
 from ecl.gravimetry import EclSubsidence
 from ecl.eclfile import Ecl3DKW, EclFile
@@ -16,8 +19,6 @@ from semeio.jobs.overburden_timeshift.ots_vel_surface import OTSVelSurface
 from semeio.jobs.overburden_timeshift.ots_res_surface import OTSResSurface
 from semeio.jobs.overburden_timeshift.ots_config import build_schema
 from semeio._exceptions.exceptions import ConfigurationError
-import configsuite
-import yaml
 
 
 # pylint: disable=consider-using-enumerate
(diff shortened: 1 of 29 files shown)


semeio/jobs/overburden_timeshift/ots.py
```diff
@@ -100,18 +102,35 @@ def ots_run(parameter_file):
...
-                for iv in range(len(vintage_pairs.ts_simple)):
-                    ts.append(tshift_ts_simple[iv][point])
-                for iv in range(len(vintage_pairs.dpv)):
-                    ts.append(tshift_dpv[iv][point])
-                for iv in range(len(vintage_pairs.ts_rporv)):
-                    ts.append(tshift_ts_rporv[iv][point])
-                f.write(line.format(xy[0], xy[1], surface_horizon[point], *ts))
+
+            point = 0
+            for x_index in range(1, surface_horizon.get_nx() + 1):
```
lars-petter-hauge, line 107: We tend to do a lot of these double loops, which is a very natural thing as we are working with grids very often (or even triple for loops, when going through 3 dimensions).

I've always disliked that it is necessary to add more indents than a single one.. 

What do you think about:

`for i, j in itertools.product(range(1, nx+1), range(1, ny+1)):`

A bit annoying that it isn't 0-indexed, but alright..

If you need the `point` as well:

`for point, (i, j) in enumerate(itertools.product(range(1, nx+1), range(1, ny+1))):`

oyvindeide: Yes, will probably do something like this, or perhaps suggest a patch upstream? I think it makes sense for a surface to be iterable, and this is at least what I would expect to get if I iterated over it (it is not iterable at the moment).

oyvindeide: My first goal was just to replace it, then clean up the implementation ;)

semeio/jobs/overburden_timeshift/ots.py
```diff
@@ -49,10 +48,13 @@ def ots_load_params(input_file):
...
 
 def write_surface(vintage_pairs, ts, output_dir, type_str):
+    output_dir += type_str
+    if not os.path.isdir(output_dir):
+        os.mkdir(output_dir)
     for iv, vp in enumerate(vintage_pairs):
         d0 = vp[0].strftime("%Y_%m_%d")
         d1 = vp[1].strftime("%Y_%m_%d")
-        ts[iv].write(output_dir + type_str + "/ots" + "_" + d0 + "_" + d1 + ".irap")
+        ts[iv].to_file(os.path.join(output_dir, "ots" + "_" + d0 + "_" + d1 + ".irap"))
```
xjules, line 57: Should the extension be the same for xtgeo surface?

oyvindeide: It is still `irap`, so I suppose so, but this implementation is failing for one of the tests, so might not be feasible to replace.

semeio/jobs/overburden_timeshift/ots.py
```diff
@@ -172,34 +190,44 @@ def _create_surface(self, z=None):
...
+            rotation=0,
         )
 
         irap_x = np.empty(nx * ny)
         irap_y = np.array(irap_x)
 
-        for i in range(len(surf)):
-            irap_x[i], irap_y[i] = surf.getXY(i)
+        counter = 0
+        for x_index in range(1, nx + 1):
```
lars-petter-hauge, line 216: Could technically do the same here as above: `for point, (x_index, y_index) in enumerate(product(`

semeio/jobs/overburden_timeshift/ots.py
```diff
@@ -50,10 +50,13 @@ def ots_load_params(input_file):
 
 
 def write_surface(vintage_pairs, ts, output_dir, type_str):
```
jondequinor, line 52: Pathlib??

Shouldn't we use pathlib for stuff like this?

```suggestion
def write_surface(vintage_pairs, ts, output_dir, type_str):
    output_path = Pathlib(output_dir + type_str)
    if not output_path.is_dir():
        output_path.mkdir()
    …
        ts[iv].to_file((output_dir / f"ots_{d0}_{d1}.irap"))    
```

jondequinor:
Both the description and issue is a bit scarce, so someone with more knowledge will have to review. Except for the one comment, code looks OK.

semeio/jobs/overburden_timeshift/ots.py
jondequinor: 🙇

oyvindeide:
> Both the description and issue is a bit scarce, so someone with more knowledge will have to review. Except for the one comment, code looks OK.

I have added some context in the issue and changed the commit text.

jondequinor (approved):
Code looks good to me.

PR #228, opened 2020-10-08
Fix bug when writing no pressure points

### LINKED ISSUE #227: [GENDATA_RFT] Not writing newline when no pressure in dataframe
A user reported a problem when there were no pressure points the defined zones, and the individual points were on the same line, and not each on a new line as they should.

oyvindeide:
Closes #227


sondreso (approved):
Nice!

PR #230, opened 2020-10-09
Move CI from travis to GitHub Actions

### LINKED ISSUE #208: Move CI to github actions

kvashchuka:
Closes #208


.travis.yml
```diff
@@ -1,65 +0,0 @@
...
-    - export PYTHONPATH=$INSTALL_DIR/lib/python$TRAVIS_PYTHON_VERSION/dist-packages:$PYTHONPATH
-    - export PYTHONPATH=$INSTALL_DIR/lib/python$TRAVIS_PYTHON_VERSION/site-packages:$PYTHONPATH
-    - PATH=$INSTALL_DIR/bin:$PATH
-    - pip install ert/ --prefix=$INSTALL_DIR
-
-script:
-  - pip install tox
-  - tox
-
-deploy:
```
oyvindeide, line 56: I know I said you could delete the travis file, but until I can set up deploy on GitHub actions, we need this part.

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,64 @@
...
+          bash ert/.build_install.sh libecl
+          bash ert/.build_install.sh libres
+          pip install -r ert/dev-requirements.txt
+          pip install ert/
+          pip list
+          ert --help
+
+      - name: "Install dependencies"
+        run: |
+          set -xe
```
oyvindeide, line 57: Is installing anything but `tox` and `tox-gh-actions` needed?

oyvindeide:
Looking good so far! Just missing a run for the `style` and a couple of comments. Suggest you make that as a independent workflow, almost like https://github.com/equinor/libres/blob/master/.github/workflows/black.yml

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,64 @@
+---
+name: CI
+
+on:
+  push:
+    branches: ["master"]
+  pull_request:
+    branches: ["master"]
```
sondreso, line 8: It's nice to have the tests run when you make pull requests to branches other than master also :slightly_smiling_face:

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Setup environment
+        run: |
+          echo "::add-path::${{ env.INSTALL_DIR}}/bin"
+          echo "::set-env name=LD_LIBRARY_PATH::${{ env.INSTALL_DIR }}/lib:${{ env.INSTALL_DIR }}/lib64"
+          echo "::set-env name=DYLD_LIBRARY_PATH::${{ env.INSTALL_DIR }}/lib:${{ env.INSTALL_DIR }}/lib64"
+          echo "::set-env name=PYTHONPATH::${{ env.INSTALL_DIR }}/lib/python${{ matrix.python-version }}/site-packages:${{ env.INSTALL_DIR }}/lib/python${{ matrix.python-version }}/dist-packages"
+      
+      - name: Install dependencies and semeio
+        run: |
+          python -m pip install --upgrade pip
```
sondreso, line 38: Is this needed?

kvashchuka: Hm, I don't know. Don't we need the latest pip?

sondreso: I think the pip than comes with setup-python step is quite new, but you can check using `pip --version` :smile:

oyvindeide: You can also set that `tox` downloads the latest `pip `by setting `download=true` in the `tox` config.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Setup environment
+        run: |
+          echo "::add-path::${{ env.INSTALL_DIR}}/bin"
+          echo "::set-env name=LD_LIBRARY_PATH::${{ env.INSTALL_DIR }}/lib:${{ env.INSTALL_DIR }}/lib64"
+          echo "::set-env name=DYLD_LIBRARY_PATH::${{ env.INSTALL_DIR }}/lib:${{ env.INSTALL_DIR }}/lib64"
+          echo "::set-env name=PYTHONPATH::${{ env.INSTALL_DIR }}/lib/python${{ matrix.python-version }}/site-packages:${{ env.INSTALL_DIR }}/lib/python${{ matrix.python-version }}/dist-packages"
+
+      - name: Install dependencies and semeio
+        run: |
+          pip install semeio
```
oyvindeide, line 32: You dont need this, as it installed in the `tox` environment (also, `semeio` is a plugin to `ert`, and should therefor be installed after `ert`).

.github/workflows/testing.yml
```diff
@@ -0,0 +1,52 @@
...
+          git clone --branch master --depth 1 https://github.com/equinor/ert
+          source ert/.libres_version
+          git clone --branch $LIBRES_VERSION --depth 1 https://github.com/equinor/libres
+          source libres/.libecl_version
+          git clone --branch $LIBECL_VERSION --depth 1 https://github.com/equinor/libecl
+          bash ert/.build_install.sh libecl
+          bash ert/.build_install.sh libres
+          pip install -r ert/dev-requirements.txt
+          pip install ert/
+          pip list
```
oyvindeide, line 45: This is not needed, also dont need to run `ert --help`, as that is to test the command line `ert` (we are not using that)

.github/workflows/style.yml
```diff
@@ -0,0 +1,17 @@
...
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install tox and pre-commit
+        run: pip install tox pre-commit
```
oyvindeide, line 15: Should not have to install `pre-commit`, tox will do that for you.

kvashchuka: Yeah, sorry, it is an artifact from trying to make this workflow to pass

.github/workflows/testing.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Install ERT and dependencies
+        run: |
+          git clone --branch master --depth 1 https://github.com/equinor/ert
+          source ert/.libres_version
+          git clone --branch $LIBRES_VERSION --depth 1 https://github.com/equinor/libres
+          source libres/.libecl_version
+          git clone --branch $LIBECL_VERSION --depth 1 https://github.com/equinor/libecl
+          bash ert/.build_install.sh libecl
+          bash ert/.build_install.sh libres
+          pip install -r ert/dev-requirements.txt
```
oyvindeide, line 39: This is not needed

.github/workflows/testing.yml
```diff
@@ -0,0 +1,46 @@
...
+      - uses: "actions/checkout@v2"
+      - uses: "actions/setup-python@v2"
+        with:
+          python-version: "${{ matrix.python-version }}"
+
+      - name: Setup environment
+        run: |
+          echo "::add-path::${{ env.INSTALL_DIR}}/bin"
+          echo "::set-env name=LD_LIBRARY_PATH::${{ env.INSTALL_DIR }}/lib:${{ env.INSTALL_DIR }}/lib64"
+          echo "::set-env name=DYLD_LIBRARY_PATH::${{ env.INSTALL_DIR }}/lib:${{ env.INSTALL_DIR }}/lib64"
```
oyvindeide, line 27: Technically this is only needed for osx testing, which is not in the matrix.

oyvindeide (commented):
Some really minor comments now, other than those this is looking good 👍 Will you have a look at the comments and rebase?

README.md
```diff
@@ -1,5 +1,5 @@
 [![PyPI version](https://badge.fury.io/py/semeio.svg)](https://badge.fury.io/py/semeio)
-[![Build Status](https://travis-ci.com/equinor/semeio.svg?branch=master)](https://travis-ci.com/equinor/semeio)
+[![Actions Status](https://github.com/equinor/semeio/workflows/CI/badge.svg)](https://github.com/equinor/semeio/actions?query=workflow=CI)
```
kvashchuka, line 2: According to this https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge#using-a-workflow-name, the following line would be enough for the CI badge
Actions Status
Should I change it? And do we want to see the status of passing tests only on master or on every branch?

oyvindeide: So the second part of that badge is the link if you actually press the badge, which I think is nice to have as a shortcut to the CI. So I would say keep that link as well.

The main reason for having badges for other branches tends to be with a slightly different workflow than us. We tend to work on master, while other projects might for example have a feature branch as well as master. Then it makes sense to have a different badge for different branches, in our case, I would say master is enough.

oyvindeide:
More or less done now, except you have not changed `.travis.yml`, and only one of the CI jobs is actually running the tests, the other two (3.6 and 3.8) are empty. You need to include them here: https://github.com/equinor/semeio/blob/d1e53185f9dfd9a0f593b4456acf3704a9da1aa9/tox.ini#L3

oyvindeide (approved):
LGTM! Thanks for a great addition to the CI!

PR #231, opened 2020-10-12
initial add ahmanalysis script

mareco701:
This is to add ahm_analysis job to the ert jobs


ertomatic:

markusdregi:
The reason the CI (automatic tests) are failing is code formatting and dependencies @mareco701... As you can see from the logs (https://travis-ci.com/github/equinor/semeio/jobs/398223563) you can reproduce them locally:

> If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.

It seems like running black on the code (a code formatting tool) and adding the dependencies you have that is not present in setup.py will get you a bit further 👍 You can install black via pip (`pip install black`)  and then `black --help` should get you started 🚀

markusdregi:
Good @mareco701 :tada: Now the code is at least formatted correctly such that the existing test pipeline passes.

The next step now is to write tests for your code. If you are not familiar with writing tests I would encourage you to start with just writing down a list of how you expect your program to behave in different situations. Each of those scenarios should then be transformed into a test that verifies the behavior of the program. This is often easier if one writes both unit and integration tests, where the unit tests verifies a smaller part of your program, while the integration tests runs the full workflow as it is intended that the users will. There are numerous examples of how this is currently done for the other workflows :)

If you write the tests such that I can:
- read the tests and get a good impression of how the code is behaving, both when successful and when failing that is a good test base and
- start editing your code (but not the tests) and as long as the tests are still passing you are quite confident the code is still correct

you have a good basis for a test base :sunglasses:

mareco701:
Both unit and integration tests have been added.

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+    test_data_dir = os.path.join(test_data_root, "snake_oil")
+
+    shutil.copytree(test_data_dir, "test_data")
+    os.chdir(os.path.join("test_data"))
+
+    res_config = ResConfig("snake_oil_field.ert")
+    res_config.convertToCReference(None)
+    ert = EnKFMain(res_config)
+    notifier = ErtCliNotifier(ert, res_config)
+    ERT.adapt(notifier)
```
oyvindeide, line 96: This should not be needed?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+        #        out_file = config_node.get_enkf_outfile()
+        path_fmt = os.path.join(
+            output_path.replace("scalar_", "field_") + state,
+            "%d_" + fieldparam + "_field.grdecl",
+        )
+        iens_list = np.arange(0, iensnb)
+        EnkfNode.exportMany(config_node, path_fmt, state_fs, iens_list)
+
+
+def list_active_observations(key, ert, active_obs, output_path, prior_name):
```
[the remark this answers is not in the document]
oyvindeide: Notice that testing this function will be alot easier if you dont have to deal with `ert`, when what you need from `ert` is just a list and a `dataframe` (or you actually just need the keys from it). If you make that the input instead, testing becomes much easier, and you can make a minimal example.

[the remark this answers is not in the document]
oyvindeide: When looking in more detail, you only need the keys. So if you load the `dataframe` in the calling function and just pass the `keys` in here, it becomes much easier to test it.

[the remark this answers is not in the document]
oyvindeide: Just as an example, you could write a test something like this if you change a few things:
```
def test_list_active_observations():
    active_obs = pd.DataFrame({'WPR_DIFF_1': {'ratio': '4 active/4'}})
    log_file = "some/correct/file/scalar_"
    result = ahmanalysis.list_active_observations('WPR_DIFF_1', ['WPR_DIFF_1'], ['WPR_DIFF_1'], active_obs, log_file)
    assert result == some_result
```

Function changed to:
```suggestion

def list_active_observations(key, misfit_keys, gen_obs_list, active_obs, output_path):
    """To get the inactive vs total observation info."""
    update_log_path = output_path.replace("scalar_", "update_log/")
    misfit_name = "MISFIT:" + key
    if key in gen_obs_list and misfit_name not in misfit_keys:
            active_obs.at["ratio", key] = "0 Active"
    else:
        df_update_log = make_update_log_df(update_log_path, key)
        df_active = df_update_log[(df_update_log["status"] == "Active")]
        active_obs.at["ratio", key] = (
            str(len(df_active)) + " active/" + str(len(df_update_log.index))
```

oyvindeide (commented):
Hi, thanks for the PR! I know this is challenging to work with, so thanks for doing it! There are quite a few comments here, but a lot of them are repetition.

There are some general comments and questions:
* First of, can you rebase this into logical commits? That proably means one commit. Keep in mind that the tests should pass on each commit, so one commit should be one isolated change (for example adding the command line entry point could be one commit)
* Try to limit functions using `ert` where possible. As a general rule a function should know as little about its surroundings as possible. I have added a few examples of where that could be done.
* I would like to see more unit tests of functions, that would really help me when reading and understanding the code. Writing unit tests will also be much easier when the `ert` is less used.
* Would be good if the tests would do more specific asserts, which also helps when understanding what a function does, if you are finding that you need to do a lot of asserts in a single (unit) test, that could indicate that the function has to much responsibility.

If any comments are unclear feel free to reach out.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+requires an ert config file as input
+optionally a name can be set for the target case used for
+update (ahm) calculation
+"""
+
+
+class AhmAnalysisJob(ErtScript):  # pylint: disable=too-few-public-methods
+    """ Define ERT workflow to evaluate change of parameters for each observation"""
+
+    def run(self, ertconfig=None, target_name="analysis_case", prior_name=None):
```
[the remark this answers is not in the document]
mareco701: this is fixed now (removed)

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+                )
+        # save/export the Ks matrix
+        ks_data.to_csv(output_path + "ks.csv")
+        active_obs.to_csv(output_path + "active_obs_info.csv")
+        misfitval.to_csv(output_path + "misfit_obs_info.csv")
+        # print on the terminal size of the KS matrix
+        print("size of KS matrix:", ks_data.shape)
+
+
+def mk_localization_scheme(ert, var_ert, obs_list):
```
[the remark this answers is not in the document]
mareco701: this is not a function anymore since I did not really see how to test (using some ert functionality)

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    prior_fs = ert.getEnkfFsManager().getFileSystem(prior_name)
+    target_fs = ert.getEnkfFsManager().getFileSystem(target_name)
+    update_log_path = output_path.replace("scalar_", "update_log/")
+    ert.analysisConfig().set_log_path(update_log_path)
+    es_update = ESUpdate(ert)
+    run_context = ErtRunContext.ensemble_smoother_update(prior_fs, target_fs)
+    es_update.smootherUpdate(run_context)
+
+
+def make_update_log_df(update_log_file, key):
```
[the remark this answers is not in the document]
mareco701: there is a test now

semeio/workflows/ahm_analysis/ahmanalysis.py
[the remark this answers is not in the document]
mareco701: there is a test now

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    else:
+        df_update_log = make_update_log_df(update_log_path, key)
+        df_active = df_update_log[(df_update_log["status"] == "Active")]
+        active_obs.at["ratio", key] = (
+            str(len(df_active)) + " active/" + str(len(df_update_log.index))
+        )
+    return active_obs
+
+
+def create_path(var_ert, target_name):
```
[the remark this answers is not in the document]
mareco701: there is a test now

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    newpath = runpathdf.rsplit("/", num)[0]
+    # create ouput folder for saving calculated data
+    output_path = newpath + "/share/output_analysis/scalar_" + target_name + "/"
+    if not os.path.isdir(output_path):
+        os.makedirs(output_path)
+        print("Directory ", output_path, " Created ")
+    return output_path
+
+
+def list_observations_misfit(key, ert, misfitval, output_path, prior_name):
```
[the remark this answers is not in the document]
mareco701: there is a test now.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+        if "All_obs" in key:
+            misfitval.at["misfit", key] = "None"
+        else:
+            for keys_obs in df_active["obs_key"]:
+                list_kobs.append("MISFIT:" + keys_obs)
+            misfitval.at["misfit", key] = misfit_df[list_kobs].mean(axis=1).mean()
+    return misfitval
+
+
+def get_input_state_df(ert, df_path, state, fieldparam):
```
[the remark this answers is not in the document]
mareco701: there is a test now

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+        print("A grid with .EGRID format is expected.")
+    else:
+        dataframeg = mygrid.dataframe(activeonly=False)
+        # reorder the XYZ to have them identical order as the field parameter
+        mygrid_ok = dataframeg.sort_values(by=["KZ", "JY"]).reset_index()
+
+    return mygrid_ok
+
+
+def calc_delta_grid(
```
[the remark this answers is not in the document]
mareco701: there is a test now

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+            result = np.all(prior_data[dkey] == flatten_arr[0])
+            if not result:
+                p_keysf.append(dkey)
+        else:
+            if not all(x == prior_data[dkey][0] for x in prior_data[dkey]):
+                p_keysf.append(dkey)
+    return p_keysf, field_parameters
+
+
+def calc_ks(ks_matrix, dkeys, prior_data, target_data, key):
```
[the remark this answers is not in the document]
mareco701: there is a test now

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    """Calculate KS matrix"""
+    for dkey in sorted(dkeys):
+        ks_val = ks_2samp(prior_data[dkey], target_data[dkey])
+        ks_matrix.at[dkey, key] = ks_val[0]
+    return ks_matrix
+
+
+def check_ertconfig(ert, ertconfig):
```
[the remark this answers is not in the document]
mareco701: function not needed anymore (removed)

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    """ to check if there is an ert config file to setup ERT"""
+    if ertconfig is not None:
+        notifier = ErtCliNotifier(ert, ertconfig)
+        ERT.adapt(notifier)
+
+
+def check_names(ert, prior_name, target_name):
```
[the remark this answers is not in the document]
mareco701: there is a test now

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    """Check input namees given"""
+    if prior_name is None:
+        prior_name = ERT.enkf_facade.get_current_case_name()
+    if target_name == "<ANALYSIS_CASE_NAME>":
+        target_name = "analysis_case"
+    return prior_name, target_name
+
+
+def get_dataframe_pca(pca, all_input):
```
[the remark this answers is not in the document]
mareco701: this is not a function anymore (part of the main function)

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    """generate the dataframe of field parameter containing PC"""
+    pc_field = pca.transform(all_input)
+    pc_field_df = pd.DataFrame(data=pc_field)
+    return pc_field_df
+
+
+def perform_pca(all_input_prior):
```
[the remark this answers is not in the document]
mareco701: this is not a function anymore (part of the main function)

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+        else:
+            for keys_obs in df_active["obs_key"]:
+                list_kobs.append("MISFIT:" + keys_obs)
+            misfitval.at["misfit", key] = misfit_df[list_kobs].mean(axis=1).mean()
+    return misfitval
+
+
+def get_input_state_df(ert, df_path, state, fieldparam):
+    """create input prior and posterior dataframe"""
+    iensnb = ert.getEnkfFsManager().getEnsembleSize()
```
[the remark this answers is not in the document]
mareco701: this is fixed no more function using ert inside the function.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    """ to check if there is an ert config file to setup ERT"""
+    if ertconfig is not None:
+        notifier = ErtCliNotifier(ert, ertconfig)
+        ERT.adapt(notifier)
+
+
+def check_names(ert, prior_name, target_name):
+    """Check input namees given"""
+    if prior_name is None:
+        prior_name = ERT.enkf_facade.get_current_case_name()
```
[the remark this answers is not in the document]
mareco701: this is fixed no more use of ERT in the script

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    ErtRunContext,
+    ResConfig,
+)
+from res.enkf.export import (
+    GenKwCollector,
+    SummaryObservationCollector,
+    GenDataObservationCollector,
+    MisfitCollector,
+)
+from ert_shared import ERT
```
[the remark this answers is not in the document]
mareco701: no, it has been removed

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+        print("size of KS matrix:", ks_data.shape)
+
+
+def mk_localization_scheme(ert, var_ert, obs_list):
+    """ Use localization to evaluate change of parameters for each observation"""
+
+    field_parameters = sorted(
+        var_ert.ensembleConfig().getKeylistFromImplType(ErtImplType.FIELD)
+    )
+    scal_parameters = sorted(
```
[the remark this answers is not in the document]
mareco701: I meant scal for scalar, I can rename

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+        if obskey_val == replace:
+            updatelog.at[i, "obs_key"] = key
+
+    return updatelog
+
+
+def make_obs_groups(ert):
+    """create dataframe with grouped observations
+    as appearing in Observations on Ert gui"""
+    gen_obs_list = GenDataObservationCollector.getAllObservationKeys(ert)
```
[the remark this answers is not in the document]
mareco701: this is fix no more use of ert inside subfunction

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    try:
+        mygrid = Grid(input_grid_name, fformat="eclipserun")
+    except OSError:
+        print("A grid with .EGRID format is expected.")
+    else:
+        dataframeg = mygrid.dataframe(activeonly=False)
+        # reorder the XYZ to have them identical order as the field parameter
+        mygrid_ok = dataframeg.sort_values(by=["KZ", "JY"]).reset_index()
+
+    return mygrid_ok
```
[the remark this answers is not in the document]
mareco701: done/updated base on comments

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: correct, it is removed

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    field_parameters = sorted(
+        var_ert.ensembleConfig().getKeylistFromImplType(ErtImplType.FIELD)
+    )
+    scal_parameters = sorted(
+        var_ert.ensembleConfig().getKeylistFromImplType(ErtImplType.GEN_KW)
+    )
+    all_parameters = field_parameters + scal_parameters
+    transfp_keys = []
+    p_keys = []
```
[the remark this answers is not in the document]
mareco701: I think it is fixed (variable is used).

semeio/workflows/ahm_analysis/ahmanalysis.py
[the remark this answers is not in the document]
mareco701: rename done

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    parser.add_argument(
+        "--case_prior",
+        type=str,
+        help="name of ert case used as prior (default is ert current_case)",
+        default=None,
+    )
+    return parser
+
+
+def main(args=None):
```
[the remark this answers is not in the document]
mareco701: it is removed now

berland:
Recommend to run-length-encode the grdecl test files in this PR with `eclcompress`.

tests/legacy_test_data/snake_oil/fields/permx0.grdecl
```diff
@@ -0,0 +1,152 @@
+PERMX
```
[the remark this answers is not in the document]
mareco701: they are now written on demand with test

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+
+def make_obs_groups(ert):
+    """create dataframe with grouped observations
+    as appearing in Observations on Ert gui"""
+    gen_obs_list = GenDataObservationCollector.getAllObservationKeys(ert)
+    summary_obs_list = SummaryObservationCollector.getAllObservationKeys(ert)
+    obs_keys = gen_obs_list + summary_obs_list
+    local_obs = ert.getObservations()
+    obs_groups = {}
+    for keys in obs_keys:
```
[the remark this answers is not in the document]
mareco701: updated

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+        nkeys = re.sub(":", "_", keys)
+        obs_groups[nkeys] = []
+    obs_groups["All_obs"] = []
```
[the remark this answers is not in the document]
mareco701: test added to fail and give message if ''All_obs" is an existing observation vector

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+
+    res_config = ResConfig("snake_oil.ert")
+    res_config.convertToCReference(None)
+    ert = EnKFMain(res_config)
+
+    gen_obs_list = GenDataObservationCollector.getAllObservationKeys(ert)
+    summary_obs_list = SummaryObservationCollector.getAllObservationKeys(ert)
+    obs_keys = gen_obs_list + summary_obs_list
+    dfobs = ahmanalysis.make_obs_groups(ert)
+    assert len(dfobs) == (len(obs_keys) * 2) + 1
```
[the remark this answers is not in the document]
mareco701: more meaningful assert added

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+
+    shutil.copytree(test_data_dir, "test_data")
+    os.chdir(os.path.join("test_data"))
+
+    res_config = ResConfig("snake_oil.ert")
+    res_config.convertToCReference(None)
+    ert = EnKFMain(res_config)
+
+    gen_obs_list = GenDataObservationCollector.getAllObservationKeys(ert)
+    summary_obs_list = SummaryObservationCollector.getAllObservationKeys(ert)
```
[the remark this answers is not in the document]
mareco701: obs_keys list is now explicitly written

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,567 @@
...
+    """generate the dataframe of field parameter containing PC"""
+    pc_field = pca.transform(all_input)
+    pc_field_df = pd.DataFrame(data=pc_field)
+    return pc_field_df
+
+
+def perform_pca(all_input_prior):
+    """perform pca on prior dataset"""
+    df_input_prior = pd.DataFrame(all_input_prior)
+    pca = PCA(0.98)
```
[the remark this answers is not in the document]
mareco701: I am not sure about this... could it be this is the same as the one used by ERT ???

mareco701:
> Recommend to run-length-encode the grdecl test files in this PR with `eclcompress`.

I have followed the advice from oyvind and the data is generated within the test temporary folder.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -262,7 +275,7 @@ def make_update_log_df(update_log_file, key):
         os.path.join(update_log_file, f) for f in os.listdir(update_log_file)
     ]
     if len(list_of_files) > 1:
-        print("Warning more than one update_log_file.")
+        raise OSError("Warning more than one update_log_file.")
```
oyvindeide, line 276: Any reason this should not be an error?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -423,20 +394,23 @@ def calc_delta_grid(
...
-            if "_" + paramet + ":" in key:
-                transfp_keys.append(key)
-    p_keys = set(prior_data.keys()).difference(transfp_keys)
+    parameter_keys = _get_parameter_keys(parameters, prior_data.keys())
     # remove parameters with constant prior distribution
-    for dkey in p_keys:
+    p_keysf = []
+    for dkey in parameter_keys:
         if prior_data[dkey].ndim > 1:
             print("WARNING: Parameter " + dkey + " defined several times.")
```
oyvindeide, line 401: I am a bit confused about this, what is this meant to do? It is never touched by a test, so a bit hard to determine.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -423,20 +394,23 @@ def calc_delta_grid(
...
     """make list of updated parameters
     (excluding duplicate transformed parameters)
     """
-    transfp_keys = []
-    p_keys = []
-    p_keysf = []
-    for paramet in parameters:
-        for key in prior_data.keys():
-            if "_" + paramet + ":" in key:
-                transfp_keys.append(key)
```
oyvindeide: What does `transfp` mean?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,507 @@
...
+    return ks_data, active_obs, misfitval
+
+
+def make_update_log_df(update_log_file, key):
+    """ Read update_log file to get active and inactive observations """
+    list_of_files = [
+        os.path.join(update_log_file, f) for f in os.listdir(update_log_file)
+    ]
+    if len(list_of_files) > 1:
+        print("Warning more than one update_log_file.")
```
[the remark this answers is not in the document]
mareco701: actually in the context of the script there are always only one update log file generated (it is overwritten for each ministep update) and also since it is generated in another folder than the one setup in the ert config it is independent (no risk of confusion/several files). But we can keep the check and I agree return an error if it happens several files are present in the update_log folder (it should not happen).

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: I agree it should be an error.

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: this is checking if some parameters are defined several times (mistakenly introduced by the user). If there are constant parameters defined several times ert will not crash and most likely do not give a message (so in this case it is just to give a warning (make the user aware of it) since it will not affect our script results) but if they are not constant ert will crash (so this is not an issue ).

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: this is to collect all the parameters which are transformed (ert can generate a new parameter named for instance LOG10_GENKWID:PARAM in addition to GENKWID:PARAM ) and therefore give the option to plot the transformed or not transformed distribution later in webviz.

semeio/workflows/ahm_analysis/ahmanalysis.py
oyvindeide: Thanks for the clarification, could you add a test that touches this part of the code? Like you said, it does not affect our script, but if it happens because of a user error we should perhaps throw an error? It should of course be the responsibility of `ert` to validate the input, and we should not have to do that at this point ideally. So perhaps the correct course of action is to create an issue in `ert`?

semeio/workflows/ahm_analysis/ahmanalysis.py
oyvindeide: Thanks for the clarification, would it be possible to add slightly more descriptive variable names here?

mareco701: changed to something more explicit.

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: I have added a test.

lars-petter-hauge:
I was asked (by reviewer) to have a go at reviewing from more of an outside "haven't looked at the code yet" kind of type - in case you're wondering why more people are reviewing @mareco701 :)

tests/jobs/ahm_analysis/test_integration.py
```diff
@@ -0,0 +1,102 @@
...
+    """check dataset without prior data"""
+    test_data_dir = os.path.join(test_data_root, "snake_oil")
+    shutil.copytree(test_data_dir, "test_data")
+    os.chdir(os.path.join("test_data"))
+    res_config = ResConfig("snake_oil.ert")
+    res_config.convertToCReference(None)
+    ert = EnKFMain(res_config)
+
+    # check that it fails
+    with pytest.raises(Exception):
```
oyvindeide, line 101: This needs to be changed to catch a specific error, because now it just blanket allows anything, this will for example not make the test fail (though it fails for a different reason than before):

```
    # check that it fails
    with pytest.raises(Exception):
        ahmanalysis.AhmAnalysisJob(ert, fdsvmfdkjsa, fdsafdsa, fdsa,fd,safdsa).run(prior_name="default")
```

even this passes, which is actually a `SyntaxError`:
```
    with pytest.raises(Exception):
        eval('x === x')
        ahmanalysis.AhmAnalysisJob(ert).run(prior_name="default")
```

tests/jobs/ahm_analysis/test_integration.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    assert os.path.isfile(output_deltafield)
+    delta_df = pd.read_csv(output_deltafield, index_col=0)
+    assert len(delta_df.columns) == 8 + (len(obs_keys) * 2) + 1
+    # check field parameter is present and not empty in the final KS matrix
+    output_ks = os.path.join(
+        "storage", "snake_oil_field", "reports", "default", "AhmAnalysisJob", "ks.csv"
+    )
+    ks_df = pd.read_csv(output_ks, index_col=0)
+    assert not ks_df.empty
```
oyvindeide, line 84: Would like to have some more concrete checks than this, to make sure that the script does the right thing

tests/jobs/ahm_analysis/test_integration.py
```diff
@@ -0,0 +1,102 @@
...
+from res.enkf import EnKFMain, ResConfig
+from res.enkf.export import (
+    SummaryObservationCollector,
+    GenDataObservationCollector,
+)
+from ecl.grid import EclGridGenerator
+
+
+@pytest.mark.usefixtures("setup_tmpdir")
+def test_ahmanalysis_run(test_data_root):
```
lars-petter-hauge, line 17: Unless I'm mistaken, the only tests for the `run` method - which does include quite a bit of logic, are these integration tests. I would suggest that we assert a bit more. For instance, we could verify that an output has been created for each of the observation groups. I'm a bit split on how much we should assert of the created values, as we are here testing the job, and not the implementation in libres. But it wouldn't be outrageous to check against an expected output set.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+        # Get the prior scalar parameter distributions
+        prior_data = check_inputs(
+            MisfitCollector.loadAllMisfitData(ert, prior_name),
+            GenKwCollector.loadAllGenKwData(ert, prior_name),
+        )
+
+        # create dataframe with observations vectors (1 by 1 obs and also all_obs)
+        combinations = make_obs_groups(key_map)
+        # setup dataframe for calculated data
+        ks_data, active_obs, misfitval = initialize_emptydf()
```
lars-petter-hauge, line 77: Could I ask what `ks` stands for?

lars-petter-hauge, line 77: I think we can replace this with e.g. `ks_data, active_obs, misfitval = pd.DataFrame(), pd.DataFrame(), pd.DataFrame()`

It's fairly trivial, and we don't need specific tests for that function.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+        # loop over keys and calculate the KS metric,
+        # conditioning one parameter at the time.
+        field_output = {}
+        for group_name, obs_group in combinations.items():
+            print("Processing:", group_name)
+
+            #  Use localization to evaluate change of parameters for each observation
+            update_log_path = create_path(
+                ert.getModelConfig().getRunpathAsString(), target_name
+            ).replace("scalar_", "update_log")
```
lars-petter-hauge, line 99: This doesn't change in the for loop, we can put it outside

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+            result = np.all(prior_data[dkey] == flatten_arr[0])
+            if not result:
+                p_keysf.append(dkey)
+        else:
+            if not all(x == prior_data[dkey][0] for x in prior_data[dkey]):
+                p_keysf.append(dkey)
+    return p_keysf
+
+
+def calc_ks(ks_matrix, dkeys, prior_data, target_data, key):
```
lars-petter-hauge, line 415: We should either modify or return a new instance, not both.

tests/jobs/ahm_analysis/test_integration.py
```diff
@@ -0,0 +1,102 @@
...
+    )
+    assert os.path.isfile(output_ks)
+    ks_df = pd.read_csv(output_ks)
+    assert not ks_df.empty
+    output_activeobs = output_ks.replace("ks.csv", "active_obs_info.csv")
+    assert os.path.isfile(output_activeobs)
+    output_misfitobs = output_ks.replace("ks.csv", "misfit_obs_info.csv")
+    assert os.path.isfile(output_misfitobs)
+    output_prior = output_ks.replace("ks.csv", "prior.csv")
+    assert os.path.isfile(output_prior)
```
lars-petter-hauge, line 42: Instead of manipulating, one could:

```
from pathlib import Path
outdir = "storage/snake_oil/reports/default_0/AhmAnalysisJob"
assert (outdir + "ks.csv").is_file()
...
```

semeio/workflows/ahm_analysis/ahmanalysis.py
lars-petter-hauge: Or, I don't think I understand this.. `create_path` simply returns a tmpdir? And replace wouldn't do any good? And you wouldn't get a file named `update_log` either then?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+def initialize_emptydf():
+    ks_data = pd.DataFrame()
+    active_obs = pd.DataFrame()
+    misfitval = pd.DataFrame()
+    return ks_data, active_obs, misfitval
+
+
+def make_update_log_df(update_log_file, key):
```
lars-petter-hauge, line 271: `update_log_file` -> `update_log_dir`?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+    try:
+        mygrid = Grid(input_grid, fformat="eclipserun")
+        dataframeg = mygrid.dataframe(activeonly=False)
+        return dataframeg
+    except OSError as err:
+        raise OSError("A grid with .EGRID format is expected.") from err
+
+
+def calc_delta_grid(
+    all_input_post, all_input_prior, mygrid_ok, caseobs, mygrid_ok_short
```
lars-petter-hauge, line 373: I would suggest some more appropriate names? Also we are here both returning new instance and manipulating an argument.

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,418 @@
...
+    delta grids for field parameters"""
+    all_input_post = [
+        [1 + i for i in range(8)]
+        + [10 + i for i in range(8)]
+        + [20 + i for i in range(8)],
+        [3 + i for i in range(8)]
+        + [30 + i for i in range(8)]
+        + [40 + i for i in range(8)],
+    ]
+    all_input_prior = [
```
lars-petter-hauge, line 127: Should we include input values that would benefit from the `np.absolute()`?

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: ok

tests/jobs/ahm_analysis/test_ahm_analysis.py
lars-petter-hauge: I think we should also modify the numbers slightly. As it is now, the `groupby` for instance doesn't mean anything, it could just as well be "pick the first layer". The difference in these numbers will be a constant (e.g. 2) for the entire range. `(2 + i)*2` etc would suffice.

semeio/workflows/ahm_analysis/ahmanalysis.py
oyvindeide: That was a change I made, as the `create_path` function originally did something else, and I forgot to change the inputs and the `replace`. I would suggest doing this:

```
            #  Use localization to evaluate change of parameters for each observation
            with tempfile.TemporaryDirectory() as update_log_path:
                _run_ministep(
                    ert,
                    facade,
                    obs_group,
                    field_parameters + scalar_parameters,
                    prior_name,
                    target_name,
                    update_log_path,
                )
                # Get the inactive vs total observation info
                df_update_log = make_update_log_df(update_log_path, group_name)
```
then we get automatic clean up of the temporary folder.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+                group_name,
+            )
+            field_output[group_name] = _get_field_params(
+                ert, facade.get_ensemble_size(), field_parameters, target_name
+            )
+
+        # Calculate Ks matrix for Fields parameters
+        if field_parameters:
+            # Get grid characteristics to be able to plot field avg maps
+            mygrid_ok = get_field_grid_char(ert.eclConfig().get_gridfile())
```
lars-petter-hauge, line 143: I suggest renaming `mygrid_ok`

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,418 @@
...
+
+from unittest.mock import MagicMock
+from itertools import product
+from pathlib import Path
+
+from scipy import stats
+from ecl.grid import EclGridGenerator
+
+
+@pytest.mark.usefixtures("setup_tmpdir")
```
lars-petter-hauge, line 18: Don't think we need a tmpdir here, we are only reading from a file using pandas.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+    return ks_data, active_obs, misfitval
+
+
+def make_update_log_df(update_log_file, key):
+    """ Read update_log file to get active and inactive observations """
+    list_of_files = [
+        os.path.join(update_log_file, f) for f in os.listdir(update_log_file)
+    ]
+    if len(list_of_files) > 1:
+        raise OSError("ERROR more than one update_log_file.")
```
lars-petter-hauge, line 277: As we are making the paths to updatelog from tmpdirs, this would really be a bug, right? I don't mind having it in though, just checking.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+        "obs_key",
+        "obs_mean",
+        "obs_std",
+        "status",
+        "sim_mean",
+        "sim_std",
+    ]
+    # drop last line ('=====...====')
+    # drop last row i.e. end of file line
+    updatelog.drop(updatelog.tail(1).index, inplace=True)
```
lars-petter-hauge, line 298: We are already skipping rows, you can also drop last line by using `skipfooter`

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+    # ---------------------------------
+    # add proper name for obs_key in rows where value is '...'
+    replace = "..."
+    for i in range(len(updatelog.index)):
```
lars-petter-hauge, line 302: Maybe I'm misunderstanding this, but the test data is fairly simple, as it only contains the single observation. With an input of more than a single observation, all the `...` gets the same value, e.g. using `allobs` test data:
```
  36  : RWI_3_OBS                                285.000 +/-            3.000  Active   |           273.958 +/-          10.529
  37  :   ...                                    258.000 +/-            3.000  Active   |           264.119 +/-           9.031
  38  :   ...                                    271.000 +/-            3.000  Active   |           269.818 +/-           5.951
  39  : RWI_2_OBS                                291.000 +/-            3.000  Active   |           272.672 +/-          14.870
  40  :   ...                                    233.000 +/-            3.000  Active   |           265.408 +/-          14.494
  41  :   ...                                    297.000 +/-            3.000  Active   |           277.500 +/-          13.479
  42  : ROP_5_OBS                                285.000 +/-            3.000  Active   |           274.673 +/-           6.869
  43  :   ...                                    279.000 +/-            3.000  Active   |           271.760 +/-           5.229
```

```
5          36    RWI_3_OBS    285.00     3.00    Active   273.958   10.529
36          37    RWI_3_OBS    258.00     3.00    Active   264.119    9.031
37          38    RWI_3_OBS    271.00     3.00    Active   269.818    5.951
38          39    RWI_2_OBS    291.00     3.00    Active   272.672   14.870
39          40    RWI_3_OBS    233.00     3.00    Active   265.408   14.494
40          41    RWI_3_OBS    297.00     3.00    Active   277.500   13.479
```

Is it only made to handle single observations in the updagelog perhaps?

lars-petter-hauge: Here's a suggestion to use pandas `ffill` instead.

```
    df = pd.read_csv(
        list_of_files[0],
        delim_whitespace=True,
        skiprows=6,
        usecols=[0, 2, 3, 5, 6, 8, 10],
        header=None,
        skipfooter=1,
    )
    df.columns = [
        "obs_number",
        "obs_key",
        "obs_mean",
        "obs_std",
        "status",
        "sim_mean",
        "sim_std",
    ]
    # Identical values for obs_key will have the obs_value in first record
    # and subsequently "..." for the following records. We want an exhaustive list
    df.replace("...", np.nan, inplace=True) # ffill requires np.nan
    df.ffill(inplace=True)

    return df
```

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,418 @@
...
+def test_make_update_log_df(test_data_root):
+    """test function creates a dataframe from update_log file
+    and replace '...' with key_obs"""
+    test_data_dir = os.path.join(test_data_root, "update_log/oneobs")
```
lars-petter-hauge, line 22: I mentioned that the input test case was a bit trivial, and for instance the `allobs` updatelog would yield an unexpected result. If one is to replace with `allobs` test case, this test wouldn't fail on the unexpected issue, it would merely fail on the status not being `Active` at the same indexes (a happy coincident perhaps).

My point is, let's be more exhaustive of what we want to test, let's assert the output instead (or a larger part of it).

We could for instance assert the part we're mostly interested in, where there have been changes to the `...` value:

```
expected = pd.DataFrame(
        columns=["obs_key", "obs_mean"],
        data=[
            ("OP_1_WGOR2", 170.00),
            ("RWI_3_OBS", 285.00),
            ("RWI_3_OBS", 258.00),
            ("RWI_3_OBS", 271.00),
            ("RWI_2_OBS", 291.00),
            ("RWI_2_OBS", 233.00),
            ("RWI_2_OBS", 297.00),
            ("ROP_5_OBS", 285.00),
            ("ROP_5_OBS", 279.00),
            ("ROP_5_OBS", 288.00),
            ("ROP_4_OBS", 292.00),
            ("ROP_4_OBS", 268.00),
            ("ROP_4_OBS", 280.00),
        ],
    )
```

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,418 @@
...
+        ("update_log/oneobs", "RWI_3_OBS", "2 active/3"),
+        ("update_log/allobs", "ALL_OBS", "41 active/47"),
+        ("update_log/noactive", "RWI_2_OBS", "0 active/3"),
+    ],
+)
+@pytest.mark.usefixtures("setup_tmpdir")
+def test_list_active_observations(test_data_root, input_dir, expected_result, obs_key):
+    """test function creates a dataframe reporting active observations during update"""
+    test_data_dir = os.path.join(test_data_root, input_dir)
+    df_update_log = ahmanalysis.make_update_log_df(test_data_dir, obs_key)
```
lars-petter-hauge, line 55: Reading this test, I wonder.. The `make_update_log_df` would yield:

```
35         36    RWI_3_OBS    285.00     3.00    Active   273.958   10.529
36         37      ALL_OBS    258.00     3.00    Active   264.119    9.031
37         38      ALL_OBS    271.00     3.00    Active   269.818    5.951
38         39    RWI_2_OBS    291.00     3.00    Active   272.672   14.870
39         40      ALL_OBS    233.00     3.00    Active   265.408   14.494
40         41      ALL_OBS    297.00     3.00    Active   277.500   13.479
41         42    ROP_5_OBS    285.00     3.00    Active   274.673    6.869
42         43      ALL_OBS    279.00     3.00    Active   271.760    5.229
43         44      ALL_OBS    288.00     3.00    Active   270.670    5.802
44         45    ROP_4_OBS    292.00     3.00    Active   279.577    8.164
45         46      ALL_OBS    268.00     3.00    Active   273.585    6.250
46         47      ALL_OBS    280.00     3.00    Active   273.649    6.049
```

For the input:

 ```
 36  : RWI_3_OBS                                285.000 +/-            3.000  Active   |           273.958 +/-          10.529
  37  :   ...                                    258.000 +/-            3.000  Active   |           264.119 +/-           9.031
  38  :   ...                                    271.000 +/-            3.000  Active   |           269.818 +/-           5.951
  39  : RWI_2_OBS                                291.000 +/-            3.000  Active   |           272.672 +/-          14.870
  40  :   ...                                    233.000 +/-            3.000  Active   |           265.408 +/-          14.494
  41  :   ...                                    297.000 +/-            3.000  Active   |           277.500 +/-          13.479
  42  : ROP_5_OBS                                285.000 +/-            3.000  Active   |           274.673 +/-           6.869
  43  :   ...                                    279.000 +/-            3.000  Active   |           271.760 +/-           5.229
  44  :   ...                                    288.000 +/-            3.000  Active   |           270.670 +/-           5.802
  45  : ROP_4_OBS                                292.000 +/-            3.000  Active   |           279.577 +/-           8.164
  46  :   ...                                    268.000 +/-            3.000  Active   |           273.585 +/-           6.250
  47  :   ...                                    280.000 +/-            3.000  Active   |           273.649 +/-           6.049
```

Is that the intention?

File: semeio/workflows/ahm_analysis/ahmanalysis.py [hunk 50]
```diff
@@ -0,0 +1,456 @@
...
+                name = "All_obs"
+            combinations[name.replace(":", "_")] = obs_group
+    return combinations
+
+
+def list_active_observations(df_update_log):
+
+    """To get the inactive vs total observation info."""
+    df_active = df_update_log[(df_update_log["status"] == "Active")]
+    return str(len(df_active)) + " active/" + str(len(df_update_log.index))
```
lars-petter-hauge, line 343: I think we should definitively avoid returning results as a string. This should rather be some sort of datastructure that whoever will write to file can format in their own way.

Consider if other functions calling this function would rather use the numbers, and that would result in them starting to parse this string. That's annoying on its own, but now you can't modify the format without breaking the other functions.

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,418 @@
...
+@pytest.mark.parametrize(
+    "input_dir, obs_key, expected_result",
+    [
+        ("update_log/oneobs", "RWI_3_OBS", "2 active/3"),
+        ("update_log/allobs", "ALL_OBS", "41 active/47"),
+        ("update_log/noactive", "RWI_2_OBS", "0 active/3"),
+    ],
+)
+@pytest.mark.usefixtures("setup_tmpdir")
+def test_list_active_observations(test_data_root, input_dir, expected_result, obs_key):
```
lars-petter-hauge, line 52: The `list_active_observations` takes a dataframe, I would consider avoiding to use `make_update_log_df` to create the input, and rather add the input data directly in `parametrize`. Would make it much easier to see what `list_active_observations` does as well, and why the expected results is as it is.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+
+def create_path(runpathdf, target_name):
+    """to create path for output data"""
+    return tempfile.mkdtemp()
+
+
+def list_observations_misfit(obs_keys, df_update_log, misfit_df):
+    """To get the misfit for total observations (active/inactive)."""
+    misfit_values = []
+    total_obs_nr = len(df_update_log[df_update_log.status.isin(["Active", "Inactive"])])
```
lars-petter-hauge, line 354: Can the status be anything other than `Active` or `Inactive`?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+    for key in obs_keys:
+        misfit_name = "MISFIT:" + key
+        misfit_values.append(misfit_df[misfit_name].mean())
+    return np.sum(misfit_values) / total_obs_nr
```
lars-petter-hauge, line 358: Should we do an early return if `df_update_log` is empty? otherwise this is a zero division error. If the updatelog indeed is empty, and we consider it a hard fail, this is likely not the place to handle it.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+    return tempfile.mkdtemp()
+
+
+def list_observations_misfit(obs_keys, df_update_log, misfit_df):
+    """To get the misfit for total observations (active/inactive)."""
+    misfit_values = []
+    total_obs_nr = len(df_update_log[df_update_log.status.isin(["Active", "Inactive"])])
+    for key in obs_keys:
+        misfit_name = "MISFIT:" + key
+        misfit_values.append(misfit_df[misfit_name].mean())
```
lars-petter-hauge, line 357: There is no reason to expect this to not have `MISFIT:<key>` for all keys I suppose?

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,418 @@
...
+            ["RWI_3_OBS", "OP_3_WWCT1"],
+            7.0 / 4.0,
+            {"status": {0: "Active", 1: "Inactive", 2: "Active", 3: "Active"}},
+        ),
+        (["OP_3_WWCT1"], 1.0, {"status": {0: "Active"}}),
+    ],
+)
+def test_list_observations_misfit(input_obs, expected_misfit, update_log):
+    """test function creates a dataframe
+    reporting misfit data for each obs vector except the ones with All_obs-"""
```
lars-petter-hauge, line 74: There is nothing in the test, nor the function itself that say anything about `All_obs-`? But there shouldn't be either, right?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+    df_active = df_update_log[(df_update_log["status"] == "Active")]
+    return str(len(df_active)) + " active/" + str(len(df_update_log.index))
+
+
+def create_path(runpathdf, target_name):
+    """to create path for output data"""
+    return tempfile.mkdtemp()
+
+
+def list_observations_misfit(obs_keys, df_update_log, misfit_df):
```
lars-petter-hauge, line 351: Perhaps this should be renamed. We also have the `list_active_observations`, which behaves quite differently, although the name is quite similar. Neither returns a list though.

semeio/workflows/ahm_analysis/ahmanalysis.py
lars-petter-hauge: Also, is it so that the dataframe containing misfit will have one `MISFIT` prefixed observation per observation in total? Should `len(obs_keys)` always be the same as `total_obs_nr`?  or can they be completely different?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+def check_inputs(misfitdata, prior_data):
+    """Check input ensemble prior is not empty
+    and if ensemble contains parameters for hm"""
+    if misfitdata.empty:
+        raise ValidationError("Empty prior ensemble")
+    if prior_data.empty:
+        raise ValidationError("Empty parameters set for History Matching")
+    return prior_data
```
lars-petter-hauge, line 440: Not sure if I get the reasoning behind this function. We simply pass along two dataframes (or any instance with an `empty` property), and if neither is `empty` we return the second argument.

In the `run` function, the `misfitdata` is loaded in the function call, but it is never used. All we do is check that it isn't empty, is that vital for this job to work? Is it merely a sanity check?

lars-petter-hauge: If this is the intended behaviour though I suggest that we don't return (since it's the same argument value), and we are explicit about what we do:

```
def raise_if_empty(dataframes, messages):
    for df, msg in zip(dataframes, messages):
        if df.empty:
            raise ValidationError(f"{message}")
```

And then call with:

```
prior_data = GenKwCollector.loadAllGenKwData(ert, prior_name)
raise_if_empty(
    dataframes=[prior_data, MisfitCollector.loadAllMisfitData(ert, prior_name)],
    messages=[
        "Empty prior ensemble",
        "Empty parameters set for History Matching",
    ],
)
```

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,418 @@
...
+                "obs_key_1": "data_key_1",
+                "obs_key_2": "data_key_2",
+                "obs_key_3": "data_key_2",
+            },
+            {"data_key_1": ["obs_key_1"], "data_key_2": ["obs_key_2", "obs_key_3"]},
+            id="Multiple obs for multiple data keys",
+        ),
+    ],
+)
+def test_group_obs_data_key(keys, expected_result):
```
lars-petter-hauge, line 395: Should we include some parts with columns, e.g. `obs:key_1`, since that is part of the functionality as well?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+def _get_parameter_keys(genkw_parameters, prior_keys):
+    """excluding duplicate transformed parameter keys"""
+    transformed_param_keys = []
+    for parameter in genkw_parameters:
+        for key in prior_keys:
+            if "_" + parameter + ":" in key:
+                transformed_param_keys.append(key)
+    return set(prior_keys).difference(transformed_param_keys)
```
lars-petter-hauge, line 392: This function doesn't really get anything, it rather filters. It also contains a lot of context on how ert formats data. If we agree that what ert would do is to prefix a parameter with the transformation name, then this is a suggestion to make it more readable:

```
def _filter_on_prefix(list_of_strings, prefixes):
    """Returns the set of strings that has a match for any of the given prefixes"""
    return {
        string
        for string in list_of_strings
        if any(string.startswith(prefix) for prefix in prefixes)
    }
```

And then where it is used we can call it by:

`parameter_keys = _filter_on_prefix(list_of_strings=prior_data.keys(), prefixes=parameters)`

Now at least we don't have to wonder what `get_parameter_keys` does..

Its a suggestion at least

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+def get_updated_parameters(prior_data, parameters):
+    """make list of updated parameters
+    (excluding duplicate transformed parameters)
+    """
+    parameter_keys = _get_parameter_keys(parameters, prior_data.keys())
+    # remove parameters with constant prior distribution
+    p_keysf = []
+    for dkey in parameter_keys:
+        if prior_data[dkey].ndim > 1:
+            warnings.warn("WARNING: Parameter " + dkey + " defined several times.")
```
lars-petter-hauge, line 404: I don't think any test hits this part, and I'm a bit unsure of the intent.. Would it be possible to expand the test with a dataset that would get here?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+        misfit_values.append(misfit_df[misfit_name].mean())
+    return np.sum(misfit_values) / total_obs_nr
+
+
+def get_field_grid_char(input_grid):
+    """Get field grid characteristics/coordinates"""
+    input_grid = Path(input_grid).with_suffix("")
+    try:
+        mygrid = Grid(input_grid, fformat="eclipserun")
+        dataframeg = mygrid.dataframe(activeonly=False)
```
lars-petter-hauge, line 366: I suggest renaming slightly. We can use `load` instead of `get` and perhaps not use `char` which is very much linked to `characters`, thus I would expect such a function to return a string or equivalent.

How about:
```
def load_grid_to_dataframe(filepath):
    fpath = Path(filepath).with_suffix("")
    try:
        grid = Grid(fpath, fformat="eclipserun")
        return grid.dataframe(activeonly=False)
    except...
```
?

semeio/workflows/ahm_analysis/ahmanalysis.py
oyvindeide: Yes it would, though it is more of a remnant of the old implementation. Could potentially remove it, but not sure how important that is.

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,456 @@
...
+                self.reporter.publish_csv("delta_field" + fieldparam, mygrid_ok_short)
+        # save/export the Ks matrix, active_obs, misfitval and prior data
+        self.reporter.publish_csv("ks", ks_data)
+        self.reporter.publish_csv("active_obs_info", active_obs)
+        self.reporter.publish_csv("misfit_obs_info", misfitval)
+        self.reporter.publish_csv("prior", prior_data)
+
+
+def _run_ministep(
+    ert, facade, obs_group, data_parameters, prior_name, target_name, output_path
```
lars-petter-hauge, line 192: I don't think facade is necessary here.. Can't we just call `ert` directly? If we were able to completely swap out `ert` with facade I would agree using that instead, but otherwise I think using `ert.getLocalConfig().getUpdatestep().attachMinistep(ministep)`

And considering the `getLocalConfig`, let's make a `local_config = ert.getLocalConfig()` while we're at it and use that instead :)

lars-petter-hauge:
We're using pandas dataframe several places here, how does this job perform with cases of large grid files and numerous observations?

tests/jobs/ahm_analysis/test_integration.py
mareco701: I have put more specific test

tests/jobs/ahm_analysis/test_integration.py
mareco701: I have added some more concrete test on min/max Ks value and expected groups of observations in the KS dataframe, Is it enough ?

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: I agree it would be a bug and if there are several files the problem the script picks the first file which does not mean it has the data expected in it so output results about active/inactive data) could be completely wrong...

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: done

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: yes  it can be Missing if the point/data is outside the grid/has no simulated data.

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: done

tests/jobs/ahm_analysis/test_integration.py
mareco701: done

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: done

tests/jobs/ahm_analysis/test_integration.py
mareco701: I have added some more tests to check all obs files are created

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: renamed to grid_xyzcenter.

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: I guess we need the tempdir since the functions search files in a dir...

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: updated to only return the number of active obs (function is renamed get_nb_active_observations)

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: I have updated the test.

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: option without comprehension is now implemented

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: changed names plus removed returning and manipulating an argument.

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: ok, done

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: ffill implemented (was not initially since there was an issue in updatelog: when the first point in gan_data had no simulated data/point outside the grid for instance then the first data would be ... so not possible to know to which well it belonged, but this is now fixed so ffill works perfect now)

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: done, allobs updatelog is used and check is made that lines with ... in input data are replaced by names of the observation.

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: done

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: as explained before initially yes when there was the bug in the updatelog file not reporting missing data. But I agree the output now should use ffill and the ... are to be replaced by the correct obs name.

tests/jobs/ahm_analysis/test_ahm_analysis.py
lars-petter-hauge: It does, but the path provided to `make_update_log_df` is simply a path in source dir (as we are only reading, and not writing, that will suffice): `root/semeio/tests/legacy_test_data/update_log/allobs`

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,466 @@
...
+    return p_keysf
+
+
+def calc_kolmogorov_smirnov(columns, prior_data, target_data):
+    """Calculate kolmogorov_smirnov matrix"""
+    ks_param = {}
+    for dkey in sorted(columns):
+        ks_val = ks_2samp(prior_data[dkey], target_data[dkey])
+        ks_param[dkey] = ks_val[0]
+    #        ks_matrix.at[dkey, key] = ks_val[0]
```
lars-petter-hauge, line 432: Suppose the comment here can be removed, guess it snuck inn during refactor :)

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: I think ert will not generate an empty update_log file. I have tested with cases where all obs in an obs_group are inactive and then the update is not performed, the script goes through with message:
image

mareco701: the dataframe containing Misfit from Ert actually have only one value for example the gen_data obs (it will be the sum of the 3 misfit values) and the summary_data obs have one misfit value per obs data. So len(obs_keys) can be different from total_obs_nr.

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: there could be if one key has a status missing, I have changed the function to handle this case.

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: the part except the ones with All_obs has been removed (should not be there)

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: renamed to calc_observations_misfit

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: suggestion ahs been implemented. The point of the checks is to not start calulcations if some inputs data are missing.

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: done

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: suggestion implemented

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: a test has been added

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: suggestion implemented

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: done

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: yes :-)

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: ok, I update this

mareco701: done

lars-petter-hauge:
Could you please add this workflow to the testing for pluginmanager as well?

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,465 @@
...
+                missing_obs = [x for x in key_map.keys() if x not in set(subset)]
+                assert len(missing_obs) == 1
+                name = f"All_obs-{missing_obs[0]}"
+            else:
+                name = "All_obs"
+            combinations[name.replace(":", "_")] = obs_group
+    return combinations
+
+
+def get_nb_active_observations(df_update_log):
```
lars-petter-hauge, line 349: I would suggest renaming this to `count_active_observations`

semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -0,0 +1,465 @@
...
+                itertools.chain.from_iterable([key_map[key] for key in subset])
+            )
+            if combination_length == len(key_map.keys()) - 1:
+                missing_obs = [x for x in key_map.keys() if x not in set(subset)]
+                assert len(missing_obs) == 1
+                name = f"All_obs-{missing_obs[0]}"
+            else:
+                name = "All_obs"
+            combinations[name.replace(":", "_")] = obs_group
+    return combinations
```
lars-petter-hauge, line 346: I think we can rewrite this slightly as well, putting the edge case outside, thus dropping a special case:

```
    combinations = key_map.copy()
    if len(combinations) == 1:
        return combinations

    combinations["All_obs"] = list(itertools.chain.from_iterable(key_map.values()))

    # Only two observation groups present, not necessary to
    # generate one off combinations
    if len(combinations) == 3:
        return combinations

    for subset in itertools.combinations(key_map.keys(), len(key_map.keys()) - 1):
        obs_group = list(
            itertools.chain.from_iterable([key_map[key] for key in subset])
        )
        missing_obs = [x for x in key_map.keys() if x not in set(subset)]
        assert len(missing_obs) == 1
        name = f"All_obs-{missing_obs[0]}"
        combinations[name.replace(":", "_")] = obs_group

    return combinations
```

Should be good @oyvindeide ?

tests/jobs/ahm_analysis/test_ahm_analysis.py
```diff
@@ -0,0 +1,421 @@
...
+                "data_key_3": ["obs_6"],
+            },
+            [
+                "data_key_1",
+                "data_key_2",
+                "data_key_3",
+                "All_obs-data_key_3",
+                "All_obs-data_key_2",
+                "All_obs-data_key_1",
+                "All_obs",
```
lars-petter-hauge, line 269: Before adding the suggestion on implementation I exhausted this expected by including the values as well, I would argue that is a good thing to better test and document what it does:

```
(
            {"data_key_1": ["obs_1", "obs_2", "obs_3"]},
            {"data_key_1": ["obs_1", "obs_2", "obs_3"]},
        ),
        (
            {"data_key_1": ["obs_1", "obs_2", "obs_3"], "data_key_2": ["obs_4"]},
            {
                "data_key_1": ["obs_1", "obs_2", "obs_3"],
                "data_key_2": ["obs_4"],
                "All_obs": ["obs_1", "obs_2", "obs_3", "obs_4"],
            },
        ),
        (
            {
                "data_key_1": ["obs_1", "obs_2", "obs_3"],
                "data_key_2": ["obs_4"],
                "data_key_3": ["obs_6"],
            },
            {
                "data_key_1": ["obs_1", "obs_2", "obs_3"],
                "data_key_2": ["obs_4"],
                "data_key_3": ["obs_6"],
                "All_obs-data_key_3": ["obs_1", "obs_2", "obs_3", "obs_4"],
                "All_obs-data_key_2": ["obs_1", "obs_2", "obs_3", "obs_6"],
                "All_obs-data_key_1": ["obs_4", "obs_6"],
                "All_obs": ["obs_1", "obs_2", "obs_3", "obs_4", "obs_6"],
            },
        ),
```

semeio/workflows/ahm_analysis/ahmanalysis.py
oyvindeide: Looks cleaner 👍

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: ok done

tests/jobs/ahm_analysis/test_ahm_analysis.py
mareco701: ok done

semeio/workflows/ahm_analysis/ahmanalysis.py
mareco701: ok done

lars-petter-hauge:
Unless you disagree, I think this can be a single commit. Could you rebase and squash all commits please?

lars-petter-hauge (approved):
Please note that currently the documentation part is empty. We need to include relevant information there

PR #232, opened 2020-10-12
Change default tox to use python version on path

oyvindeide:


oyvindeide:
Not really feasible with gh actions, closing.

PR #237, opened 2020-10-15
Update deprecated commands in GitHub Actions

### LINKED ISSUE #236: Change commands "add-path" and "set-env" in Github Actions to using Environment Files
The `add-path` and `set-env` commands are deprecated and will be disabled soon. We need to upgrade to using Environment Files. More info can be found here: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

kvashchuka:
Closes #236

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -22,9 +22,9 @@ jobs:
 
       - name: Setup environment
         run: |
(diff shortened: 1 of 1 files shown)


oyvindeide (approved):
LGTM!

Issue #235, opened 2020-10-15
[DESIGN2PARAMS] Improve error message, pick parameter name


berland:
design2params emits error messages when cells are empty in the design matrix, and says that "Realization 31, key 151" is empty f.ex. To locate the cell in Excel, the user has to calculate that 151 corresponds to column "EV" in excel terms, which is non-trivial. 

design2params could try to obtain the parameter name from row zero and include that in the error message instead.

berland:
This can be closed

Issue #234, opened 2020-10-15
[DESIGN2PARAMS] Crashes for all realization for realization-specific errors


berland:
If a cell in a design matrix cell is empty for one parameter in only one realization, the design2params job fails hard, even though it only needs to fail hard in the realization with the error.

berland:
This can be  closed

PR #238, opened 2020-10-16
Allow missing values in other realizations than current.

berland:
Improve warning message pointing to empty cells

Solves #234 and #235

### DIFF
--- a/tests/jobs/design2params/test_design2params.py
+++ b/tests/jobs/design2params/test_design2params.py
@@ -241,16 +241,35 @@ def test_open_excel_file_header_missing(input_data):
         )
 
 
-def test_open_excel_file_value_missing(input_data):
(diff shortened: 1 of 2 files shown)


tests/jobs/design2params/test_design2params.py
```diff
@@ -304,7 +323,7 @@ def test_existing_parameterstxt(
 def test_open_excel_file_wrong_defaults(input_data):
```
berland, line 323: (I don't think this test is doing what its name suggests)

berland: as it stands, it can be deleted without loosing test coverage.

berland: Changed.

oysteoh (approved):
👍

PR #239, opened 2020-10-16
Add test runs on MacOS

### LINKED ISSUE #71: Add Mac OS to CI test matrix
Currently we are only running our CI in a linux environment. I'm experiencing failing tests on my MacBook and suggest that we try adding MacOS to our build matrix.

kvashchuka:
Closes #71

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -8,11 +8,15 @@ env:
 
 jobs:
   tests:
-    name: "Python ${{ matrix.python-version }}"
-    runs-on: "ubuntu-latest"
+    name: "Python ${{ matrix.python-version }} on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         python-version: ["3.6", "3.7", "3.8"]
+        os: [ ubuntu-latest ]
+        include:
+          - python-version: 3.6
+            os: macos-latest
 
     steps:
       - uses: "actions/checkout@v2"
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yml
```diff
@@ -40,5 +46,9 @@ jobs:
       - name: "Install dependencies"
         run: python -m pip install tox tox-gh-actions
 
-      - name: "Run tox targets for ${{ matrix.python-version }}"
+      - name: "Run tox targets on Ubuntu for ${{ matrix.python-version }}"
+        if: matrix.os == 'ubuntu-latest'
+        run: "python -m tox"
+      - name: "Run tox targets on MacOS for ${{ matrix.python-version }}"
+        if: matrix.os == 'macos-latest'
         run: "python -m tox"
```
sondreso, line 54: You don't need to separate these since they are running the same command :slightly_smiling_face: The reason why it's different in ERT is that they have slightly different commands!

kvashchuka: Fair point

.github/workflows/testing.yml
```diff
@@ -8,11 +8,16 @@ env:
 
 jobs:
   tests:
-    name: "Python ${{ matrix.python-version }}"
-    runs-on: "ubuntu-latest"
+    name: "Python ${{ matrix.python-version }} on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
oyvindeide, line 14: Any particular reason why dont want to `fail-fast`?

kvashchuka: Not really, will remove it :)

oyvindeide (approved):
LGTM!

Issue #243, opened 2020-10-22
[design2params] Warn, fail or fix trailing whitespace in parameter names.


berland:
If a user accidentally includes trailing whitespace in the parameter names in the design matrix excel sheet, `design_kw` will fail with (correct) error messages that the parameter does not exist. 

`2020-10-22 13:07:13,236 - semeio.jobs.design_kw.design_kw - ERROR - <H10_START> not found in design matrix`

This is non-trivial for users to notice, as the parameter seemingly exists in the Excel gui.

On disk, the problem can be identified by looking in `parameters.txt`:

```
$ grep "H10_START" parameters.txt
"H10_START " 4350
```
where the trailing space is evident. 

This can be solved by one of:
1. Disallow trailing whitespace, and error hard
2. Give warning from design2params (but these warnings are not likely to be discovered, as the failure occurs potentially much later in `design_kw`)
3. Strip trailing whitespace from parameter names in design matrix (and in defaultssheet)
4. Modify design_kw to look for this specific situation, and then pick the parameter with the trailing whitespace (this sounds ugly)
5. Do nothing.

Trailing whitespace in cell values should be conserved, as it is a feature that may already be in use.

berland:
If noone can come up with a use-case for a trailing white-space in a parameter name, my vote goes towards 1.

markusdregi:
I'm also in favor of 1 :+1:

oyvindeide:
Solved by: https://github.com/equinor/semeio/pull/244

PR #244, opened 2020-10-23
Disallow whitespace in parameter names in design2params

berland:
Solves #243

### DIFF
--- a/semeio/jobs/design2params/design2params.py
+++ b/semeio/jobs/design2params/design2params.py
@@ -203,6 +203,16 @@ def _validate_design_matrix(design_matrix):
     if len(empties) > 0:
         raise SystemExit("Design matrix contains empty cells {}".format(empties))
 
+    # Look for initial or trailing whitespace in column headers. This
+    # is disallowed as it can create user confusion and has no use-case.
+    for col_header in design_matrix:
+        if col_header != col_header.strip():
+            raise SystemExit(
+                'Column header "{}" contains initial or trailing whitespace.'.format(
+                    col_header
+                )
+            )
+
 
 def _read_defaultssheet(xlsfilename, defaultssheetname):
     """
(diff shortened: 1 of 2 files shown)


semeio/jobs/design2params/design2params.py
```diff
@@ -203,6 +203,14 @@ def _validate_design_matrix(design_matrix):
     if len(empties) > 0:
         raise SystemExit("Design matrix contains empty cells {}".format(empties))
 
+    # Look for initial or trailing whitespace in column headers. This
+    # is disallowed as it can create user confusion and has no use-case.
+    for col_header in design_matrix:
+        if col_header != col_header.strip():
+            raise SystemExit(
+                'Column header "{}" contains whitespace.'.format(col_header)
```
markusdregi, line 211: Perhaps we should change the message to say that is has initial or trailing whitespaces? Would be good if the condition and the message corresponds ;)

tests/jobs/design2params/test_design2params.py
```diff
@@ -424,3 +424,67 @@ def test_pair_cell_values(cellvalues, expected_parameters_strs, tmpdir):
...
+
+
+@pytest.mark.parametrize(
+    "paramnames",
+    (
+        [
+            ["REAL", "SOMEKEY "],
+            ["REAL", " SOMEKEY"],
+            ["REAL", " SOMEKEY "],
+            ["REAL", "\tSOMEKEY"],
```
markusdregi, line 436: Tabs :heart:

semeio/jobs/design2params/design2params.py
berland: Yes. Good point!

tests/jobs/design2params/test_design2params.py
berland: Not sure it is possible to even inject through Excel..

markusdregi (approved):
Besides fixing the message this looks good :+1:

berland:
Ready to 🚢

PR #252, opened 2020-11-03
Catch pandas invalid header

oyvindeide:


berland:
Can you add to the denylist testing something as:
`@pytest.mark.parametrize("paramname", design2params.DENYLIST_DEFAULTS + [1] + [1.1])
`
(both for defaults and for design-matrix)

oyvindeide:
> Can you add to the denylist testing something as:
> `@pytest.mark.parametrize("paramname", design2params.DENYLIST_DEFAULTS + [1] + [1.1]) `
> (both for defaults and for design-matrix)

Do you mean something like:
```
@pytest.mark.parametrize("paramname", [1, 1.0])
def test_invalid_pandas_header_error(tmpdir, paramname):
    tmpdir.chdir()

    write_design_xlsx(
        "design_matrix.xlsx",
        designdf=pd.DataFrame(columns=["REAL", paramname], data=[[0, "foo"]]),
    )

    expected_error = "Invalid value in design matrix header"
    with pytest.raises(SystemExit, match=expected_error):
        design2params.run(
            0,
            "design_matrix.xlsx",
        )
``` 
That actually fails for another reason

oyvindeide:
To pass that has to be: 
```
@pytest.mark.parametrize("paramname", [1, 1.0])
def test_invalid_pandas_header_error(tmpdir, paramname):
    tmpdir.chdir()

    write_design_xlsx(
        "design_matrix.xlsx",
        designdf=pd.DataFrame(columns=["REAL", paramname], data=[[0, "foo"]]),
    )

    expected_error = "Location based indexing can only have"
    with pytest.raises(ValueError, match=expected_error):
        design2params.run(
            0,
            "design_matrix.xlsx",
        )
```

berland:
Yes, either `expected_error`must be adapted, or there must be code in `design2params.py` that gives a similar error message.

berland:
Can you add [1, 1.0] to the defaults test as well? (in defaults sheet it is easier for users to inject numerical parameter names)

oyvindeide:
Like so? It will error in the same way, is that expected/ok?

berland:
> Like so? It will error in the same way, is that expected/ok?

LGTM!

tests/jobs/design2params/test_design2params.py
```diff
@@ -432,7 +433,11 @@ def test_headers_trailing_whitespace(paramnames, tmpdir):
             0,
             "design_matrix.xlsx",
         )
-    with pytest.raises(SystemExit, match="whitespace"):
+    if version.parse(pd.__version__) < version.parse("0.25"):
```
lars-petter-hauge, line 436: I wonder how long we should support pandas < 1. At least we should make an issue. I wanted to suggest a due date on said issue, but searching for the possibility of adding due dates to issues was quite depressing: https://github.com/isaacs/github/issues/432

Looks like a lot of people still want it though!

semeio/jobs/design2params/design2params.py
```diff
@@ -181,22 +181,27 @@ def _read_excel(file_name, sheet_name, header=0):
...
 
 def _validate_design_matrix_header(design_matrix):
     """
     Validate header in user inputted design matrix
     :raises: SystemExit if design matrix contains empty headers
     """
-    unnamed = design_matrix.loc[:, design_matrix.columns.str.contains("^Unnamed")]
+    try:
+        unnamed = design_matrix.loc[:, design_matrix.columns.str.contains("^Unnamed")]
+    except (AttributeError, ValueError) as err:
```
lars-petter-hauge, line 201: Let's add a comment stating why we're catching these two errors, as they originally stem from the same issue (or close to?). Never the less, whenever we stop supporting pandas < 1.0, we only need to catch one of them.

tests/jobs/design2params/test_design2params.py
oyvindeide: Agreed, I think the hope is for another month. https://github.com/equinor/semeio/issues/254

semeio/jobs/design2params/design2params.py
```diff
@@ -199,6 +199,10 @@ def _validate_design_matrix_header(design_matrix):
     try:
         unnamed = design_matrix.loc[:, design_matrix.columns.str.contains("^Unnamed")]
     except (AttributeError, ValueError) as err:
+        # We catch these errors because:
+        # 1) AttributeError is a pandas 0.24 compatilibity
```
berland, line 203: typo: compatibility

oyvindeide:
I considered adding a separate test for pandas `0.24.2`, but opted against it as we would remove it in a month or so. You will have to believe me that I tested locally (or do so yourself ;))

lars-petter-hauge (approved):
Good to go!

PR #253, opened 2020-11-03
Skip tests on osx

oyvindeide:
Skipping these tests inside the scope of the test means that the test is
actually run. That is a problem because that means libres will do chmod
on some files, which is not desirable for tests.


lars-petter-hauge (approved):
LGTM

PR #255, opened 2020-11-05
Fix bug where summary data was not updated when no index list was given

oyvindeide:


sondreso (approved):
Need to pin the ERT version in the pipeline, otherwise this looks good!

PR #256, opened 2020-11-09
Add trajectory file error handling to entry point

### LINKED ISSUE #34: Can not raise argparse.ArgumentError without catching outside parsing
The simple `raise argparse.ArgumentError` works as long as you are within the context of parsing. If the exception is raised after the parsing of arguments (e.g. in `gendata_rft`, where we are loading file content after the parsing is completed):

```
(ert_36) [lpha@be-lx923569 norne]$ python /work/GitProjects/semeio/semeio/jobs/scripts/gendata_rft.py -e NORNE_ATW2013 -t gendata_rft_input_files -w gendata_rft_input_files/well_and_time.txt
  File "/work/GitProjects/semeio/semeio/jobs/scripts/gendata_rft.py", line 127, in <module>
    main_entry_point(sys.argv[1:])
  File "/work/GitProjects/semeio/semeio/jobs/scripts/gendata_rft.py", line 110, in main_entry_point
    for wname in well_names
  File "/work/GitProjects/semeio/semeio/jobs/scripts/gendata_rft.py", line 110, in <dictcomp>
    for wname in well_names
  File "/work/venvs/ert_36/lib/python3.6/site-packages/semeio/jobs/rft/trajectory.py", line 98, in load_from_file
    filename=filename
argparse.ArgumentTypeError: Warning: Trajectory file gendata_rft_input_files/B-1AH.txt not found!
```

I think the best way of handling this is to pass the error upwards, preferably by catching the thrown error inside the `main_entry_point` and then calling `parser.error(exception.msg)`

oyvindeide:
Closes #34

### DIFF
--- a/semeio/jobs/rft/trajectory.py
+++ b/semeio/jobs/rft/trajectory.py
@@ -260,9 +260,9 @@ class Trajectory:
             floats = [float(v) for v in point[:4]]
         except ValueError as err:
             raise ValueError(
-                "Error: Failed to extract data from line {line}. Expected the format "
+                f"Error: Failed to extract data from line {line}. Expected the format "
                 "'utm_x utm_y md tvd zone' - zone is optional, where utm coordinates, "
-                "md and tvd are numbers".format(line=line)
+                "md and tvd are numbers"
             ) from err
 
         zone = point[4].strip() if len(point) == 5 else None
--- a/semeio/jobs/rft/trajectory.py
+++ b/semeio/jobs/rft/trajectory.py
@@ -1,4 +1,3 @@
-import argparse
 import os
 
 import pandas as pd
@@ -250,7 +249,7 @@ class Trajectory:
         """
         point = line.split()
         if len(point) < 4 or len(point) > 5:
-            raise argparse.ArgumentTypeError(
+            raise ValueError(
                 (
                     "Trajectory data file not on correct format: "
                     "'utm_x utm_y md tvd <zone>' - zone is optional"
@@ -260,7 +259,7 @@ class Trajectory:
         try:
             floats = [float(v) for v in point[:4]]
         except ValueError as err:
-            raise argparse.ArgumentTypeError(
+            raise ValueError(
                 "Error: Failed to extract data from line {line}. Expected the format "
                 "'utm_x utm_y md tvd zone' - zone is optional, where utm coordinates, "
                 "md and tvd are numbers".format(line=line)
@@ -276,11 +275,7 @@ class Trajectory:
 
         filename = os.path.join(filepath)
         if not os.path.isfile(filename):
-            raise argparse.ArgumentTypeError(
-                "Warning: Trajectory file {filename} not found!".format(
-                    filename=filename
-                )
-            )
+            raise IOError(f"Trajectory file {filename} not found!")
 
         with open(filename, "r") as f:
             trajectory_lines = f.readlines()
--- a/semeio/jobs/scripts/gendata_rft.py
+++ b/semeio/jobs/scripts/gendata_rft.py
@@ -173,14 +173,18 @@ def main_entry_point():
     options = arg_parser.parse_args()
     logger.setLevel(options.log_level)
 
-    well_names = [w_info[0] for w_info in options.well_and_time_file]
+    context_errors = []
+    trajectories = {}
+    for well_name, *_ in options.well_and_time_file:
+        try:
+            trajectories[well_name] = Trajectory.load_from_file(
+                filepath=os.path.join(options.trajectory_path, well_name + ".txt")
+            )
+        except (IOError, ValueError) as err:
+            context_errors.append(str(err))
 
-    trajectories = {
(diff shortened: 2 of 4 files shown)


semeio/jobs/scripts/gendata_rft.py
```diff
@@ -173,14 +173,18 @@ def main_entry_point():
...
-
-    trajectories = {
-        wname: Trajectory.load_from_file(
-            filepath=os.path.join(options.trajectory_path, wname + ".txt")
-        )
-        for wname in well_names
-    }
+    context_errors = []
+    trajectories = {}
+    for well_name in [w_info[0] for w_info in options.well_and_time_file]:
```
lars-petter-hauge, line 178: I have to admit, injecting the list comprehension directly into the for loop expression didn't really look that good.. If you think this is better than having the additional line above, then I guess it's okay. Perhaps it was just that I didn't expect to see it like that..

semeio/jobs/rft/trajectory.py
```diff
@@ -275,9 +275,7 @@ def load_from_file(cls, filepath):
         filename = os.path.join(filepath)
         if not os.path.isfile(filename):
             raise argparse.ArgumentTypeError(
```
lars-petter-hauge, line 277: I think the raising of an argparse error here was a mistake to be honest. Should we just swap out with an IO error or something? And rather catch that in the entry point (which at that point will throw the correct argparse error).

semeio/jobs/scripts/gendata_rft.py
oyvindeide: I am fine either way, did it like this because it was not used for anything else.

semeio/jobs/rft/trajectory.py
oyvindeide: Just to clarify, you would like to keep the entry point as is (except for catching a different error), and change the raised error?

semeio/jobs/scripts/gendata_rft.py
oyvindeide: Think it is more readable now, do you agree?

semeio/jobs/rft/trajectory.py
oyvindeide: _Maybe_ we should actually change the whole configuration and make it more explicit, currently the well file is deduced from a separate file, which is a bit strange.

lars-petter-hauge: I would very much like to change it, we merely used what was already in place to be honest. But then we'll break all setups already in use.

semeio/jobs/scripts/gendata_rft.py
```diff
@@ -173,14 +173,18 @@ def main_entry_point():
...
-
-    trajectories = {
-        wname: Trajectory.load_from_file(
-            filepath=os.path.join(options.trajectory_path, wname + ".txt")
-        )
-        for wname in well_names
-    }
+    context_errors = []
+    trajectories = {}
+    for well_name, *_ in options.well_and_time_file:
```
lars-petter-hauge, line 178: Better! I was wondering if I should suggest `well_name, *_` instead of the `wellinfo[0]`, just reads better. Thanks for changing

semeio/jobs/rft/trajectory.py
lars-petter-hauge: What you did is what I suggested yes

I suppose the ´parse_trajectory_line` would then also need to throw a different error? And that error would need to be caught as well

oyvindeide: Do you have a preference on the error? I used `ValueError`, but can change it, or make a custom if wanted.

semeio/jobs/scripts/gendata_rft.py
```diff
@@ -173,14 +173,18 @@ def main_entry_point():
...
+        try:
+            trajectories[well_name] = Trajectory.load_from_file(
+                filepath=os.path.join(options.trajectory_path, well_name + ".txt")
+            )
+        except (IOError, ValueError) as err:
+            context_errors.append(str(err))
+
+    if context_errors:
+        arg_parser.error("\n".join(context_errors))
```
oyvindeide, line 187: I am not convinced this is the right approach, but seems to be the one you prefer? I would rather just exit with the errors at this point, and not include the `argparse` `usage`, as that is just confusing in my mind, as this is a problem with the input, sure, but how does that `argparse` `usage` help?

lars-petter-hauge: I suppose it was just to have errors on input be consistently handled through argparse (as in that will be what the user sees).

But I don't have that strong opinion really - the main point was more that one shouldn't throw the argparse error outside of the scope of argparse itself. As we are no longer in that scope I'm fine with sys.exit

semeio/jobs/rft/trajectory.py
lars-petter-hauge: `ValueError` should be okay

lars-petter-hauge (approved):
Thanks for fixing!

PR #257, opened 2020-11-12
Add integration test for csv_export2 through mocked ert config

berland:
Exposes bug in libres 7.1.0, MAKE_DIRECTORY is not recognized as a workflow job

The test pass if MAKE_DIRECTORY is taken out.

MAKE_DIRECTORY should be present in the test, because the inline documentation for csv_export uses MAKE_DIRECTORY.

### DIFF
--- a/tests/jobs/csv_export2/test_integration.py
+++ b/tests/jobs/csv_export2/test_integration.py
@@ -1,3 +1,7 @@
+import os
+import sys
+import subprocess
+
 import pytest
 import rstcheck
 import pandas as pd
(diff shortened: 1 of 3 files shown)


berland:
Adding MAKE_DIRECTORY as a standard workflow job was supposedly solved in https://github.com/equinor/ert/pull/1029

tests/jobs/csv_export2/test_integration.py
```diff
@@ -186,3 +189,37 @@ def verifyExportedFile(exported_file_name, result_header, result_iter_rel):
         set(dframe[["ENSEMBLE", "REAL"]].itertuples(index=False, name=None))
         == result_iter_rel
     )
+
+
```
berland, line 193: Should we have `@pytest.mark.integration` here (requires declaring it in `setup.cfg`)?

oyvindeide: That does make sense, we should still default to running all the tests though, as this is a plugin, `ert` will always be available.

berland:
Await a new tag in `libres`, `>7.1.0`

oyvindeide:
Sorry! Missed the request, will have a look tomorrow.

tests/jobs/csv_export2/test_integration.py
```diff
@@ -186,3 +190,85 @@ def verifyExportedFile(exported_file_name, result_header, result_iter_rel):
...
+    ertoutput = subprocess.run(
+        ["ert", "test_run", ert_config_fname], check=True, capture_output=True
+    )
+
+    assert "fmu.ensemble.realization - WARNING - No STATUS file" in str(
+        ertoutput.stdout
+    )
+    assert "realization-2/iter-0" in str(ertoutput.stdout)
+
+    assert os.path.exists("data.csv")
```
oyvindeide, line 273: This is superfluous as trying to read the file on the next line should reveal if the file was not created (and give a good enough message if it fails).

oyvindeide (approved):
LGTM! Just a tiny comment, you can change if you want.

tests/jobs/csv_export2/test_integration.py
berland: Fixed, squashed and rebased.

PR #258, opened 2020-11-13
Fix bug where SUMMARY_OBS were not updating correctly

oyvindeide:


lars-petter-hauge:
Please provide some more context on why this was done in the commit message here - what was the issue up front, why didn't it work.

oyvindeide:
> Please provide some more context on why this was done in the commit message here - what was the issue up front, why didn't it work.

Done, I also added some unit testing to the std updating to make it a bit more transparent. Is the message enough?

tests/jobs/test_scale_observations/test_scale_observations.py
```diff
@@ -0,0 +1,54 @@
...
+
+    obs_vector = snake_oil_obs["FOPR"]
+    for index, node in enumerate(obs_vector):
+        if not index_list or index in index_list:
+            assert node.getStdScaling(index) == 1.2345, f"index: {index}"
+        else:
+            assert node.getStdScaling(index) == 1.0, f"index: {index}"
+
+
+@pytest.mark.parametrize("index_list", [None, [35]])
```
oyvindeide, line 34: The 1 mismatch between the input index list and the step number is a bit concerning. But the consequence of changing it would make the `GEN_OBS` go out of sync. Unless we special case it, but not sure that is a good solution.

lars-petter-hauge: As the behaviour has not changed I believe we can go ahead and merge this.

lars-petter-hauge (approved):
Good to go!

PR #259, opened 2020-11-16
Add pyscal fm model docs from wiki

berland:
Adapted from: https://wiki.equinor.com/wiki/index.php/Pyscal


sondreso (approved):
LGTM and works on my machine :heavy_check_mark:

PR #267, opened 2020-12-02
Update docs

oyvindeide:


lars-petter-hauge:
I agree on most here, but I don't like that unknown items have been modified in the storage data files. We don't know if this will hit us at a later stage, and debugging this could be a nightmare.

How about doing everything but the binary files?

lars-petter-hauge (approved):
LGTM!

PR #272, opened 2020-12-14
Pin segyio and xldr

jondequinor:


oyvindeide (approved):
Can you also create an issue so we remember to unfix them?

PR #280, opened 2020-12-18
Check eclfile

### LINKED ISSUE #203: OTS config segfaulting
If `extract_ots_context` is not mocked the test segfaults with: `test_ots_config.py::test_invalid_config Fatal Python error: Segmentation fault`

Steps to reproduce:

Remove the mocking of `extract_ots_context` (this _should_ only raise a ConfigurationError) and run the test:

https://github.com/equinor/semeio/blob/66f6ae542c79a2c7ad9b6cadb93a2f39fa66cabe/tests/jobs/overburden_timeshift/test_ots_config.py#L42

oyvindeide:
Closes #203

### DIFF
--- a/semeio/jobs/overburden_timeshift/ots.py
+++ b/semeio/jobs/overburden_timeshift/ots.py
@@ -2,6 +2,8 @@ import logging
 from collections import namedtuple
 import os.path
 from datetime import datetime as dt
+from pathlib import Path
+
 import numpy as np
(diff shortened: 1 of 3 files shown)


lars-petter-hauge (approved):
Looks good!

We did some checking on the underlaying issue and created https://github.com/equinor/ecl/issues/786

So we shouldn't have the issue with segfaults at all. Still support the changes done in this PR though, as both better user feedback and testing is very much welcome!

PR #284, opened 2021-01-06
Fix switch from xlrd to openpyxl as engine in pandas

### LINKED ISSUE #273: MIgrate to openpyxl (or at least get rid of xldr<2 requirement)
xldr was pinned to <2 since version 2 refuse to parse other spreadsheet files besides xls files

oyvindeide:
Closes #273

### DIFF
--- a/semeio/jobs/design2params/design2params.py
+++ b/semeio/jobs/design2params/design2params.py
@@ -184,7 +184,8 @@ def _read_excel(file_name, sheet_name, header=0):
     :raises: SystemExit if file not loaded correctly
     """
     try:
-        return pd.read_excel(file_name, sheet_name, header=header, dtype=str)
+        df = pd.read_excel(file_name, sheet_name, header=header, dtype=str)
+        return df.dropna(axis=1, how="all")
     except IOError as err:
         raise SystemExit("File {} not found".format(file_name)) from err
     except Exception as err:


hnformentin:
Hello! A curiosity: would not be useful to report users that columns with NAs would be dropped? Or xldr had this same behaviour?

oyvindeide:
With this we drop support of `xls`-files. I think that is fine, and discussing with @berland the input to this job is mostly generated by https://github.com/equinor/fmu-tools/blob/db5faad96cd4500b31801fbcc34c4f41d8ee018f/src/fmu/tools/sensitivities/fmudesignrunner.py#L103, which will create `xlsx` files. I can add a check for extension, but `xls` was last used in Excel 2003 so seems redundant.

oyvindeide:
> Hello! A curiosity: would not be useful to report users that columns with NAs would be dropped? Or xldr had this same behaviour?

This was the old behavior, which is why I added it. Columns with only spaces in them would not be read, they are in `openpyxl`. We have a test ensuring that columns with only spaces are dropped, so it is desired behavior.

hnformentin:
Thanks. That's good, the `how=all` specifies only columns where all are NA! Great!

hnformentin (approved):
LGTM!

PR #286, opened 2021-01-11
Add kmeans clustering function

### LINKED ISSUE #241: [MPP] Running slow with large number of observations
We are testing the job with a large number of realizations (~30000s), but only a low number of observations (100s). It has been reported that the MisfitPreprocessor is slow with a large number of observations (https://github.com/equinor/flownet/pull/217) 

We should:
- [ ] Create a test case that has a large number of observations and a large number of realizations to test this
- [ ] I suspect the culprit might be the exporting of data from `ert`, we should `profile` some of the integration tests to see
where the bottleneck is.

oyvindeide:
Closes #241

### DIFF
--- a/semeio/workflows/misfit_preprocessor/job.py
+++ b/semeio/workflows/misfit_preprocessor/job.py
@@ -4,30 +4,23 @@ from semeio.workflows.spearman_correlation_job.job import spearman_job
 
 def run(config, measured_data, reporter):
     workflow = config.workflow
-    if workflow.type == "custom_scale":
-        sconfig = workflow.clustering.cluster_args()
-        scaling_configs = spearman_job(
-            measured_data,
-            reporter,
-            **sconfig,
-        )
-        pca_threshold = config.workflow.pca.threshold
-    elif workflow.type == "auto_scale":
+    sconfig = workflow.clustering.cluster_args()
+    if workflow.type == "auto_scale":
         job = ObservationScaleFactor(reporter, measured_data)
         nr_components, _ = job.perform_pca(workflow.pca.threshold)
-        sconfig = workflow.clustering.cluster_args()
-        scaling_configs = spearman_job(
-            measured_data,
-            reporter,
-            criterion="maxclust",
-            threshold=nr_components,
-            **sconfig,
-        )
-        pca_threshold = workflow.pca.threshold
-    else:
-        raise AssertionError(
-            "Unknown clustering method: {}".format(config.workflow.method)
-        )
+        if workflow.clustering.type == "limited_hierarchical":
+            sconfig["criterion"] = "maxclust"
+            sconfig["threshold"] = nr_components
+        elif workflow.clustering.type == "limited_kmeans":
+            sconfig["n_clusters"] = nr_components
+
+    scaling_configs = spearman_job(
+        measured_data,
+        reporter,
+        cluster_function=workflow.clustering._cluster_function,
+        **sconfig,
+    )
+    pca_threshold = workflow.pca.threshold
 
     scaling_params = {"threshold": pca_threshold}
     for scaling_config in scaling_configs:
--- a/semeio/workflows/misfit_preprocessor/kmeans_config.py
+++ b/semeio/workflows/misfit_preprocessor/kmeans_config.py
@@ -1,0 +1,62 @@
+try:
+    from typing import Literal
+except ImportError:
+    from typing_extensions import Literal
+from pydantic import conint, Field, PyObject, PrivateAttr
+
+from semeio.workflows.misfit_preprocessor.hierarchical_config import (
+    AbstractClusteringConfig,
+)
+from semeio.workflows.spearman_correlation_job.cluster_analysis import (
+    kmeans_analysis,
+)
+
+
+class LimitedKmeansClustering(AbstractClusteringConfig):
+    """
+    The kmeans implementation is backed by sklearn and for a
+    more detailed description we refer the reader to the
+    documentation of sklearn.cluster.KMeans
+    (https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html).
+    """
+
+    type: Literal["limited_kmeans"] = "limited_kmeans"
+    init: Literal["k-means++", "random"] = Field(
+        "k-means++",
+        description=("The criterion to use in forming flat clusters. "),
+    )
+    n_init: conint(gt=0, strict=True) = Field(
+        10,
+        description=(
+            "Number of time the k-means algorithm will be run with "
+            "different centroid seeds. The final results will be the "
+            "best output of n_init consecutive runs in terms of inertia."
+        ),
+    )
+    max_iter: conint(gt=0, strict=True) = Field(
+        300,
+        description=(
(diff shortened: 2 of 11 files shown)


tests/jobs/misfit_preprocessor/test_misfit_preprocessor.py
```diff
@@ -147,16 +147,28 @@ def parameter_distribution():
...
 
 
+@pytest.mark.parametrize("clustering_function", ["hierarchical", "kmeans"])
 @pytest.mark.parametrize("method", ["spearman_correlation", "auto_scale"])
 @pytest.mark.parametrize(
     "num_polynomials",
     tuple(range(1, 5)) + (20, 100),
 )
-def test_misfit_preprocessor_n_polynomials(num_polynomials, method):
+def test_misfit_preprocessor_n_polynomials(
```
eivindjahren, line 156:
```
from hypothesis import given, assume
import hypothesis.strategies as st

clustering_functions = st.sampled_from(["hierarchical", "kmeans"])
methods = st.sampled_from(["spearman_correlation", "auto_scale"])

@given(st.integers(min_value=1, max_value=100), methods, clustering_functions)
def test_misfit_preprocessor_n_polynomials(
   num_polynomials, method, clustering_function
):
    if (
        clustering_function == "kmeans"
        and method == "spearman_correlation"
     ):
       assume(num_polynomials in [4,5]) # or more

File: semeio/workflows/misfit_preprocessor/job.py [hunk 2]
```diff
@@ -18,7 +17,7 @@ def run(config, measured_data, reporter):
     scaling_configs = spearman_job(
         measured_data,
         reporter,
-        cluster_function=workflow.clustering.type,
+        cluster_function=workflow.clustering._cluster_function,
```
oyvindeide, line 20: Not entirely sure how I feel about this, but think I might like it, it keeps what is being configured close to the configuration. But it could be over complicating things.

semeio/workflows/misfit_preprocessor/job.py
```diff
@@ -4,30 +4,23 @@
...
-            threshold=nr_components,
-            **sconfig,
-        )
-        pca_threshold = workflow.pca.threshold
-    else:
-        raise AssertionError(
-            "Unknown clustering method: {}".format(config.workflow.method)
-        )
+        if workflow.clustering.type == "limited_hierarchical":
+            sconfig["criterion"] = "maxclust"
```
lars-petter-hauge, line 12: I am still against modifying the configuration, even if it is set to `None` before we get to this place.

I would want the `sconfig` object to be immutable. A possible implementation would then be to add a `job_kwargs = {}`, update that instead of the `sconfig` and pass it to the function along with `sconfig`: `spearman_job(**sconfig, **job_kwargs)`

semeio/workflows/misfit_preprocessor/job.py
lars-petter-hauge: I don't think it overcomplicates things by passing the function as part of the configuration. Better than having some sort of mapping somewhere  which maps string to func. I think we should be consistent though, and if there are other places which would map to a function, we could likely have those places use the same behaviour.

semeio/workflows/misfit_preprocessor/kmeans_config.py
```diff
@@ -0,0 +1,62 @@
...
+        ),
+    )
+    random_state: conint(gt=0, strict=True) = Field(
+        None,
+        description=(
+            "Determines random number generation for centroid initialization. "
+            "Use an int to make the randomness deterministic"
+        ),
+    )
+    _cluster_function: PyObject = PrivateAttr(kmeans_analysis)
```
lars-petter-hauge, line 49: I have to admit, I look at this and start thinking about `PyBind11` and `PyObject` definitions. But we didn't write Pedantic..

tests/jobs/misfit_preprocessor/unit/test_config.py
```diff
@@ -327,3 +328,21 @@ def test_config_workflow_valid(workflow):
...
+    [
+        {},
+        {"init": "random"},
+        {"init": "k-means++"},
+        {"n_init": 1},
+        {"max_iter": 10},
+        {"random_state": 10},
+    ],
+)
+def test_kmeans_schema(workflow, clustering_config):
```
lars-petter-hauge, line 345: Just testing that what you set is what you get? And that they make valid configs? Just to get a bit more out of the test, you could verify that the rest are kept as defaults. Not a game changer, but simply a suggestion.

Could also test some invalid configs, for instance several of these are specified to be constrained types, verify that the constraints are correct?

tests/jobs/misfit_preprocessor/unit/test_config.py
```diff
@@ -327,3 +328,21 @@ def test_config_workflow_valid(workflow):
...
+    config = KmeansClustering(**clustering_config)
+    for key, val in clustering_config.items():
+        assert config.dict()[key] == val
```
lars-petter-hauge, line 348: Not a biggy, but the creation of a dict could be performed outside of the loop.

semeio/workflows/spearman_correlation_job/job.py
```diff
@@ -1,12 +1,25 @@
...
 import logging
-from scipy.cluster.hierarchy import fcluster, linkage
+
+from semeio.workflows.spearman_correlation_job.cluster_analysis import (
+    fcluster_analysis,
+    kmeans_analysis,
+)
+
+
+_cluster_map = {
```
lars-petter-hauge, line 11: Where is this used?

semeio/workflows/misfit_preprocessor/job.py
oyvindeide: But we are not modifying the configuration though, we are adding things that are not present in the configuration.

lars-petter-hauge: even if ~it is set to None~ it does not exist before we get to this place.

lars-petter-hauge:
Had some minor comments, otherwise I think the implementation looks good!

semeio/workflows/misfit_preprocessor/job.py
oyvindeide: But is that kind of not what we are doing? It is just a dict (`sconfig`), a copy of the properties of the scaling function, the actual configuration object (`workflow`) is not being modified in any way.

tests/jobs/misfit_preprocessor/unit/test_config.py
oyvindeide: I can add a couple of other tests for what you suggest, I would like to keep this test as is to avoid testing too many things at once.

semeio/workflows/misfit_preprocessor/job.py
lars-petter-hauge: I didn't look properly where the `sconfig` came from (....`workflow.cluster.cluster_args()`). I thought we were changing the instance coming from pydantic. I agree that passing just the dicts makes for a better interface.

lars-petter-hauge (approved):
LGTM! Nice job!

PR #287, opened 2021-01-12
Add linkage

### LINKED ISSUE #285: Add a level after config -> clustering workflow -> hierarchical
It should enable adding more configurations within hierarchical clustering method.
Initially add f_cluster and linkage in the config.
Follows https://github.com/equinor/semeio/pull/268.

oyvindeide:
Closes #285


lars-petter-hauge:
Although we're adding an additional level of configuration here, it will be more explicit. Considering that most users will run the workflow with default settings, and the configuration details we are modifying here are mainly for expert users, I think we can allow "advanced settings" to be more verbose.

lars-petter-hauge (approved):

PR #288, opened 2021-01-12
Let design2param fail hard if design==defaults

### LINKED ISSUE #283: [design2params] Error when defaultsheet and designsheet is the same
**Is your feature request related to a problem? Please describe.**
If a user erroneously sets default-sheet and designsheet to the same value, a strange Pandas error can/will occur that
the user finds hard to understand:

`Invalid value in design matrix header, error: Location based indexing can only have [labels (MUST BE IN THE INDEX), slices of labels (BOTH endpoints included! Can be slices of integers if the index is integers), listlike of labels, boolean] types`

**Describe the solution you'd like**
Testing the sheetparameters for equivalence and error with clear message.

berland:
Allowing designsheetname==defaultsheetname can produce quite a number of arbitrary wrong results or error messages (if lucky) as the sheets are attempted interpreted with different assumed structure. Avoid this pitfall by erroring hard on equal sheetname input.

Resolves #283


oysteoh (approved):
🚀

PR #291, opened 2021-01-19
Use pydantic

oyvindeide:


semeio/workflows/misfit_preprocessor/hierarchical_config.py
```diff
@@ -105,18 +106,28 @@ class LinkageConfig(BaseMisfitPreprocessorConfig):
...
+    for key, value in dictionary.items():
+        if isinstance(value, collections.MutableMapping):
+            items.extend(_flatten(value).items())
+        else:
+            items.append((key, value))
+    return dict(items)
+
+
+class AbstractClusteringConfig(BaseMisfitPreprocessorConfig):
+    def cluster_args(self):
```
oyvindeide, line 120: Not sure this is the best approach, but the goal is for the configuration to be able to provide the arguments to the clustering function when we add a new function. In the case of `hierarchical` the configuration is split into `linkage` and `fcluster`, but the function we are configuring is not so the return `dict` is flattened. It could be argued that either the function or the configuration should change, but I kind of like it as it sits.

semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
```diff
@@ -76,7 +74,5 @@ def _get_observations(facade):
 @hook_implementation
 def legacy_ertscript_workflow(config):
     workflow = config.add_workflow(MisfitPreprocessorJob, "MISFIT_PREPROCESSOR")
-    _schema = misfit_preprocessor.config._SCHEMA
-    rst_doc = configsuite.docs.generate(_schema)
-    workflow.description = rst_doc
+    workflow.description = "No description as of now"
```
eivindjahren, line 77: Maybe use pydantic schema: https://pydantic-docs.helpmanual.io/usage/schema/

semeio/workflows/spearman_correlation_job/job.py
```diff
@@ -99,6 +93,13 @@ def _calculate_correlation_matrix(data):
     return data.rank().corr(method="pearson")
 
 
-def _cluster_analysis(correlation_matrix, threshold, criterion, depth, method, metric):
+def _cluster_analysis(
```
eivindjahren, line 96: Maybe best to just inline it and avoid `**cluster_args`.

eivindjahren (changes requested):

semeio/workflows/spearman_correlation_job/job.py
oyvindeide: In this specific case I would tend to agree, the reason I have chosen this approach is because the next PR will introduce a second clustering function, with completely different arguments which will need to be passed to this function. Am happy to take suggestions though.

semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
oyvindeide: Yes, though I would like to avoid creating my own `rst` from the schema. Have you found a good way of doing that?

eivindjahren: Maybe something like https://pypi.org/project/sphinx-jsonschema/ ?

semeio/workflows/spearman_correlation_job/job.py
eivindjahren: Alright then we will see how it looks later :)

semeio/workflows/misfit_preprocessor/hierarchical_config.py
eivindjahren: Seems like a reasonable approach, maybe

```
def cluster_args(config_object):
    return _flatten(config_object.dict())
```

semeio/workflows/misfit_preprocessor/misfit_preprocessor.py
oyvindeide: The problem with most of these packages is that they are sphinx plugins, while what we are doing in this case is passing an `rst` string to `ert` for it to create the actual documentation.

eivindjahren: Could the consumer of the documentation be required to understand the sphinx extension?

eivindjahren: There is also https://github.com/inspirehep/jsonschema2rst which is not well maintained but does the job.

oyvindeide: It looks like it could do the job, though the non-cli entry point is not documented.

eivindjahren: Then, I guess your solution is the best we can do for now.

eivindjahren (approved):

PR #294, opened 2021-02-11
Use scikit-learn instead of sklearn

oyvindeide:


ManInFez (approved):
Remember to add this to the dependencies in komodo-releases/repository :)

PR #296, opened 2021-02-16
Change output folder from config name to case name

### LINKED ISSUE #210: [Reporter] Output folder has config name - should perhaps have case name

oyvindeide:
Closes #210, though the issue is phrased as a question, so should consider other options.


lars-petter-hauge (approved):
Looks good!

PR #301, opened 2021-02-26
Move reporter output to output path

### LINKED ISSUE #211: [Reporter] The reporter output is located next to the config file, should perhaps be in the output folder

oyvindeide:
Closes #211

### DIFF
--- a/tests/jobs/correlated_observations_scaling/test_integration.py
+++ b/tests/jobs/correlated_observations_scaling/test_integration.py
@@ -134,7 +134,14 @@ def test_main_entry_point_gen_data(setup_ert):
         np.sqrt(3.0 / 2.0),
     )
 
-    svd_file = "reports/default_0/CorrelatedObservationsScalingJob/svd.json"
+    svd_file = os.path.join(
+        "storage",  # ens_path == storage/snake_oil/ensemble
+        "snake_oil",
+        "reports",
+        "default_0",
+        "CorrelatedObservationsScalingJob",
+        "svd.json",
+    )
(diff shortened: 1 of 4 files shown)


tests/jobs/correlated_observations_scaling/test_integration.py
```diff
@@ -134,7 +134,14 @@ def test_main_entry_point_gen_data(setup_ert):
         np.sqrt(3.0 / 2.0),
     )
 
-    svd_file = "reports/default_0/CorrelatedObservationsScalingJob/svd.json"
+    svd_file = os.path.join(
```
lars-petter-hauge, line 137: The path here is really confusing, because what this PR intends to do is to put the output of `Reporter` _next to_ the internal storage (by default `storage`) folder, while this test seems to assert that it is within the storage directory - pretty much what was prevented in https://github.com/equinor/semeio/pull/169

The reason is that the configuration of snake_oil sets the `ENSPATH` to `storage/<CONFIG_FILE_BASE>/ensemble`, hence `storage/snake_oil/ensemble` now contains ERTs internal storage (which by default is named `storage`). Therefore the `reports` folder will be next to the internal storage as intended.

So although the test looks correct, could you add a comment to detail why this is the path we are asserting?

oyvindeide: Sure. I have added a comment, though this is a bit incidental as this function is not really testing the reporter functionality, just a side effect of it.

lars-petter-hauge (approved):

PR #302, opened 2021-02-26
Dump full stea response

### LINKED ISSUE #27: Dump the stea response json
Today the stea job dumps the response json in a peculiar format tailored to ERT's dataformat. We should instead dump the richer format and let the user further process that to their needs.

I think this has been requested previously by @berland

oyvindeide:
Closes #27

### DIFF
--- a/semeio/jobs/scripts/fm_stea.py
+++ b/semeio/jobs/scripts/fm_stea.py
@@ -1,3 +1,4 @@
+import json
 import stea
 import argparse
 from semeio import valid_file
@@ -25,6 +26,12 @@ def _get_args_parser():
         help="STEA config file, yaml format required",
         type=valid_file,
     )
+    parser.add_argument(
+        "-r",
+        "--response_file",
+        help="STEA response, json format",
+        default="stea_response.json",
+    )
     return parser
 
 
@@ -32,7 +39,28 @@ def main_entry_point():
     parser = _get_args_parser()
     options = parser.parse_args()
     stea_input = stea.SteaInput([options.config])
-    res = stea.calculate(stea_input)
-    for res, value in res.results(stea.SteaKeys.CORPORATE).items():
+    result = stea.calculate(stea_input)
+    for res, value in result.results(stea.SteaKeys.CORPORATE).items():
         with open("{}_0".format(res), "w") as ofh:
             ofh.write("{}\n".format(value))
+    profiles = _get_profiles(
+        stea_input.stea_server,
+        stea_input.project_id,
+        stea_input.project_version,
+        stea_input.config_date,
+    )
+    full_response = _build_full_response(
+        result.data[stea.SteaKeys.KEY_VALUES], profiles
+    )
+    with open(options.response_file, "w") as fout:
+        json.dump(full_response, fout, indent=4)
+
+
+def _get_profiles(server, project_id, project_version, config_date):
+    client = stea.SteaClient(server)
+    project = client.get_project(project_id, project_version, config_date)
+    return project.profiles
+
+
+def _build_full_response(result, profiles):
+    return {"response": result, "profiles": profiles}
--- a/tests/jobs/stea/test_stea.py
+++ b/tests/jobs/stea/test_stea.py
@@ -1,10 +1,12 @@
+import json
 import sys
-from unittest.mock import patch
+from unittest import mock
 import pytest
-from semeio.jobs.scripts.fm_stea import main_entry_point
+from semeio.jobs.scripts import fm_stea
 import shutil
 import os
 from stea import SteaResult, SteaKeys
+import stea
 
 
 TEST_STEA_PATH, _ = os.path.split(os.path.abspath(__file__))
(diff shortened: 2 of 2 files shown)


oyvindeide:
@berland you know more about this than me, does this solve the issue? Would we like the option of naming the full response?

berland:
For reference, this is a forward model that  I could dig up from 2019, probably written while adding issues github/stea:  
```python
#!/usr/bin/env python
import sys
import stea
import json
import pprint

STEA_PRETAX_OUTPUT = "stea_pretax_results.json"
STEA_CORPORATE_OUTPUT = "stea_corporate_results.json"
STEA_PROFILES = "stea_profiles.json"
PARAMETERSTXT = "parameters.txt"

def main(argv):
    """
    Forward model script for STEA-FMU

    In ert-config:
    -------
    INSTALL_JOB STEA_FMU ../bin/jobs/FM_STEA_FMU
    ...
    FORWARD_MODEL STEA_FMU(<CONFIG>=<CONFIG_PATH>/../input/config/steaconfig.yml)
    -------
    In bin/jobs/FM_STEA_FMU:
    ------
    STDERR stea_fmu.stderr
    STDOUT stea_fmu.stdout

    TARGET_FILE stea_corporate_results.json

    PORTABLE_EXE ../scripts/stea_fmu.py
    ARGLIST   "--ecl_case" <ECLBASE> <CONFIG>
    -------
    """
    # Parse the YAML input file and command line arguments
    stea_input = stea.SteaInput(sys.argv[1:])  # Refactor this later, we should NOT pass
                                               # around sys.argv!
    # Do calculation
    res = stea.calculate(stea_input)

    with open(STEA_PRETAX_OUTPUT, "w") as pretax_fh:
        pretax_fh.write(json.dumps(res.results(stea.SteaKeys.PRETAX)))
    with open(STEA_CORPORATE_OUTPUT, "w") as corp_fh:
        corp_fh.write(json.dumps(res.results(stea.SteaKeys.CORPORATE)))

    # Obtain some information from the STEA server for debugging purposes etc.
    # (this code is taken from the stea/__init__.py called above through
    # stea.calculate, so this should be refactored)
    client = stea.SteaClient(stea_input.stea_server)
    project = client.get_project(
        stea_input.project_id, stea_input.project_version, stea_input.config_date
    )
    pp = pprint.PrettyPrinter(indent=2)
    with open(STEA_PROFILES, "w") as profiles_fh:
        profiles_fh.write(pp.pformat(project.profiles))
        profiles_fh.write("\n")  # Newline at end of file

    # Write to individual files (for GEN_DATA)
    for res_key, value in res.results(stea.SteaKeys.CORPORATE).items():
        with open("%s_0" % (res_key), "w") as ofh:
            ofh.write("%f\n" % (value))

    # Append to parameters.txt (for CSV_EXPORT1)
    for res_key, value in res.results(stea.SteaKeys.CORPORATE).items():
        with open(PARAMETERSTXT, "a") as params_fh:
            params_fh.write("%s %f\n" % (res_key, value))


if __name__ == "__main__":
    main(sys.argv)

```

berland:
From this, it seems that pretax is also interesting to have the dumped json for.

There  was  probably a good reason why I added the dumping of profiles to json  as well.

oyvindeide:
In what I set up now it will dump:
```
[
    {
        "TaxMode": "Corporate",
        "Values": {
            "NPV": 30
        }
    }
]
```
but I dont have that many real examples to test against, even in the `stea` repo there is not that much to go with.

oyvindeide:
Do we know anyone using this we can ask what they would like in the response?

berland:
I think  we should dump the three json files from the code above, instead of waiting for someone to request it. They can probably be merged to  one json file, though, indexed accordingly.

On-prem, this can be tested using a yml file like
```yaml
project-id: 56892  # FMU-dummy
project-version: 1
  # Date to base prices and assumptions on
config-date: 2019-01-01 00:00:00
  # # Profile data to replace into STEA project
ecl-profiles:
# Comment attached to profile or unique identifier from database
#  69259963-a52a-4505-8358-1a50e6b890a0:
  FOPT:
  # Eclipse summary key to be used for the profile in STEA
    ecl-key: FOPT
    # Eclipse case to extract summary based on
ecl-case: model/2_R001_REEK
  # Results to request from STEA
results:
  - IRR
  - CEI
  - NPV
```
By passing `--ecl_case`, we can submit the production profile for the Norne case in semeio test-data, and results from the STEA server will be returned (on-prem).

The profiles export is needed because it lists UUIDS which are hard to get by other means.

Internal docs on https://wiki.equinor.com/wiki/index.php/FMU_tutorial/STEA

oyvindeide:
Thanks! Will do some testing on prem then.

oyvindeide:
What is the reason behind splitting on `TaxMode` and only outputting a subset? I would suggest we just dump the full response. We could do a json structure with: 
```
"response": [{"TaxMode": "Corporate", "Rate": 0.0}, {"TaxMode": "Special", "Rate": 0.0} ....]
"profiles": {"UUID_1": {"Id": "UUID_1"}....}
```

berland:
Not sure why the files was splitted on TaxMode, support merging them.

semeio/jobs/scripts/fm_stea.py
```diff
@@ -25,14 +26,34 @@ def _get_args_parser():
...
+
+def _build_full_response(stea_input, result):
+    full_response = {}
+    client = stea.SteaClient(stea_input.stea_server)
+    project = client.get_project(
+        stea_input.project_id, stea_input.project_version, stea_input.config_date
+    )
+    full_response["response"] = result.data[stea.SteaKeys.KEY_VALUES]
+    full_response["profiles"] = project.profiles
+    return full_response
```
lars-petter-hauge, line 59: I find this function a bit weird.. It both takes some already generated response, and also creates a new client to get more information - and then merges. I think it would make sense to separate the functionality of returning the profile response in a single function, and have another that formats the result to something that can be dumped to a json.

oyvindeide: I dont disagree, but we would have to refactor `STEA` in order to achieve that, and dont really think that is within the scope.

lars-petter-hauge: Thanks for splitting the functions. Yeah, we can't really do more now without starting to change things in the stea library, and I agree that is not within the scope.

Looks better though!

lars-petter-hauge (approved):
Looks good!

PR #303, opened 2021-03-01
Add OUTPUTDIRECTORY to doc example, gendata_rft

berland:
Some minor addition/changes to the rest of the doc


berland:
@rnyb, the addition of "gendata_rft" (OUTPUTDIRECTORY) to the GENDATA example here was based on my hunch,  not  tested.  Does it look correct?

rnyb:
I think i will need to run a test on it, but from first glance it looks good

berland:
Tested now in #304 (testing revealed that MAKE_DIRECTORY should be added to this doc example).

Should be ready to be merged.

oyvindeide (approved):

PR #304, opened 2021-03-02
Add test function for gendata_rft ert integration

berland:
First step towards having a proven  reference example for GENDATA_RFT documentation.

There should possibly be some more asserts that ERT is able to internalize, but I don't know how.

Next step is to try more advanced setups with RFT's at different wellpoints at different dates. Ping @rnyb

### DIFF
--- a/tests/jobs/rft/test_gendata_rft.py
+++ b/tests/jobs/rft/test_gendata_rft.py
@@ -1,14 +1,15 @@
 # pylint: disable=unsubscriptable-object  # pylint issue
-import sys
 import os
+import shutil
+import subprocess
+import sys
 from pathlib import Path
 
-import pandas as pd
 import numpy
-
+import pandas as pd
 import pytest
 
-from semeio.jobs.scripts.gendata_rft import main_entry_point, _build_parser
+from semeio.jobs.scripts.gendata_rft import _build_parser, main_entry_point
 from tests.jobs.rft import conftest
 
 ECL_BASE_NORNE = conftest.get_ecl_base_norne()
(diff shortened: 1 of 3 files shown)


berland:
Any known reasons it fails on MacOS? @lars-petter-hauge

lars-petter-hauge:
Typical issue with OSX is that the file system is case insensitive, and we have the job configuration named the same as the executable. The executable should be in `bin` though, and not located next to the job configuration file. I think `libres` will check for `EXECUTABLE` value relative to the job configuration before it checks path, which is why we end up with this problem.

lars-petter-hauge:
A quick comment on this challenge though.

We can solve this issue by having different names on the executable and the job configuration file. The change that has the least effect on other users would be to rename the job configuration file. The `INSTALL_JOB job_name config_name` allows for different names on those two, so we can keep both the name of the job for users in ert config e.g. `INSTALL_JOB GENDATA_RFT GENDATA_RFT_CONFIG`, and at the same time keep `gendata_rft` the same on the command line. The Hook manager in ert allows for the users to provide both of the argument values for `INSTALL_JOB`. So if we change:

https://github.com/equinor/semeio/blob/3990ee2a2110f89328eac5077998dffac8094f9b/semeio/hook_implementations/jobs.py#L17

To return e.g. `GENDATA_RFT: GENDATA_RFT_CONFIG` instead, we'll have this solved.

berland:
Ok. Integration tests like this are proven valuable in ecl2df and subscript, which probably would face the same issue if their tests was run on OSX.

oyvindeide:
Do you want to attempt a fix in this PR, or should we create an issue?

berland:
> Do you want to attempt a fix in this PR, or should we create an issue?

Do you mean issue #306?

oyvindeide:
> > Do you want to attempt a fix in this PR, or should we create an issue?
> 
> Do you mean issue #306?

Yes I do!

berland:
> > > Do you want to attempt a fix in this PR, or should we create an issue?
> > Do you mean issue #306?
> Yes I do! A little to quick there...

Then the answer is that I hope someone with access to a Mac will fix it.

oyvindeide:
I poked @lars-petter-hauge to implement his suggestion, in the mean time feel free to add the `pytest.skip` workaround.

berland:
Workaround added.

lars-petter-hauge:
Sorry for this not having been reviewed.

Could I ask for the reasoning behind adding the integration test? Preferably, if we can avoid invoking `ert` as part of our pipeline, we probably should. Not because I'm against integrationtests, but because `ert` takes so long to setup. Unless I'm reading this wrong, these integrationtests added a minute or so on the test setup. Integation tests for all our forward model jobs would then by far take the most of our test time.

How would this test bring additional value over asserting the output that the `GENDATA_RFT` produce? If it isn't compatible with what ert accepts, then it's pretty much a bug in our testcase. (unless ert has broken it's interface..)

berland:
It would probably be ok to not run these tests always, i.e. maybe activated through an option to  `pytest`. 

But for ensuring that we understand how to use this tool in different modelling circumstances, and ensuring that the job documentation is correct, these tests are gold.

berland:
PR is changed to have these tests (2 x 30 sec when run on local iron) only run when an explicit option `--ert_integration` is supplied to `pytest`.

It could be considered to add this option to the CI pipeline, Jenkins and/or Github actions?

tests/jobs/rft/test_gendata_rft.py
```diff
@@ -425,3 +426,222 @@ def test_defaults():
...
+
+
+@pytest.mark.ert_integration
+@pytest.mark.skipif(
+    sys.platform == "darwin",
+    reason=(
+        "Forward models have same name as executable and "
+        "libres will find wrong exe on case insensitive system "
+        "bug report here: https://github.com/equinor/libres/issues/1053"
+    ),
```
lars-petter-hauge, line 438: Can we not remove this then? As we have fixed the issue of case sensitivity

berland: Done.

oyvindeide:
@lars-petter-hauge do you recon this is ready to merge? We should add the option to run the test in at least one of the CI pipelines, perhaps `Jenkins` would be the best?

lars-petter-hauge:
Yes,  I think having the integration tests running within Jenkins makes sense, as they will also be run for bundled environments. Could you please update the https://github.com/equinor/semeio/blob/main/ci/jenkins/testkomodo.sh script @berland ?

lars-petter-hauge (approved):

Issue #308, opened 2021-03-17
Interactive usage of csv_export2


berland:
**Allow interactive usage of csv_export2 outside ERT**
For interactive usage outside of ERT, a command line API to `csv_export2`  can be a good idea

**Describe the solution you'd like**
csv_export2 already has a command line API, but it requires an existing RUNPATHFILE which is non-trivial to create ad-hoc. Suggest to allow a glob-path to be supplied instead of RUNPATHFILE, so that if csv_export2 encounters a non-existing runpath file, it will attempt to interpret it as a glob-path that can be supplied to `EnsembleSet`s initialization code as a `frompath` argument.

markusdregi:
This will move the export further away from the current direction of ERT development and no shared disk for data communication. I'm suggesting that this command line tool is instead implemented closer to the `EnsembleSet`, maybe in `fmu-ensemble`?

berland:
Fair point, an endpoint for this has always been relevant to make for `fmu-ensemble`.  `csv_export2.py` is the path of least resistance, the diff to support this starts with only  three extra lines, and turns out to be handy for debugging/testing.

markusdregi:
Could you please elaborate on why the situation arise when you do not have a runpath file?

berland:
> Could you please elaborate on why the situation arise when you do not have a runpath file?

* Yesterday I needed it (and used this PR) to test if https://github.com/equinor/fmu-ensemble/pull/206 gave any speedup on `csv_export2.py` on real-life ensembles from `/scratch/`. Digging up the associated runpathfiles for a random directory on `/scratch/` is a no-go. 

* If a user requests support, but in the first pass only provide the scratch directory, this PR will make it easier to generate a CSV-file quickly for simple tests/analysis/debugging.  When a user needs support on csv exporting,  it is also possible that the problem is in the runpathfile, so a possibility to skip that is handy, if not to solve the problem, but at least isolate it.

* For situations where the user has rerun only a failed part of an ensemble, there are situations where the runpathfille becomes incorrect (or at least not what you want), and the ability to work around using the command line can be handy.

This functionality is *not* intended to be used in ERT-config, so it is not documented as such.

markusdregi:
Those are good user stories that I understand that you are motivated to cater for! Still, `semeio` is exactly for functionality that **is** intended to be used within ERT setups. Hence, I still believe this tool would fit better within `fmu-ensemble`, or possibly `subscript`...

berland:
[...]

PR #309, opened 2021-03-18
Let csv_export2 work without runpathfile

berland:
Makes csv_export2 a handy debugging tool, skipping the need
to generate a runpathfile. Relevant in failed scenarios etc.

More robustness can be coded if this PR will be accepted, i.e. if user forgets to quote `"realization-*/iter-*"` the current PR will error.


berland:
See discussion in #308

PR #311, opened 2021-03-18
wip: Adding ert integration test for STEA

berland:
This test function runs on-prem.

PR #316, opened 2021-04-13
Add ecl_case command line argument

### LINKED ISSUE #310: STEA forward model & ecl-case
The STEA forward model doesn't pick up the realisation number for the ECLBASE in the ecl-case automatically. Currently a workaround symlinking the ECLBASE to a general name, has to be used. 
image
image
image

oyvindeide:
Closes #310


oyvindeide:
I dont think this is the best solution possible, but it is already exposed in the library we are wrapping, so in that regard it makes sense to include it.

lars-petter-hauge (approved):
I don't have any strong opinions on the naming of the argument, there are pro's and con's both ways.

@berland if you have any objections?

berland:
Probably good. Can maybe #311 be integrated into this PR to verify it?

oyvindeide:
> Probably good. Can maybe #311 be integrated into this PR to verify it?

Sure, no problem for me, so basically this:

```
RUNPATH realization-%d/iter-%d
ECLBASE eclipse/model/NORNE_ATW2013-0
QUEUE_SYSTEM LOCAL
NUM_REALIZATIONS 1
FORWARD_MODEL STEA(<CONFIG>=stea_conf.yml,<ECL_CASE>=<ECLBASE>)
```
?

lars-petter-hauge (approved):
Nice job on the integrationtest!

oyvindeide:
@berland I kind of hijacked your PR, do you agree with my changes? I kept you as the author.

berland:
Looks good!

PR #317, opened 2021-04-15
Add init files in test folders

oyvindeide:
There was a mix of folders with and without init files in the test subdirectory, this adds to the missing for consistency.

### DIFF
--- a/tests/jobs/overburden_timeshift/test_ots.py
+++ b/tests/jobs/overburden_timeshift/test_ots.py
@@ -8,7 +8,7 @@ from ecl.util.geometry import Surface
 import segyio
 
 from semeio.jobs.overburden_timeshift.ots import OverburdenTimeshift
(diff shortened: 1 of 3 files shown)


lars-petter-hauge:
quick question, wouldn't pytest test discovery fail for any subsequent folders when `__init__.py` is missing?

oyvindeide:
No, not really, pytest will allow both, but there could be namespace problems if the test files are called the same in multiple folders without `__init__.py`

lars-petter-hauge:
This has been an issue with `UnitTest`' discovery at least: https://bugs.python.org/issue23882

>Whenever this method finds a subdirectory of start_dir it checks for `__init__.py` file. If there's no `__init__.py` then finder assumes that files within this package is not importable and stops recursion.

Anywho, I would agree with the change

lars-petter-hauge (approved):

PR #319, opened 2021-04-27
Fix minor documentation issues ahmanalysis

lars-petter-hauge:


eivindjahren (approved):

oyvindeide (changes requested):
I object to the commit message, otherwise fine.

lars-petter-hauge:
> I object to the commit message, otherwise fine.

Better?

oyvindeide (approved):
Better, but capital Examples? ;)

lars-petter-hauge:
I expect the same level of precision in the review for any PR coming in ;)

PR #320, opened 2021-04-29
Correct ahm_analysis workflow description

mareco701:
This is to correct the ahm_analysis workflow description for the output path of the created csv files.


ertomatic:

lars-petter-hauge (approved):
Thanks for correcting 👍

PR #333, opened 2021-06-16
Change output dir for reports

### LINKED ISSUE #321: Change default output directory name for workflow utility SemeioScript
https://github.com/equinor/semeio/blob/3605d379b4af1f4f221050d93a2038ac5606ad46/semeio/communication/semeio_script.py#L81
Having the results ert/output/storage/reports/currentcasename/workflow_name is not ideal since we need to be able to relate them to a case (ert config name) and avoid overwritting if the workflow is run from different ert config files. Could it be changed to  ert/output/storage/reports/ertconfigname/currentcasename/workflow_name ? i.e. change base_dir to be Enspath for example.

oyvindeide:
Closes #321

### DIFF
--- a/semeio/communication/semeio_script.py
+++ b/semeio/communication/semeio_script.py
@@ -93,15 +93,15 @@ class SemeioScript(ErtScript):
         return self._reporter
 
     @property
-    def output_dir(self):
-        return self._output_dir
+    def _reports_dir(self):
+        return self.reporter._output_dir
 
-    @output_dir.setter
-    def output_dir(self, output_dir):
+    @_reports_dir.setter
+    def _reports_dir(self, output_dir):
         output_dir = Path(output_dir)
         if not output_dir.is_absolute():
             res_config = self.ert().resConfig()
             base_dir = Path(res_config.model_config.getEnspath()).parent.absolute()
-            self._reporter._output_dir = base_dir / output_dir
+            self.reporter._output_dir = base_dir / output_dir
         else:
-            self._reporter._output_dir = output_dir
+            self.reporter._output_dir = output_dir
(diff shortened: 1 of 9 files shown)


semeio/communication/semeio_script.py
```diff
@@ -91,3 +91,17 @@ def _get_output_dir(self):
     @property
     def reporter(self):
         return self._reporter
+
+    @property
+    def output_dir(self):
+        return self._output_dir
+
+    @output_dir.setter
+    def output_dir(self, output_dir):
```
markusdregi, line 100: Do we really need to expose this publicly? Could it not be part of the initialisation?

oyvindeide: > Do we really need to expose this publicly? Could it not be part of the initialisation?

Technically we could, but, because we are wrapping `ErtScript` that would be very hacky as we dont control the creation of the class when the workflow is run from `ert`.

markusdregi: That is true. But since we are only using it from within `SemeioScript`, perhaps we could make the setter private? 🤔

oyvindeide: I`ve no problem with that solution

semeio/communication/semeio_script.py
```diff
@@ -83,10 +83,25 @@ def _get_output_dir(self):
...
             sub_dir,
             type(self).__name__,
         )
 
     @property
     def reporter(self):
         return self._reporter
+
+    @property
+    def _reports_dir(self):
```
markusdregi, line 96: How is this working out when the method and the private variable shares the same name?

markusdregi (approved):
Looks good to me 🚀

PR #334, opened 2021-06-17
ahm_analysis script: Correct misfit calculation and update test 

mareco701:
This is to correct the calc_observationsgroup_misfit function to actually return the average of the misfit over all the realizations in the ensemble and not only the  value of realization 0. This also then prevents the workflow to fail if realization 0 has failed ( so has no data).


ertomatic:

oyvindeide (approved):
Hi and thanks for your PR! Looks very good, thanks for the bug fix!

mareco701:
Thank you and for information I cannot merge the PR (do not have the right to) so you have to it for me if ok :-).

oyvindeide:
Jenkins: test this please

PR #335, opened 2021-06-23
Add documentation from json schema

### LINKED ISSUE #325: [MISFIT_PREPROCESSOR] Reintroduce schema documentation
As we changed from Configsuite to Pydantic we lost the documentation of the schema, we should get that part of the documentation restored.

oyvindeide:
Closes #325

There are alternatives to implementing something new, and I would really like to do that, except I can not find any existing good solution. There is `jsonschema2rst`, but the only documented entry point is from the command line:
```python
    import tempfile
    import subprocess
    import json
    import os

    with tempfile.TemporaryDirectory() as tmpdirname:
        with open(os.path.join(tmpdirname, "json_schema.json"), "w") as fout:
            json.dump(schema, fout)
        subprocess.call(
            ["jsonschema2rst", tmpdirname, os.path.join(tmpdirname, "result")]
        )
        with open(os.path.join(tmpdirname, "result", "json_schema.rst"), "r") as fin:
            rst_result = fin.read()
```
This _could_ work (or we could call the internal parser in `jsonschema2rst`, but is not documented). Though the vanilla result from this is not valid rst, so need to do some post-processing.

### DIFF
--- a/semeio/_docs_utils/_json_schema_2_rst.py
+++ b/semeio/_docs_utils/_json_schema_2_rst.py
@@ -1,0 +1,84 @@
+from copy import deepcopy
+
+
+def _insert_ref(schema):
+    for k, v in schema.items():
+        if isinstance(v, dict):
+            _insert_ref(v)
+        elif isinstance(v, list):
+            for index, val in enumerate(v.copy()):
+                if "$ref" in val:
+                    v[index] = val["$ref"] + "_"
+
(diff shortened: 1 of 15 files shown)


jondequinor:
I don't mind this, but I think you should write some of this information in the module itself.

jondequinor (approved):

PR #336, opened 2021-06-29
Add localisation mvp

oddvarlia:
A first version of the workflow script to specify localisation from a configuration file. 
This PR covers parameters of type:

- scalar parameters specified by ERT keyword GEN_KW   (Parameters of form  nodename;parametername)
- scalar parameters specified by ERT keyword GEN_PARAM  (Parameters of form nodename:index
- field parameters specified by ERT keyword FIELD              (Nodename)
- field parameters specified by ERT keyword SURFACE        (Nodename)

For parameters of type FIELD or SURFACE, there is an option to specify scaling of correlations using a scaling factor.
Implemented scaling functions depends on distance from reference point (typically observation localisation). The currently implemented functions are:

- Gaussian decay from a reference point
- Exponential decay from a reference point

Scaling of correlations between field parameters and observations apply row_scaling function from libres.

### DIFF
--- a/semeio/workflows/localisation/local_config_script.py
+++ b/semeio/workflows/localisation/local_config_script.py
@@ -1,0 +1,45 @@
+from ert_shared.libres_facade import LibresFacade
+from ert_shared.plugins.plugin_manager import hook_implementation
+
+import semeio.workflows.localisation.local_script_lib as local
+from semeio.communication import SemeioScript
+from semeio.workflows.localisation.localisation_config import LocalisationConfig
+
+
+class LocalisationConfigJob(SemeioScript):
+    def run(self, *args):
+        ert = self.ert()
+        facade = LibresFacade(self.ert())
+        # Clear all correlations
+        local.clear_correlations(ert)
+
+        # Read yml file with specifications
+        config_dict = local.read_localisation_config(args)
+
+        # Get all observations from ert instance
+        obs_keys = [
+            facade.get_observation_key(nr)
+            for nr, _ in enumerate(facade.get_observations())
+        ]
+
+        ert_parameters = local.get_param_from_ert(ert.ensembleConfig())
+
+        config = LocalisationConfig(
+            observations=obs_keys,
+            parameters=ert_parameters.to_list(),
+            **config_dict,
+        )
+
+        local.add_ministeps(
+            config,
+            ert_parameters.to_dict(),
+            ert.getLocalConfig(),
+            ert.ensembleConfig(),
+            ert.eclConfig().getGrid(),
+        )
+
+
+@hook_implementation
+def legacy_ertscript_workflow(config):
+    workflow = config.add_workflow(LocalisationConfigJob, "LOCALISATION_JOB")
+    workflow.category = "observations.correlation"
--- a/semeio/workflows/localisation/local_script_lib.py
+++ b/semeio/workflows/localisation/local_script_lib.py
@@ -1,0 +1,517 @@
+# pylint: disable=W0201
+import math
+from collections import defaultdict
+from typing import List
+
+import yaml
+import cwrap
+import numpy as np
+import logging
+
+from ecl.util.geometry import Surface
+from res.enkf.enums.ert_impl_type_enum import ErtImplType
+from res.enkf.enums.enkf_var_type_enum import EnkfVarType
+
+from dataclasses import dataclass, field
+
+from semeio.workflows.localisation.localisation_debug_settings import (
+    LogLevel,
+    debug_print,
+)
+
+
+@dataclass
+class Parameter:
+    name: str
+    type: str
+    parameters: List = field(default_factory=list)
+
+    def to_list(self):
+        if self.parameters:
+            return [f"{self.name}:{parameter}" for parameter in self.parameters]
+        else:
+            return [f"{self.name}"]
+
+    def to_dict(self):
+        return {self.name: self.parameters}
+
+
+@dataclass
+class Parameters:
+    parameters: List[Parameter] = field(default_factory=list)
+
+    def append(self, new):
+        self.parameters.append(new)
+
+    def to_list(self):
+        # pylint: disable=E1133
+        result = []
+        for parameter in self.parameters:
+            result.extend(parameter.to_list())
+        return result
+
+    def to_dict(self):
+        # pylint: disable=E1133
+        result = {}
+        for parameter in self.parameters:
+            if parameter.name in result:
+                raise ValueError(f"Duplicate parameters found: {parameter.name}")
+            result.update(parameter.to_dict())
+        return result
+
+    @classmethod
+    def from_list(cls, input_list):
+        result = defaultdict(list)
+        for item in input_list:
+            words = item.split(":")
+            if len(words) == 1:
+                name = words[0]
+                parameters = None
+            elif len(words) == 2:
+                name, parameters = words
+            else:
+                raise ValueError(f"Too many : in {item}")
+            if name in result:
+                if not parameters:
+                    raise ValueError(
+                        f"Inconsistent parameters, found {name} in "
+                        f"{dict(result)}, but did not find parameters"
+                    )
+                if not result[name]:
+                    raise ValueError(
+                        f"Inconsistent parameters, found {name} in {dict(result)} but "
+                        f"did not expect parameters, found {parameters}"
+                    )
+            if parameters:
+                result[name].append(parameters)
+            else:
+                result[name] = []
+        return cls([Parameter(key, "UNKNOWN", val) for key, val in result.items()])
+
+
+def get_param_from_ert(ens_config):
+    new_params = Parameters()
+    keylist = ens_config.alloc_keylist()
+    implementation_type_not_scalar = [
+        ErtImplType.GEN_DATA,
+        ErtImplType.FIELD,
+        ErtImplType.SURFACE,
+    ]
+    for key in keylist:
+        node = ens_config.getNode(key)
+        impl_type = node.getImplementationType()
+        if node.getVariableType() == EnkfVarType.PARAMETER:
+            my_param = Parameter(key, impl_type)
+            new_params.append(my_param)
+            if impl_type == ErtImplType.GEN_KW:
+                # Node contains scalar parameters defined by GEN_KW
+                kw_config_model = node.getKeywordModelConfig()
+                my_param.parameters = kw_config_model.getKeyWords().strings
+            elif impl_type in implementation_type_not_scalar:
+                # Node contains parameter from FIELD, SURFACE or GEN_PARAM
+                # The parameters_for_node dict contains empty list for FIELD
+                # and SURFACE.
+                # The number of variables for a FIELD parameter is
+                # defined by the grid in the GRID keyword.
+                # The number of variables for a SURFACE parameter is
+                # defined by the size of a surface object.
+                # For GEN_PARAM the list contains the number of variables.
+                # parameters_for_node[key] = []
+                if impl_type == ErtImplType.GEN_DATA:
+                    gen_data_config = node.getDataModelConfig()
+                    data_size = gen_data_config.get_initial_size()
+                    if data_size <= 0:
+                        # Cannot here know if it will be used or not.
+                        logging.warning(
+                            "\nThe ERT config has defined parameter nodes of type\n"
+                            "GEN_PARAM. If this is used in localisation, \n"
+                            "the localisation workflow must be run AFTER initial \n"
+                            "ensemble is created, but BEFORE first update is run.\n\n"
+                        )
+
+                    my_param.parameters = [str(item) for item in range(data_size)]
+
+    return new_params
+
+
+def read_localisation_config(args):
+    if len(args) == 1:
+        specification_file_name = args[0]
+    else:
+        raise ValueError(f"Expecting a single argument. Got {args} arguments.")
+
+    print(f"\nDefine localisation setup using config file: {specification_file_name}")
+    logging.info(
+        "\nDefine localisation setup using config file: %s", specification_file_name
+    )
+    with open(specification_file_name, "r", encoding="utf-8") as yml_file:
+        localisation_yml = yaml.safe_load(yml_file)
+    return localisation_yml
+
+
+def active_index_for_parameter(node_name, param_name, ert_param_dict):
+    # For parameters defined as scalar parameters (coming from GEN_KW)
+    # the parameters for a node have a name. Get the index from the order
+    # of these parameters.
+
+    ert_param_list = ert_param_dict[node_name]
+    if len(ert_param_list) > 0:
+        index = -1
+        for count, name in enumerate(ert_param_list):
+            if name == param_name:
+                index = count
+                break
+        assert index > -1
+    return index
+
+
+def activate_gen_kw_param(
+    model_param_group, node_name, param_list, ert_param_dict, log_level=LogLevel.OFF
+):
+    """
+    Activate the selected parameters for the specified node.
+    The param_list contains the list of parameters defined in GEN_KW
+    for this node to be activated.
+    """
+    active_param_list = model_param_group.getActiveList(node_name)
+    debug_print("Set active parameters", LogLevel.LEVEL2, log_level)
+    for param_name in param_list:
+        index = active_index_for_parameter(node_name, param_name, ert_param_dict)
+        if index is not None:
+            debug_print(
+                f"Active parameter: {param_name}  index: {index}",
+                LogLevel.LEVEL3,
+                log_level,
+            )
+            active_param_list.addActiveIndex(index)
+
+
+def activate_gen_param(
+    model_param_group, node_name, param_list, data_size, log_level=LogLevel.OFF
+):
+    """
+    Activate the selected parameters for the specified node.
+    The param_list contains a list of names that are integer numbers
+    for the parameter indices to be activated for parameters belonging
+    to the specified GEN_PARAM node.
+    """
+    active_param_list = model_param_group.getActiveList(node_name)
+    for param_name in param_list:
+        index = int(param_name)
+        if index < 0 or index >= data_size:
+            raise ValueError(
+                f"Index for parameter in node {node_name} is "
+                f"outside the interval [0,{data_size-1}]"
+            )
+        debug_print(f"Active parameter index: {index}", LogLevel.LEVEL3, log_level)
+        active_param_list.addActiveIndex(index)
+
+
+def activate_and_scale_correlation(
+    grid,
+    method,
+    row_scaling,
+    ref_pos,
+    data_size,
+    main_range,
+    perp_range,
+    azimuth,
+    log_level=LogLevel.OFF,
+):
+    debug_print(
+        f"Scale correlations using method: {method}",
+        LogLevel.LEVEL3,
+        log_level,
+    )
+    if method == "gaussian_decay":
+        row_scaling.assign(
+            data_size,
+            GaussianDecay(ref_pos, main_range, perp_range, azimuth, grid),
+        )
+    elif method == "exponential_decay":
+        row_scaling.assign(
+            data_size,
+            ExponentialDecay(ref_pos, main_range, perp_range, azimuth, grid),
+        )
+    else:
+        logging.warning("Scaling method: %s is not implemented.", method)
+
+
+def add_ministeps(
+    user_config,
+    ert_param_dict,
+    ert_local_config,
+    ert_ensemble_config,
+    grid_for_field,
+):
+    # pylint: disable-msg=too-many-branches
+    # pylint: disable-msg=R0915
+
+    debug_print("Add all ministeps:", LogLevel.LEVEL1, user_config.log_level)
+    ScalingValues.initialize()
+    for count, corr_spec in enumerate(user_config.correlations):
+        ministep_name = corr_spec.name
+        ministep = ert_local_config.createMinistep(ministep_name)
+        debug_print(
+            f"Define ministep: {ministep_name}", LogLevel.LEVEL1, user_config.log_level
+        )
+
+        param_group_name = ministep_name + "_param_group"
+        model_param_group = ert_local_config.createDataset(param_group_name)
+
+        obs_group_name = ministep_name + "_obs_group"
+        obs_group = ert_local_config.createObsdata(obs_group_name)
+
+        obs_list = corr_spec.obs_group.result_items
+        param_dict = Parameters.from_list(corr_spec.param_group.result_items).to_dict()
+
+        # Setup model parameter group
+        for node_name, param_list in param_dict.items():
+            node = ert_ensemble_config.getNode(node_name)
+            impl_type = node.getImplementationType()
+
+            debug_print(
+                f"Add node: {node_name} of type: {impl_type}",
+                LogLevel.LEVEL2,
+                user_config.log_level,
+            )
+            model_param_group.addNode(node_name)
+            if impl_type == ErtImplType.GEN_KW:
+                activate_gen_kw_param(
+                    model_param_group,
+                    node_name,
+                    param_list,
+                    ert_param_dict,
+                    user_config.log_level,
+                )
+            elif impl_type == ErtImplType.FIELD:
+                if corr_spec.field_scale is not None:
+                    field_config = node.getFieldModelConfig()
+                    check_if_ref_point_in_grid(corr_spec.ref_point, grid_for_field)
+                    activate_and_scale_correlation(
+                        grid_for_field,
+                        corr_spec.field_scale.method,
+                        model_param_group.row_scaling(node_name),
+                        corr_spec.ref_point,
+                        field_config.get_data_size(),
+                        corr_spec.field_scale.main_range,
+                        corr_spec.field_scale.perp_range,
+                        corr_spec.field_scale.azimuth,
+                        user_config.log_level,
+                    )
+                    if user_config.write_scaling_factors:
+                        ScalingValues.write(
+                            node_name, corr_spec, grid_for_field, user_config.log_level
+                        )
+                else:
+                    debug_print(
+                        f"No correlation scaling for node {node_name} "
+                        f"in {ministep_name}",
+                        LogLevel.LEVEL3,
+                        user_config.log_level,
+                    )
+            elif impl_type == ErtImplType.GEN_DATA:
+                gen_data_config = node.getDataModelConfig()
+                data_size = gen_data_config.get_initial_size()
+                if data_size > 0:
+                    activate_gen_param(
+                        model_param_group,
+                        node_name,
+                        param_list,
+                        data_size,
+                        user_config.log_level,
+                    )
+                else:
+                    debug_print(
+                        f"Parameter {node_name} has data size: {data_size} "
+                        f"in {ministep_name}",
+                        LogLevel.LEVEL3,
+                        user_config.log_level,
+                    )
+            elif impl_type == ErtImplType.SURFACE:
+                if corr_spec.surface_scale is not None:
+                    surface_file = corr_spec.surface_scale.surface_file
+                    debug_print(
+                        f"Get surface size from: {surface_file}",
+                        LogLevel.LEVEL3,
+                        user_config.log_level,
+                    )
+                    surface = Surface(surface_file)
+                    data_size = surface.getNX() * surface.getNY()
+                    activate_and_scale_correlation(
+                        surface,
+                        corr_spec.surface_scale.method,
+                        model_param_group.row_scaling(node_name),
+                        corr_spec.ref_point,
+                        data_size,
+                        corr_spec.surface_scale.main_range,
+                        corr_spec.surface_scale.perp_range,
+                        corr_spec.surface_scale.azimuth,
+                    )
+                else:
+                    debug_print(
+                        f"No correlation scaling for node {node_name} "
+                        f"in {ministep_name}",
+                        LogLevel.LEVEL3,
+                        user_config.log_level,
+                    )
+
+        # Setup observation group
+        for obs_name in obs_list:
+            debug_print(
+                f"Add obs node: {obs_name}", LogLevel.LEVEL2, user_config.log_level
+            )
+            obs_group.addNode(obs_name)
+
+        # Setup ministep
+        debug_print(
+            f"Attach {param_group_name} to ministep {ministep_name}",
+            LogLevel.LEVEL1,
+            user_config.log_level,
+        )
+        ministep.attachDataset(model_param_group)
+
+        debug_print(
+            f"Attach {obs_group_name} to ministep {ministep_name}",
+            LogLevel.LEVEL1,
+            user_config.log_level,
+        )
+        ministep.attachObsset(obs_group)
+
+        debug_print(
+            f"Add {ministep_name} to update step\n",
+            LogLevel.LEVEL1,
+            user_config.log_level,
+        )
+        ert_local_config.getUpdatestep().attachMinistep(ministep)
+
+
+def clear_correlations(ert):
+    local_config = ert.getLocalConfig()
+    local_config.clear()
+
+
+def check_if_ref_point_in_grid(ref_point, grid):
+    try:
+        grid.find_cell_xy(ref_point[0], ref_point[1], 0)
+    except ValueError as err:
+        raise ValueError(
+            f"Reference point {ref_point} corresponds to undefined grid cell "
+            f"or is outside the area defined by the grid {grid.get_name()}\n"
+            "Check specification of reference point."
+        ) from err
+
+
+@dataclass
+class Decay:
+    obs_pos: list
+    main_range: float
+    perp_range: float
+    azimuth: float
+    grid: object
+
+    def __post_init__(self):
+        angle = (90.0 - self.azimuth) * math.pi / 180.0
+        self.cosangle = math.cos(angle)
+        self.sinangle = math.sin(angle)
+
+    def get_dx_dy(self, data_index):
+        try:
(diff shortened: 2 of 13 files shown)


ertomatic:

tests/jobs/localisation/test_integration.py
```diff
@@ -0,0 +1,252 @@
...
+        for count, obsnode in enumerate(ministep.getLocalObsData()):
+            obs_list.append(obsnode.key())
+        obs_list.sort()
+
+        if index in [0, 1]:
+            assert obs_list == expected_obs1
+        else:
+            assert obs_list == expected_obs2
+        key = ministep_names[index] + "_param_group"
+        assert ministep[key].keys() == expected
```
oyvindeide, line 80: Because you are not using parameterize for running more than one case here, I would suggest you rewrite this slightly, my suggestion would be something like this:

```suggestion
@pytest.mark.parametrize(
    "obs_group_add, param_group_add, expected",
    [
        (
            ["FOPR", "WOPR_OP1_190"],
            ["SNAKE_OIL_PARAM:OP1_DIVERGENCE_SCALE", "SNAKE_OIL_PARAM:OP1_OFFSET"],
            ["SNAKE_OIL_PARAM"],
        ),
    ],
)
def test_localisation(
    setup_ert, obs_group_add, param_group_add, expected
):
    ert = EnKFMain(setup_ert)
    config = {
        "correlations": [
            {
                "name": "CORR1",
                "obs_group": {"add": "*", "remove": obs_group_add},
                "param_group": {
                    "add": "SNAKE_OIL_PARAM:OP1_PERSISTENCE",
                },
            },
            {
                "name": "CORR2",
                "obs_group": {"add": "*", "remove": obs_group_add},
                "param_group": {
                    "add": "SNAKE_OIL_PARAM:*",
                    "remove": "SNAKE_OIL_PARAM:OP1_PERSISTENCE",
                },
            },
            {
                "name": "CORR3",
                "obs_group": {"add": obs_group_add},
                "param_group": {
                    "add": param_group_add,
                },
            },
        ],
    }
    with open("local_config.yaml", "w") as fout:
        yaml.dump(config, fout)
    LocalisationConfigJob(ert).run("local_config.yaml")
    assert ert.getLocalConfig().getMinistep("CORR1").name() == "CORR1"
    assert (
        ert.getLocalConfig().getObsdata("CORR1_obs_group").name() == "CORR1_obs_group"
    )
    result = {}
    for index, ministep in enumerate(ert.getLocalConfig().getUpdatestep()):
        result[ministep.name()] = {
            "obs": [obs_node.key() for obs_node in ministep.getLocalObsData()],
            "key": ministep.name() + "_param_group",
        }

    expected_result = {
        "CORR1": {
            "obs": [
                "WOPR_OP1_108",
                "WOPR_OP1_144",
                "WOPR_OP1_36",
                "WOPR_OP1_72",
                "WOPR_OP1_9",
                "WPR_DIFF_1",
            ],
            "key": "CORR1_param_group",
        },
        "CORR2": {
            "obs": [
                "WOPR_OP1_108",
                "WOPR_OP1_144",
                "WOPR_OP1_36",
                "WOPR_OP1_72",
                "WOPR_OP1_9",
                "WPR_DIFF_1",
            ],
            "key": "CORR2_param_group",
        },
        "CORR3": {"obs": ["FOPR", "WOPR_OP1_190"], "key": "CORR3_param_group"},
    }
    assert result == expected_result
```

It is always good to avoid `assert` in loops where possible, as that can obscure what the error really is, and `pytest` is really good at showing comparisons when asserting on lists and dicts. This can easily be converted back in to a parameterized test if you would like to add another test.

oyvindeide (commented):
There is quite a lot that has been added, but that is not being used, so I am having a hard time reviewing the code. Is everything part of a MVP, is for example the surface part needed in a first pass? It adds a lot of complexity, and would it not be possible to split it into a separate PR?

semeio/workflows/localisation/list_all_obs_and_model_param.py
```diff
@@ -0,0 +1,90 @@
...
+    result_node_names = []
+    for key in keylist:
+        node = ens_config.getNode(key)
+        variable_type = node.getVariableType()
+        if variable_type == EnkfVarType.DYNAMIC_RESULT:
+            result_node_names.append(key)
+    return result_node_names
+
+
+class LocalisationConfigJob(ErtScript):
```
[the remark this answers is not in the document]
oddvarlia: list_all_obs_and_model_param.py should not be a part of any release of localisation. Its purpose is for me to check functionality of libres and to print out which observations and parameters are defined for an ERT project to simplify specification of localisation config files. So it is no point spending time on this file, but I have it in git to have version control over it. (A polished version of it could of course also be of help for end users, but that is another thing to be discussed with Eivind S and Tudor.

tests/jobs/localisation/test_integration.py
oddvarlia: I have included your version of this test, but I am not sure I should remove the original test since it actually loops through the created ministeps and get info from those and verify that the created ministeps contain what they should. I thought thtat could be a good test to check that the ministeps are initialized as they should. By keeping both your version and the original the simplicity is take care of by your test and the check for the state of the ministep objects by the original test

oddvarlia:
> There is quite a lot that has been added, but that is not being used, so I am having a hard time reviewing the code. Is everything part of a MVP, is for example the surface part needed in a first pass? It adds a lot of complexity, and would it not be possible to split it into a separate PR?

There are some functions not in use in the code at the moment:
- I had to move a check for specification of reference point to add_ministep function since I needed access to the ERT instance to get access to the libres function that could look up the grid cell for the reference point to check that it was within the grid or not. This resulted in some code that can be removed. I will do that to clean it up. My intention was to do the checks as far as possible in the localisaton_config.py file where pydantic is used for validation, but we don't want this to depend on the ERT instance. The ERT instance is accessed in a few functions to make the code more clean and easy to validate. 

When it comes to SURFACE keyword I got one problem: There were not libres function to get access to a reference surface which I will need to get access to the dimensions and orientation of the surface in order to verify that the reference point is within the surface area. So the current solution is to specify the reference surface (used in the ERT config file) also in the localisation config file. I asked in ERT_DEV channel and also Joakim about this but assume vacation is the reason for no response so far, but I think I know the anser already: There are no way to get the info I want from the ERT instance and libres and libecl. I have to specify the reference surface myself instead in the localisation config file.

When it comes to GEN_PARAM: The info about the size (number of parameters belonging to a GEN_PARAM node) can only be accessed after initialisation of initial ensemble. Hence the localisation workflow MUST be run after initialisation and prior to the first update. (This works). The parameters in GEN_PARAM is currently treated in the same ways as parameters for nodes specified by GEN_KW but with the difference that I have to use the indices for the parameters since they are not named as in GEN_KW. This solution is one of several alternative ways of handling GEN_PARAM parameters. Since GEN_PARAM is so general, it could also be used to keep FIELD data (1D, 2D or 3D) and row_scaling can be applied (as soon as a spatial location is associated with each parameter (each index). This is a proposal I have briefly discussed with project leader Remus but not with Eivind S and Tudor yet.

I think the current treatment of GEN_PARAM parameters using parameter name expansion such that a GEN _PARAM node  genparam with N parameters is a bit strange, but it works. The current treatment is that e.g genparam:*  is expanded to the parameter names: genparam:1  genparam:2 ... genparam:(N-1)

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -0,0 +1,444 @@
...
+    ert_param_list = ert_param_dict[node_name]
+    if len(ert_param_list) > 0:
+        index = -1
+        for count, name in enumerate(ert_param_list):
+            if name == param_name:
+                index = count
+                break
+        assert index > -1
+    else:
+        index = None
```
oyvindeide, line 172: This never happens in a test, can it happen in a real scenario?

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -0,0 +1,444 @@
...
+    return index
+
+
+def activate_gen_kw_param(model_param_group, node_name, param_list, ert_param_dict):
```
oyvindeide, line 176: Should be possible to merge this with `activate_gen_param`?

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+        (
+            "PARAM_FIELD1",
+            ["PARAM_FIELD1"],
+        ),
+        (
+            "PARAM_GEN:*",
+            ["PARAM_GEN:0", "PARAM_GEN:1", "PARAM_GEN:2", "PARAM_GEN:3", "PARAM_GEN:4"],
+        ),
+    ],
+)
```
oyvindeide, line 185: Are all of these testing something unique, or is there a lot of repetition here? If there is duplication, a simplification would be good.

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+def test_simple_config(param_group_add, expected):
+    data = {
+        "log_level": 2,
```
oyvindeide, line 188: Not relevant in this test is it?

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+                "obs_group": {"add": ["OBS1"]},
+                "param_group": {
+                    "add": param_group_add,
+                },
+            }
+        ],
+    }
+    conf = LocalisationConfig(observations=ERT_OBS, parameters=ERT_PARAM, **data)
+    assert len(conf.correlations) == 1
+    assert conf.correlations[0].obs_group.result_items == ["OBS1"]
```
oyvindeide, line 201: Is this relevant to assert? There is a separate test for observations?

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+            {
+                "name": "some_name",
+                "obs_group": {"add": ["OBS1"]},
+                "param_group": {
+                    "add": param_group_add,
+                },
+            }
+        ],
+    }
+    conf = LocalisationConfig(observations=ERT_OBS, parameters=ERT_PARAM, **data)
```
oyvindeide, line 199: Giving in `ERT_OBS` and `ERT_PARAM` here makes it a bit harder to read the test, could you just not use `["OBS1"]` for observations, and something small enough to cover your use case for `ERT_PARAM`?

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+            "P*:*:*",
+            "No match for: P*:*:*",
+        ),
+        (
+            [],
+            "PARAM_NODE1:PARAM1",
+            [],
+            "",
+        ),
+    ],
```
oyvindeide, line 238: Same comment as above, are these all testing something unique?

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+                },
+                "param_group": {
+                    "add": param_group_add,
+                    "remove": param_group_remove,
+                },
+            }
+        ],
+    }
+    with pytest.raises(pydantic.error_wrappers.ValidationError, match=expected_error):
+        LocalisationConfig(observations=ERT_OBS, parameters=ERT_PARAM, **data)
```
oyvindeide, line 259: Same comment WRT `ERT_OBS` and `ERT_PARAM`

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+                "PARAM_GEN:0",
+                "PARAM_GEN:1",
+                "PARAM_GEN:2",
+                "PARAM_GEN:3",
+                "PARAM_GEN:4",
+                "PARAM_SURFACE1",
+                "PARAM_SURFACE2",
+            ],
+        ),
+    ],
```
oyvindeide, line 485: Same comment WRT uniqueness .

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+        ),
+        (
+            "OBS1*",
+            "PARAM_FIELD1",
+            [500, 750],
+            "gaussian_decay",
+            0.1,
+            0.1,
+            0,
+            ["OBS1", "OBS11", "OBS12", "OBS13", "OBS14"],
```
oyvindeide, line 667: If there values does not change during parameterize is should probably only be written in the test body.

tests/jobs/localisation/test_integration.py
```diff
@@ -0,0 +1,262 @@
...
+                "param_group": {
+                    "add": "*",
+                },
+            },
+        ],
+    }
+
+    with open("local_config.yaml", "w") as fout:
+        yaml.dump(config, fout)
+    LocalisationConfigJob(ert).run("local_config.yaml")
```
oyvindeide, line 126: Is it possible to assert something after this test?

semeio/workflows/localisation/local_config_script.py
```diff
@@ -4,38 +4,45 @@
...
 
         # Get all observations from ert instance
-        ert_obs_list = local.get_observations_from_ert(ert)
+        obs_keys = [
+            facade.get_observation_key(nr)
+            for nr, _ in enumerate(facade.get_observations())
+        ]
 
-        ert_param_dict, ert_node_type_dict = local.get_param_from_ert(ert)
+        ert_parameters = local.get_param_from_ert(ert.ensembleConfig())
```
oyvindeide, line 31: If possible it is always best to send in variables to functions as opposed to unpacking them on the function side. That way the functions have to know a minimum about the structure of the input.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -1,42 +1,98 @@
...
+            return [f"{self.name}:{parameter}" for parameter in self.parameters]
+        else:
+            return [f"{self.name}"]
+
+    def to_dict(self):
+        return {self.name: self.parameters}
+
+
+@dataclass
+class Parameters:
```
oyvindeide, line 43: I wanted to change this into a self contained class which was able to convert between the list and dict formats.

oyvindeide (commented):
It is starting to look good. I have made some changes, can you confirm that everything is still working as expected?

tests/jobs/localisation/test_config1.py
```diff
@@ -0,0 +1,831 @@
...
+                    "method": method,
+                    "main_range": range1,
+                    "perp_range": range2,
+                    "angle": angle,
+                    "filename": "surface.txt",
+                },
+            }
+        ],
+    }
+    conf = LocalisationConfig(observations=ERT_OBS, parameters=ERT_PARAM, **data)
```
oyvindeide, line 776: This is not really asserting anything, what is the main purpose of the test?

semeio/workflows/localisation/local_script_lib.py
oddvarlia: No it is only used with GEN_KW parameters and a node of type GEN_KW should always have at least one active parameter to be defined. We can remove this else branch.

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Since GEN_KW uses parameter names and GEN_PARAM uses indices only to identify parameters, I think it is OK to have two separate functions, otherwise we have to check on implementation type on this level also if we merge active_gen_param with activate_gen_kw_param

tests/jobs/localisation/test_config1.py
oddvarlia: Maybe not, but I often run with -s to check the output and in that case it might be relevant.

tests/jobs/localisation/test_config1.py
oddvarlia: It is not relevant here. It is duplicated in test for observations.

tests/jobs/localisation/test_config1.py
oddvarlia: I think these tests are relevant. They test misspelling  like the P*;*:* and tests the 'remove' part of the specification.

tests/jobs/localisation/test_config1.py
oddvarlia: I added so many parameter names to test expansion  for both add and remove part for both parameters and observations. I have 2-3 parameters of the various implementation types and for GEN_KW parameters I have several parameter nodes with 2-3 parameters per node to test expansion that effect both node name and parameter name for same node.

It is of course possible to reduce the number of tests by first removing some parameters or parameter nodes in ERT_PARAM.

semeio/workflows/localisation/local_script_lib.py
oddvarlia: ok

tests/jobs/localisation/test_config1.py
oddvarlia: Most of the integration tests does not assert anything or just a few things. I have used the tests to verify that the loclisation config is created without errors and as a fast way to check that and find bugs when developing the code since it is faster than opening the ert gui and run the reek or drogon cases. The tests create the input data they need also.

tests/jobs/localisation/test_integration.py
oddvarlia: As long as the ert model is not run but only the localisation config job, the only I know about can be asserted is that the object created and updated for ministeps contain the correct parameter nodes, parameter names, observation names and which parameters are active or not. No check of the effect can be asserted unless the ert is run through an update step with ensemble smoother and we can compare the result with the expected result we know is correct. Any verification of correctness is difficult (visualisation of e.g fields can be one solution) except for very simple (and often uninterested cases), But it is of course possible to verify that an expected result is reproduced.  Maybe we can include such a test (compare with reference) later ?

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -21,6 +20,7 @@
 
 # The global variable is only used for controlling output of log info to screen
 debug_level = log_level_setting.debug_level
+scaling_param_number = log_level_setting.scaling_param_number
```
oyvindeide: I really think we should avoid global parameters, so if you have a rewrite that will get rid of it, I suggest you add that in this PR.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -236,21 +236,24 @@ def write_scaling_values(
         raise KeyError(f" Method name: {method_name} is not implemented")
 
     scaling_values = np.zeros((nx, ny, nz), dtype=np.float32)
-    for k, j, i in itertools.product(nz, ny, nx):
-        global_index = grid_for_field.global_index(ijk=(i, j, k))
-        scaling_values[i, j, k] = decay_obj(global_index)
+    for k in range(nz):
+        for j in range(ny):
```
oyvindeide, line 240: This was just a typo on my end, it should be `itertools.product(range(nz), range(ny), range(nx))`.

The fact that this was not discovered in the testing is worrying me a bit, can we add a test covering it?

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -236,21 +236,24 @@ def write_scaling_values(
...
+    filename = corr_name + "_" + node_name + "_" + scaling_kw_name + ".GRDECL"
+    debug_print(f"Write file: {filename}", LogLevel.LEVEL3)
     with cwrap.open(filename, "w") as file:
         grid_for_field.write_grdecl(scaling_kw, file)
 
 
 def activate_and_scale_correlation(
     grid, method, row_scaling, ref_pos, data_size, main_range, perp_range, azimuth
 ):
+    debug_print(f"Scale correlations using method: {method}", LogLevel.LEVEL3)
```
oyvindeide, line 256: Do we really need all these `print` statements? I get why some of them are important, but are all of them equally relevant?

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Any reason why using itertools ? It is more readable as a triple loop. The reason why it was not testet was that it dependet on a test where LogLevel was set to LEVEL4 and that was not included, but I will remove the dependency on LogLevel for the write scaling to file functionality and then it is natural to have a test where this is activated by a keyword in the config file.

semeio/workflows/localisation/local_script_lib.py
oddvarlia: I prefer to have sufficent prints to be able to see all essential steps in the output log, but it is easy to turn off by adjusting log level.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -344,7 +348,7 @@ def add_ministeps(
                     )
             elif impl_type == ErtImplType.SURFACE:
                 if corr_spec.surface_scale is not None:
-                    surface = Surface(corr_spec.surface_scale.filename)
+                    surface = Surface(corr_spec.surface_scale.surf_filename)
```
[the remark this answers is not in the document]
oddvarlia: Now I have checked it. What seems to work is: the function for the node: get_init_file_fmt() The function get_enkf_infile() returns only None both when running the localisation prior to creating initial ensemble and after the initial ensemble is created. I will update the code to use  get_init_file_fmt() and remove the filename keywword under surface_scale keyword.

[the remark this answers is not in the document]
oddvarlia: I have now rewritten the code to avoid specifying surface file in localisation config. But there are cases where I still need to supply a relative file path to the location of the surface files in the sub keyword INIT_FILES which is part of the SURFACE keyword in ERT config file. There are no way to get the file in the sub keyword BASE_SURFACE which have file path relative to CONFIG_PATH The files in the INIT_FILES subkeywords are often located in other directories. This is why we still must be able to specify a (relative) file path as a keyword in surface_scale

semeio/workflows/localisation/local_script_lib.py
oddvarlia: This is fixed and there are no longer any global parameter.

semeio/workflows/localisation/local_script_lib.py
oyvindeide: The reason for `itertools` is that the triple loop does not really add anything as you are not really doing anything in loops 1 and 2, the idea is that you are looping through a grid, and not the individual layers. Great if you add a test, you can also call the function directly, so you do not depend on the log level for it to run.

semeio/workflows/localisation/list_all_obs_and_model_param.py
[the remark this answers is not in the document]
oddvarlia: It is removed from this PR

tests/jobs/localisation/test_config1.py
oddvarlia: I have removed a few that does not contribute with more than those that still remains.

tests/jobs/localisation/test_config1.py
oddvarlia: Have removes a few that should be covered by those remaining.

semeio/workflows/localisation/local_script_lib.py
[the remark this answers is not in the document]
oddvarlia: Explanation of the issue:
libres seems not to have any function to retrieve the filename for the surface specified in the sub-keyword BASE_SURFACE in keyword SURFACE in ERT config file. There are then two alternatives to be able to get the size of the data for the SURFACE node:
1. Let the user specify the same file also in the localisation config file as in the ERT config file
2. Try a workaround using the libres function that get the surface file name from the INIT_FILES sub keyword of keyword SURFACE. The problem with this is:
   - The file name may be different from the one used in the BASE_SURFACE sub keyword, at least the file path is 
      usually different.
   -  The filename in INIT_FILES may have a %d as part of the name while the filename in BASE_SURFACE does not.
   
   If the only difference between the file name in INIT_FILES and BASE_SURFACE is the %d and the file path, this ca be handled by supplying the file path and replace the %d with 0.
   
What is preferred way of handling this?
1. Let the user supply a relative file path in localisation config file and create the filename we need from INIT_FILES since we cannot get the filename in BASE_SURFACE?
 2. Supply the filename specified in ERT config for BASE_SURFACE once more in the localisation config file?

tests/jobs/localisation/test_config1.py
oddvarlia: I have removed unnecessary assert

tests/jobs/localisation/test_config1.py
oddvarlia: I have removed this test as it does not contribute with anything new compared with the other tests

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Are there any performance gain by using itertools instead of triple loop? In that case I can change this back and use itertools,.  It is used in the function writing calculated scaling factor and that functionality does not any longer depend on log level, only on whether the user has turned on or off writing this to file using a new keyword in the config file. In the updated test_integration I have a test where the calculated scaling factor is written to file.

semeio/workflows/localisation/local_script_lib.py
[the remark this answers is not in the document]
oddvarlia: I have reverted this back to the original solution where the surface file is specified in localisation config file (should be the same as the file specified in the ERT keyword SURFACE under the sub keyword BASE_SURFACE.

semeio/workflows/localisation/localisation_config.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # String with relative path to surface files relative to config path
+        key = "surface_file"
+        if key not in value.keys():
+            raise ValueError(f"Missing keyword: '{key}' in keyword: 'surface_scale' ")
+
+        filename = value[key]
+        # Check  that the file exists
+        if not os.path.exists(filename):
+            raise IOError(f"File for surface: {filename} does not exist.")
```
oyvindeide, line 244: Because this is part of the pydentic schema you want to raise `ValueError` or ` TypeError` so pydantic re-raises as validation error (`ValueError` in this case)

semeio/workflows/localisation/localisation_config.py
```diff
@@ -0,0 +1,311 @@
...
+            return value
+        if not isinstance(value, dict):
+            raise ValueError("value must be dict")
+
+        # String with relative path to surface files relative to config path
+        key = "surface_file"
+        if key not in value.keys():
+            raise ValueError(f"Missing keyword: '{key}' in keyword: 'surface_scale' ")
+
+        filename = value[key]
```
oyvindeide, line 241: Would use:
```suggestion
        filename = pathlib.Path(value[key])
        if not filename.exists():
            ...
```

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -0,0 +1,482 @@
...
+    elif method == "exponential_decay":
+        row_scaling.assign(
+            data_size,
+            ExponentialDecay(ref_pos, main_range, perp_range, azimuth, grid),
+        )
+    else:
+        print(f" --  Scaling method: {method} " "is not implemented.")
+
+
+def get_file_with_surface_format(directory_path, file_with_surface_format):
```
oyvindeide, line 222: Not used

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -0,0 +1,482 @@
...
+        grid.find_cell_xy(ref_point[0], ref_point[1], 0)
+    except ValueError as err:
+        raise ValueError(
+            f"Reference point {ref_point} corresponds to undefined grid cell "
+            f"or is outside the area defined by the grid {grid.get_name()}\n"
+            "Check specification of reference point."
+        ) from err
+
+
+# This is an example callable which decays as a gaussian away from a position
```
oyvindeide, line 369: Is this comment correct? If it is intended it should probably be in the doc string instead?

tests/jobs/localisation/test_configs/test_field_config.py
```diff
@@ -0,0 +1,220 @@
...
+                "surface_file": "surface_file.txt",
+            },
+            id="Field scale and surface scale",
+        ),
+    ],
+)
+def test_correlation_config_ref_point(field_config, surface_config):
+    # Create a surface file (should only exist)
+    with open("surface_file.txt", "w", encoding="utf-8") as fout:
+        fout.write("A test file")
```
oyvindeide, line 115: You dont really need to create a file here, you could just patch it:
```suggestion
import os
from unittest.mock import MagicMock
   
def test_correlation_config_no_ref_point(field_config, surface_config, monkeypatch):
    monkeypatch.setattr(os.path, "exists", MagicMock())
    ...
```

oyvindeide: The same applies below.

oyvindeide (commented):
Looking through now, we are nearly there! :smile:

semeio/workflows/localisation/localisation_config.py
oddvarlia: OK, I change this

semeio/workflows/localisation/localisation_config.py
oddvarlia: OK, changing this

semeio/workflows/localisation/local_script_lib.py
oddvarlia: OK, remove this

semeio/workflows/localisation/local_script_lib.py
oddvarlia: I removed it. Belongs more to docstring, yes.

tests/jobs/localisation/test_configs/test_field_config.py
oddvarlia: Need some more explanation. Where is monkeypatch defined for instance?

tests/jobs/localisation/test_configs/test_field_config.py
```diff
@@ -174,11 +174,8 @@ def test_correlation_config_ref_point(field_config, surface_config):
         ),
     ],
 )
-def test_correlation_config_no_ref_point(field_config, surface_config):
-    # Create a surface file (should only exist)
-    with open("surface_file.txt", "w", encoding="utf-8") as fout:
-        fout.write("A test file")
-
+def test_correlation_config_no_ref_point(field_config, surface_config, monkeypatch):
+    monkeypatch.setattr(os.path, "exists", MagicMock())
```
oyvindeide, line 178: The reason `monkeypatch` does not work is because you are not patching the correct thing. You refactored it from using `os.path.exists` to `pathlib`, so what would work is:
```
monkeypatch.setattr(pathlib.Path, "exists", MagicMock())
```
(also need to import pathlib)

tests/jobs/localisation/test_configs/test_field_config.py
oyvindeide: Documentation is here: https://docs.pytest.org/en/latest/how-to/monkeypatch.html

oddvarlia: OK

tests/jobs/localisation/test_configs/test_field_config.py
oddvarlia: OK

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -0,0 +1,469 @@
...
+                # defined by the grid in the GRID keyword.
+                # The number of variables for a SURFACE parameter is
+                # defined by the size of a surface object.
+                # For GEN_PARAM the list contains the number of variables.
+                # parameters_for_node[key] = []
+                if impl_type == ErtImplType.GEN_DATA:
+                    gen_data_config = node.getDataModelConfig()
+                    data_size = gen_data_config.get_initial_size()
+                    if data_size <= 0:
+                        print(
```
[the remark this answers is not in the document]
oddvarlia: When expanding the <nodename>:*   and nodename is a GEN_PARAM node, there will be 0 parameters in the node if the workflow is run prior to creating initial ensemble and then an error will be raised telling the user that it cannot find any match in ERT instance matching the expansion.  This error seen together with the warning will explain the user why it failed, but without the warning, this error is not easy to understand.

tests/jobs/localisation/test_configs/test_config1.py
```diff
@@ -0,0 +1,518 @@
+from mock import MagicMock
```
oyvindeide, line 1: Can you rename this file to something more descriptive?

tests/legacy_test_data/snake_oil/jobs/snake_oil_simulator.py
```diff
@@ -22,7 +22,7 @@ def readParameters(filename):
 
 
 def runSimulator(simulator, history_simulator, time_step_count):
-    """ @rtype: EclSum """
+    """@rtype: EclSum"""
```
oyvindeide, line 25: Can you add this in a separate commit?

semeio/workflows/correlated_observations_scaling/cos.py
```diff
@@ -7,9 +7,7 @@
 from ert_shared.libres_facade import LibresFacade
 from ert_shared.plugins.plugin_manager import hook_implementation
 from semeio.communication import SemeioScript
-from semeio.workflows.correlated_observations_scaling import (
-    ObservationScaleFactor,
-)
+from semeio.workflows.correlated_observations_scaling import ObservationScaleFactor
```
oyvindeide, line 10: Can you add this as a separate commit?

oyvindeide (commented):
Just a few finishing touches and it is good to go! Could you make the commit header a bit shorter, and instead write more text in the commit body? Something like `Add localisation workflow` should be enough as the header, but a longer description in the body would be good.

oyvindeide (approved):
LGTM!

PR #343, opened 2021-08-25
Extend functionality for localisation of fields with new methods

oddvarlia:
Extend functionality for localisation of ERT nodes of type FIELD.
New methods:

- Read scaling factor from file
- Read region definition from file and specify scaling factor for relevant regions. Use 0 as factor for irrelevant regions.
- Add more validation checks for input localisation configuration file.


oddvarlia:
A new PR is created for this feature and replaces this one. So it is closed

PR #345, opened 2021-09-09
Add documentation for first version

oddvarlia:
Add user documentation with example for the first version of the LOCALISATION_JOB.

### DIFF
--- a/tests/jobs/localisation/test_valid_rst.py
+++ b/tests/jobs/localisation/test_valid_rst.py
@@ -1,0 +1,12 @@
+from semeio.workflows.localisation.local_config_script import DESCRIPTION, EXAMPLES
+import rstcheck
+import pytest
+
+
+@pytest.mark.parametrize("rst_text", [DESCRIPTION, EXAMPLES])
+def test_valid_rst(rst_text):
+    """
+    Check that the documentation passed through the plugin system is
+    valid rst
+    """
+    assert not list(rstcheck.check(rst_text))
--- a/semeio/workflows/localisation/local_config_script.py
+++ b/semeio/workflows/localisation/local_config_script.py
@@ -39,7 +39,182 @@ class LocalisationConfigJob(SemeioScript):
         )
 
 
+DESCRIPTION = """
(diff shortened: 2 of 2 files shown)


tests/jobs/localisation/test_valid_rst.py
```diff
@@ -0,0 +1,11 @@
...
+import rstcheck
+
+
+def test_valid_rst():
+    """
+    Check that the documentation passed through the plugin system is
+    valid rst
+    """
+    assert not list(rstcheck.check(DESCRIPTION))
+    assert not list(rstcheck.check(EXAMPLES))
```
oyvindeide, line 11: Just a tiny thing, but think this makes a lot of sense as parameterization:
```suggestion
@pytest.mark.parameterize("rst_text", [DESCRIPTION, EXAMPLES])
def test_valid_rst(rst_text):
    """
    Check that the documentation passed through the plugin system is
    valid rst
    """
    assert not list(rstcheck.check(rst_text))
```

oyvindeide (commented):
This looks good, just a tiny comment. You dont want to use the automatic generation of schema? Was it something that did not work properly? It was made for another case, so might very well be.

tests/jobs/localisation/test_valid_rst.py
oddvarlia: The autogenerated doc from the structure defined by pydantic was not easy to understand since it refers to variables (keys) not used in the config file as well. The best user documentation is a documentation showing examples of config files with explanations and without any irrelevant info that is not part of the config file. E.g. it is not very clear for me to setup any config file for some of the other workflows as well. I have problems understanding how to setup a config file for e.g the MISFIT_PREPROCESSOR using the documentation there. It lacks extensive examples showing configuration files and the autogenerated doc from pydantic was not clear for me.

oddvarlia: I will update the test as you suggest

oyvindeide (approved):
LGTM!

Issue #306, opened 2021-03-10
ERT does not run on OSX for semeio forward models


berland:
**Describe the bug**
Semeio CI on github actions cannot run ERT on OSX due to case insensitive file system

**To Reproduce**
Run #304 on MacOSX

**Expected behavior**
Tests to pass

**Additional context**

https://github.com/equinor/semeio/pull/304#issuecomment-789533300

oyvindeide:
There is this workaround for your PR, but does nothing to fix the underlying problem: https://github.com/equinor/semeio/blob/15565c26e02c791954468c8a9982cadd45ffb0a5/tests/test_ert_integration.py#L24

berland:
I am fine with both a workaround and a proper fix.

PR #352, opened 2021-10-04
Rename job configuration files

lars-petter-hauge:
Solves #306 

libres will look for an executable in the same folder as the job
configuration file is located. If the name of the config is the same as
the name of the executable, libres will be confused. The usual standard
in ERT would be to capitalize the config file. On OSX systems, which are
case-insensitive, this isn't viable. The job config files are therefore
appended with "_CONFIG". The jobs will be installed as JOB_NAME
JOB_NAME_CONFIG, and the JOB_NAME_CONFIG will point to an executable
named job_name - which we install with entry-points. The user can use
the forward model job as normal: SIMULATION_JOB JOB_NAME


lars-petter-hauge:
New issues spawned from this PR: https://github.com/equinor/semeio/issues/354

oyvindeide (approved):
LGTM! Thanks for fixing, was quite annoying to run the tests on mac before this.

PR #356, opened 2021-10-06
Add log level argument to OTS

lars-petter-hauge:


lars-petter-hauge:
I didn't add any tests for this.. what we really wanted to test (ref issue referenced), is that log statements are passed to stdout file of ert.. I could add an integration test, but I can't make a new job, cause it has to be included in the setup.py to be inline with the rest.. so it would be a bit awkward

PR #363, opened 2021-10-28
Localisation: Added option to specify GEN_OBS nodes in the form nodename:index in localisation config file

oddvarlia:
To be able to specify localisation were individual observations within an obs node of type GEN_OBS is used, it can be specified in the localisation config file in the same way as model parameters of type GEN_PARAM  with nodename:index

To avoid problems mixing observation nodes of type GEN_OBS with notation:  *nodename* and *nodename:index* a keyword is defined to specify if individual GEN_OBS observations is allowed to be specified or not. Since a GEN_OBS node specified in the ERT config file can be expanded into N individual observations, a GEN_OBS node is expanded into a list of N observation names: *[ nodename:0, nodename:1, ....,nodename:N-1]*.
The user will then have to specify observations of type GEN_OBS in the form *nodename:index*.

But for some nodes of type GEN_OBS, like seismic data, the number N can be very large and for those cases it is  best not to specify individual observations, hence, the form *nodename*  is used, not expanded into individual observations.

### DIFF
--- a/tests/jobs/localisation/test_integration.py
+++ b/tests/jobs/localisation/test_integration.py
@@ -265,7 +265,7 @@ def test_localisation(setup_ert, obs_group_add, param_group_add, expected):
 
 
 def test_localisation_gen_kw(setup_ert):
-    ert = EnKFMain(setup_ert, verbose=True)
+    ert = EnKFMain(setup_ert)
     config_dict = {
         "log_level": 4,
         "max_gen_obs_size": 1000,
@@ -345,7 +345,7 @@ def test_localisation_gen_param_and_gen_obs_expanded(
     In this case the GEN_OBS node is expanded and the user specify
     obs like POLY_OBS:index
     """
-    ert = EnKFMain(setup_poly_gen_param_ert, verbose=True)
+    ert = EnKFMain(setup_poly_gen_param_ert)
     fs = ert.getEnkfFsManager().getFileSystem("default_smoother_update")
     ert.getEnkfFsManager().switchFileSystem(fs)
     config_dict = {
@@ -374,7 +374,7 @@ def test_localisation_gen_param_and_gen_obs_not_expanded1(
     max_gen_obs_size < number of obs in the GEN_OBS node.
     """
 
-    ert = EnKFMain(setup_poly_gen_param_ert, verbose=True)
+    ert = EnKFMain(setup_poly_gen_param_ert)
     fs = ert.getEnkfFsManager().getFileSystem("default_smoother_update")
     ert.getEnkfFsManager().switchFileSystem(fs)
     config_dict = {


.pre-commit-config.yaml
```diff
@@ -1,27 +1,34 @@
...
+        name: pylint
+        language: system
+        types: [python]
+        entry: "python -m pylint"
+exclude: tests/jobs/localisation/snapshots/
+#exclude:   |
+#        (?x)^(
+#        tests/jobs/localisation/snapshots/ |
+#        tests/legacy_test_data/poly_gen_param/
+#        )$
```
oyvindeide, line 34: Comments left over from debugging? Also, are the other diffs in this file needed? Does not look like they do anything?

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -714,12 +721,38 @@ def add_ministeps(
...
             debug_print(
-                f"Add obs node: {obs_name}", LogLevel.LEVEL2, user_config.log_level
+                f"Add obs node: {obs_node_name}", LogLevel.LEVEL2, user_config.log_level
             )
-            obs_group.addNode(obs_name)
+            if obs_node_name in key_list_gen_obs:
+                # An observation node of type GEN_OBS
+                if len(obs_index_list) > 0:
+                    active_obs_list = obs_group.getActiveList(obs_node_name)
+                    if len(obs_index_list) > 50:
```
oyvindeide, line 739: Think this should be in the documentation instead? 50 is a very arbitrary number, and might not hold over time.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -714,12 +721,38 @@ def add_ministeps(
...
+                        debug_print(
+                            f"More than 50 active obs for {obs_node_name}",
+                            LogLevel.LEVEL3,
+                            user_config.log_level,
+                        )
+
+                    for string_index in obs_index_list:
+                        index = int(string_index)
+                        if len(obs_index_list) <= 50:
```
oyvindeide, line 748: Same comment as above, this should probably be in the documentation instead.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -744,6 +777,155 @@ def add_ministeps(
         ert_local_config.getUpdatestep().attachMinistep(ministep)
 
 
+def get_corr_group_spec(correlations_spec_list, name):
+    found = False
+    corr_spec = None
+    for count, corr_spec in enumerate(correlations_spec_list):
```
oyvindeide, line 783: Why are you doing `enumerate`here? `count` does not appear to be used?

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -744,6 +777,155 @@ def add_ministeps(
         ert_local_config.getUpdatestep().attachMinistep(ministep)
 
 
+def get_corr_group_spec(correlations_spec_list, name):
+    found = False
+    corr_spec = None
```
oyvindeide, line 782: Dont need to set this to `None`, if it is not `found` this is not used. The exception is if you would like to get rid of `found` and just to a check for `if not corr_spec` instead. That might be better, assuming there can not be `None` in `correlations_spec_list`

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -744,6 +777,155 @@ def add_ministeps(
...
+    for count, corr_spec in enumerate(correlations_spec_list):
+        corr_group_name = corr_spec.name
```
oyvindeide, line 784: Dont need to store this to a variable, just do: `if corr_spec.name == name`

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -744,6 +777,155 @@ def add_ministeps(
...
+            found = True
+            break
+    if not found:
+        raise ValueError(
+            f"Can not find correlation group: {name} in user specification."
+        )
+    return corr_spec
+
+
+def verify_ministep_active_param(
```
oyvindeide, line 795: Thought this could be removed?

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -744,6 +777,155 @@ def add_ministeps(
...
+                        f" For ministep: {ministep.name()} and "
+                        f"parameter node: {node_name}:\n"
+                        "Mismatch between specified active parameters "
+                        f"and active parameters in the ministep.\n"
+                        f"Specified: {spec_index_list}\n"
+                        f"In ministep: {active_index_list}\n"
+                    )
+
+
+def verify_ministep_active_obs(corr_spec_list, ert):
```
oyvindeide, line 874: Same comment as `verify_ministep_active_param`

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -744,6 +777,155 @@ def add_ministeps(
         ert_local_config.getUpdatestep().attachMinistep(ministep)
 
 
+def get_corr_group_spec(correlations_spec_list, name):
```
oyvindeide, line 780: No test for this function

.pre-commit-config.yaml
oddvarlia: Yes, can reset this file back to its original before my edits in this commit. How to do that in an easy way?

semeio/workflows/localisation/local_script_lib.py
oddvarlia: This number (50) only influences the log output. I don't want printing the full obs_index_list if the length of the list is too large. This number has no effect on any calculations or use of any data. So I think it is OK to hard code it here. 50 is arbitrary (roughly a full screen...)

semeio/workflows/localisation/local_script_lib.py
oddvarlia: I think this is not important to document since it only influence the number of lines of the log output for cases with more than 50 observations. And the message says exactly what is the case.

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Agree. Fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: I keep the found variable and remove the corr_spec = None assignment.

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Yes, fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: They are used in test_integration.py   I will move them to that file? They are used in check_consistency_for_active_param_and_obs in test_integration.py

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Is moved to test_integration.py

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Is only used in the test_integration.py functions.

.pre-commit-config.yaml
oyvindeide: This should work: `git checkout upstream/main .pre-commit-config.yaml`

oddvarlia: OK (except that I have called the remote for origin instead of upstream)

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -853,3 +878,31 @@ def write_qc_parameter(
                 grid.write_grdecl(scaling_kw, file)
             # Increase parameter number to define unique parameter name
             cls.scaling_param_number = cls.scaling_param_number + 1
+
+
+def get_obs_from_ert(ert, expand_gen_obs_max_size):
```
oyvindeide, line 883: Is it possible to avoid the special casing of `GEN_OBS`?

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -712,20 +717,40 @@ def add_ministeps(
...
-                f"Add obs node: {obs_name}", LogLevel.LEVEL2, user_config.log_level
+                f"Add obs node: {obs_node_name}", LogLevel.LEVEL2, user_config.log_level
             )
-            obs_group.addNode(obs_name)
+            if obs_node_name in key_list_gen_obs:
+                # An observation node of type GEN_OBS
+                if len(obs_index_list) > 0:
+                    active_obs_list = obs_group.getActiveList(obs_node_name)
+                    if len(obs_index_list) > 50:
+                        debug_print(
```
oyvindeide, line 736: I would still argue that this should be a documentation issue instead.

semeio/workflows/localisation/local_script_lib.py
oyvindeide: If not we should add some testing of it

oddvarlia: The reason why I defined this 'expand_gen_obs_max_size' was to avoid having to expand an GEN_OBS node into the list of type  nodename: index for index from 0 to size where the size sometimes can be very large (seismic data with 100.000's of values. The use of individual GEN_OBS observations is primarily meant for GEN_OBS nodes of moderate size. Examples I was thinking about was e.g geophysical data like gravimetry measurements or other cases where the number of observations in a GEN_OBS node is < 100. In this case it can be meaningful to setup localisations between individual observations and model parameters.  For seismic data, the obvious way of handling that is to split the seismic data into a moderate number of groups (e.g one per fault block or segment) and define one GEN_OBS node for each of these and use these without referring to individual observations within the GEN_OBS node. Sometimes, like for isolated segments, it can be relevant to setup localisation where a GEN_OBS node for a segment is used together with the model parameter values related to the same segment. I thought at least that there might be unnecessary overhead to work with lists of 100.000's of observations in the localisation script and introduced this expand_gen_obs_max_size parameter to handle ERT models where there exists GEN_OBS nodes having large number of observations and GEN_OBS nodes having moderate number or small number of observations ( < expand_gen_obs_max_size).  Do you have any better idea here?

semeio/workflows/localisation/local_script_lib.py
oddvarlia: You mean that the user should set the number, hardcoded to 50, or that I should drop the debug print out here? I can of course write a sentence in the user doc, but I don't think the users care much about this as long as the output is not too large. Maybe I have misunderstood something?

oddvarlia:
This PR is waiting for some use cases before we merge it into main branch

ManInFez:
Is this a draft or is it ready for review?

berland:
Closing due to no traction. Feel free to reopen whenever there is something new to add to this.

PR #365, opened 2021-11-07
Test for localisation for number of active parameters

oddvarlia:
Improve test coverage of number of active parameters defined per ministep in localisation setup.
Requires that PR for adding python function ActiveList.getActiveIndexList() is accepted in ert repo


oddvarlia:
Improved tests to check that what is specified to be active observations and parameters also are represented as active in the created ministeps. These will be included in another PR.

PR #371, opened 2021-11-12
Remove mention of removed ERT keyword CUSTOM_KW

berland:
CUSTOM_KW has been gone for a long time.

Its presence causes warnings when running tests, and the documentation for `DESIGN2PARMAMS` was wrong.


sondreso (approved):
LGTM!

PR #377, opened 2021-11-22
Modified localisation config file for ref_point keyword

oddvarlia:
Small modification of config file: Keyword ref_point is now put under keywords field_scale and surface_scale and only relevant for method: gaussian_decay and exponential_decay. Reference point data are now specified in class GaussianConfig.

The argument for doing this is:
1. Reference point is only relevant for scaling functions that actually use this, e.g. distance based localisation using gaussian_decay and exponential_decay. It is not relevant if scaling functions is defined outside ERT and imported or defined by segments.
2. There were no check telling the user that keyword ref_point was irrelevant for other methods than gaussian_decay, exponential_decay. 
3. The examples in documentation shows cases where ref_point was used but was irrelevant. This is fixed.

The original argument was that ref_point is logically related to observations, not to parameters, but it is relevant to say that reference point can be specified at other locations than at observation locations. For an injector / producer pair, the reference point should probably be located such that the area between the injector/producer pair where the sweep is most dominant is a good place for having the reference point to ensure that the correlations between field parameters an observations are not reduced too much in this area.

### DIFF
--- a/semeio/workflows/localisation/local_config_script.py
+++ b/semeio/workflows/localisation/local_config_script.py
@@ -131,7 +131,7 @@ with ``aps_``. ::
           main_range: 1700
           perp_range: 850
           azimuth: 310
-       ref_point: [463400, 5932915]
+          ref_point: [463400, 5932915]
 
     - name: CORR2
        obs_group:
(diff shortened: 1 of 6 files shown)


oyvindeide (commented):
Changes looks fine, I seem to remember us discussing this location of the configuration before, but am not 100% sure. Did we? If so, do you remember the pros and cons? Can you also but the reason for moving it into the commit text?

oyvindeide (approved):
LGTM!

PR #378, opened 2021-11-23
Update documentation of localisation job

oddvarlia:
Added info about requirement to avoid double specification of a pair of observation and model parameter.


oddvarlia:
Will include the doc update as a parto of another pull request

Issue #375, opened 2021-11-18
[GENDATA_RFT] - report_step as column in csv-output


tnatt:
**Is your feature request related to a problem? Please describe.**
When merging rft observations with simulated rft using `merge_rft_ertobs` from [subscript](https://github.com/equinor/subscript/tree/master/src/subscript/merge_rft_ertobs), a bug occurs if there is more than one report_step for a well https://github.com/equinor/subscript/issues/455. For these cases we need not only to merge observed and simulated rft's on well and order, but also on the time information -> which is easiest done using the report_step integer. 

**Describe the solution you'd like**
I would like the report_step added as a column in the csv output from gendata_rft, to be able to use this information when connecting observed and simulated rft using `merge_rft_ertobs`.

The report_step could be assigned here together with well and time:
https://github.com/equinor/semeio/blob/69230292b27acb60ef4c65fd2e18a5c365b89ae0/semeio/jobs/rft/gendata_rft.py#L178-L183

berland:
Solved in #379

PR #379, opened 2021-11-25
Add report step to the outputted CSV file

berland:
Solves #375 

~~Testing is limited, as semeio does not have a test dataset where the RFT binary file contains the same well at multiple timesteps.~~


semeio/jobs/rft/gendata_rft.py
```diff
@@ -182,6 +182,8 @@ def run(
                 well=well, time=time
```
lars-petter-hauge, line 182: How about just adding it as another entry here? 
`.assign(well=well, time=time, report_step=report_step)`

lars-petter-hauge:
Feel free to also squash into single commit, I believe these two can go together :)

semeio/jobs/rft/gendata_rft.py
berland: yup! That is neater.

lars-petter-hauge (approved):

PR #382, opened 2021-12-02
Support relevant xlrd+pandas version combinations

berland:
Not pretty, but aim to remove xlrd support in the future.

An xls file has not been added to the test data, so for now, a manual test with both xls data and pip installing of different xlrd/pandas combinations are needed to verify the PR.


berland:
Superseded by #383

Issue #380, opened 2021-12-02
Remove pinning of xlrd < 2


berland:
We should remove this pinning to allow for progress. xlrd at version 2 brings 3.8 compatibility, and removes `.xls`-support, which is better handled elsewhere.

https://github.com/equinor/semeio/blob/c9bfe8d08b535a59669d68fa0c26364532efeea3/setup.py#L59

berland:
Superseded by #383

PR #383, opened 2021-12-03
Handle xlrd upgrade

lars-petter-hauge:
solves #380 

Manually verified that all tests pass with `xlrd == 1.2.0`

Otherwise we will get `pandas==1.1.5` in our pipeline (with py36) and `pandas==1.3.4` (with py38)
together with `xlrd == 2.0.1`

### DIFF
--- a/semeio/jobs/design2params/design2params.py
+++ b/semeio/jobs/design2params/design2params.py
@@ -1,11 +1,15 @@
 # pylint: disable=logging-fstring-interpolation
+import logging
+import warnings
+from pathlib import Path
+
 import numpy as np
 import pandas as pd
-
-import warnings
-import logging
+from xlrd.biffh import XLRDError
 
 from semeio._exceptions.exceptions import ValidationError
+
+warnings.filterwarnings("default", category=DeprecationWarning, module="semeio")
 
 # Filenames created/modified by this script.
 # Don't change unless you understand the consequences.
@@ -176,7 +180,7 @@ def _complete_parameters_file(
         target_file.write("OK\n")
 
 
-def _read_excel(file_name, sheet_name, header=0):
+def _read_excel(file_name, sheet_name, header=0, engine=None):
     """
     Make dataframe from excel file
     :return: Dataframe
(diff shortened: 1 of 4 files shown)


tests/jobs/design2params/test_design2params.py
```diff
@@ -182,6 +182,20 @@ def test_three_column_defaults(tmpdir):
     assert (parsed_defaults.columns == ["keys", "defaults"]).all()
 
 
+@pytest.mark.parametrize("ext, engine", [("xls", "xlwt"), ("xlsx", None)])
```
berland, line 185: Using `xlwt` as an engine is deprecated according to Pandas docs, so this test will have to be changed at some point.
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_excel.html

lars-petter-hauge: Yepp, pandas suggests that we use `to_excel` without specifying engine and let them figure out what to use based on extension. However that isn't possible in earlier versions of pandas.

berland: The code also supports OpenOffice spreadsheets. We can add that to these tests by the tuple `("ods", "odf")` and adding `odfpy` to `test_requirements.txt`.

lars-petter-hauge: I'll see how that works with pandas 1.1.5 (might need to catch another error and specify engine again)

semeio/jobs/design2params/design2params.py
```diff
@@ -196,6 +207,14 @@ def _read_excel(file_name, sheet_name, header=0):
             )
         ) from err
 
+    p = Path(file_name)
+    if p.suffix == ".xls":
```
berland, line 211: Always dangerous to depend on file suffices, but it should be ok in this context.

test_requirements.txt
```diff
@@ -5,3 +5,5 @@ pytest-console-scripts
 pytest-httpserver
 pytest-snapshot
 hypothesis
+xlwt
+odfpy
```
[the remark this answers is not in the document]
lars-petter-hauge: Done, no test on it though so it will become unsorted eventually

semeio/jobs/design2params/design2params.py
lars-petter-hauge: I agree, and `xlrd` even has an `inspect_format`. That is not available in version 1.2.0 though, would need to invert the pinning of `xlrd` in the same commit then

berland (approved):
LGTM

berland:
Passes manual test matrix with py=3.6, 3,8, pd=1.1.5,1.2.0,1.2.2,1.3.3 and xlrd=1.2.0,2.0.1

Issue #384, opened 2021-12-03
Remove warning on unsupported defaultvalues format


berland:
The alternative format for the default values predates semeio by many years, it can be purged.

https://github.com/equinor/semeio/blob/c9bfe8d08b535a59669d68fa0c26364532efeea3/semeio/jobs/design2params/design2params.py#L290-L298

berland:
The warning is misleading, as supplying defaults in a transposed version has never worked in semeio.

PR #385, opened 2021-12-03
Remove deprecation warning from defaults-parsing

berland:
Solves #384 

Ignore anything beyond the second column. The reason for the deprecation
warning was historic as there previously (pre-semeio) existed a version
of design2params that read the defaults in a transposed manner. Such
spreadsheets have probably not been in use since 2015, and it should be
safe now just to ignore these columns and continue as before.


lars-petter-hauge (approved):
LGTM

PR #392, opened 2022-01-06
Adapting to new API after local_dataset was removed

### LINKED ISSUE #391: Make ahm_analysis and localisation compatible with main version of ert
There is a breaking change in ministeps in https://github.com/equinor/ert/pull/2645 which affects ahm_analysis and localisations workflows. We should adjust to this.

BjarneHerland:
Fixes #391


ertomatic:

oyvindeide:
Looks good to me, can you check that everything works for you @oddvarlia?

oyvindeide (approved):

PR #395, opened 2022-01-19
Add replace string job using sed

### LINKED ISSUE #312: Bringing REPLACE back
The `REPLACE` job was abandoned some time ago, but appears too still have some usage for very small replacement tasks. Due to this it is suggested by @eivindsm to reintroduce it. However, I suggest that instead of reintroducing the previous one we make an alternative powered by `sed`.

Something along the lines of
```
EXECUTABLE sed
ARGLIST -i `s/<FROM>/<TO>/g` <FILE>
```
should be sufficient.

As this is really hard to unit test, I suggest that we add the workflow to the HM tutorial version that does integration testing for most of the jobs in libres + semeio

ManInFez:
resolves #312

### DIFF
--- a/semeio/jobs/config_jobs/REPLACE_STRING_CONFIG
+++ b/semeio/jobs/config_jobs/REPLACE_STRING_CONFIG
@@ -1,0 +1,2 @@
+EXECUTABLE sed
+ARGLIST -i s/<FROM>/<TO>/g <FILE>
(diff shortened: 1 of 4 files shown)


dafeda (approved):
Clever way to test this, LGTM 👍

ManInFez:
Test this please!

PR #398, opened 2022-01-24
Add simple WYSIWYG replace_string script

### LINKED ISSUE #312: Bringing REPLACE back

ManInFez:
resolves #312

### DIFF
--- a/semeio/jobs/scripts/replace_string.py
+++ b/semeio/jobs/scripts/replace_string.py
@@ -1,8 +1,13 @@
+import argparse
+
+from pathlib import Path
+
+
 description = """
-Performs inplace string replacement in a file using sed.
+Performs inplace string replacement in a files.
 
 ARGS
-FROM: regex to match what to replace
+FROM: string to match what to replace
 TO: expression to replace the match
 FILE: name of file to perform replacement in.
 
@@ -10,6 +15,39 @@ Example:
 FORWARD_MODEL_JOB REPLACE_STRING(<FROM>=something, <TO>=else, <FILE>=file.txt)
  > replace all something to else in file.txt
 
+Due to ERT particular way of parsing config files, there are some curiosities
+To escape characters like , use '', however this misses white space
+To escape whitespace use ""
+
 """
 
 category = "utility.templating"
+
+
+def _get_args_parser():
+
+    parser = argparse.ArgumentParser(description=description)
+
+    parser.add_argument(
+        "-o",
+        "--original",
+        help="Original text",
+    )
+    parser.add_argument(
+        "-n",
+        "--new",
+        help="New text",
+    )
+    parser.add_argument(
+        "-f",
+        "--file",
+        help="File to where replacement should occur",
+    )
+    return parser
+
+
+def main_entry_point():
+    parser = _get_args_parser()
+    options = parser.parse_args()
+    file = Path(options.file)
+    file.write_text(file.read_text().replace(options.original, options.new))
--- a/tests/jobs/replace_string/test_integration.py
+++ b/tests/jobs/replace_string/test_integration.py
@@ -1,4 +1,3 @@
-import sys
 import subprocess
 import pytest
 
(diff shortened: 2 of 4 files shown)


semeio/jobs/scripts/replace_string.py
```diff
@@ -13,3 +18,34 @@
...
+    )
+    return parser
+
+
+def main_entry_point():
+    parser = _get_args_parser()
+    options = parser.parse_args()
+    file = Path(options.file)
+    file.write_text(
+        file.read_text().replace(options.original.strip("'"), options.new.strip("'"))
```
asnyv, line 50: Is it necessary to strip away `'`? Very common character in Eclipse files, which I would think it is likely is one of the main use cases for this job.

tests/jobs/replace_string/test_integration.py
```diff
@@ -8,16 +7,21 @@
...
+        (
+            "I got some/thing to tell you",
+            "some/thing",
+            "a story",
+            "I got a story to tell you",
+        ),
+    ],
 )
 def test_replace_string(tmpdir, input_text, replace_from, replace_to, expected):
     with tmpdir.as_cwd():
```
oyvindeide, line 27: You should be able to just call `main_entry_point` directly now, and not have to do `subprocess`

ManInFez: The whole point as to test it from ert, since ert does funny things to both ert-config and when parsing args...

ManInFez:
Test this please!

semeio/jobs/scripts/replace_string.py
ManInFez: I guess it is not necessary, removed the stripping :)

ManInFez:
Anything else @oyvindeide @asnyv?

asnyv:
> Anything else @oyvindeide @asnyv?

Think it looks alright from my side.

tests/jobs/replace_string/test_integration.py
```diff
@@ -8,16 +7,21 @@
 QUEUE_SYSTEM LOCAL
 NUM_REALIZATIONS 1
 FORWARD_MODEL COPY_FILE(<FROM>=<CONFIG_PATH>/file.txt, <TO>=file.txt)
-FORWARD_MODEL REPLACE_STRING(<FROM>={FROM}, <TO>={TO}, <FILE>=file.txt)
+FORWARD_MODEL REPLACE_STRING(<FROM>={FROM}, <TO>="{TO}", <FILE>=file.txt)
```
oyvindeide, line 10: There is only quotes on `"{TO}"`, I am sure there is a reason for that, and that reason might be very well suited for the documentation for this job?

tests/jobs/replace_string/test_integration.py
```diff
@@ -8,16 +7,21 @@
...
 
 
-@pytest.mark.skipif(
-    sys.platform == "darwin", reason="Skip test for Mac OS - invalid use of sed"
-)
 @pytest.mark.parametrize(
     "input_text, replace_from, replace_to, expected",
-    [("something", "something", "else", "else"), ("1,2,3,4", "[0-9]", "N", "N,N,N,N")],
+    [
+        ("something", "something", "else", "else"),
```
oyvindeide, line 17: Could you also add an example for multiple occureses of the same sub-string?

oyvindeide (commented):
Think this looks good, but could maybe add some of the outside context in the documentation of the job. Would also be nice with a test making sure the documentation is valid rst as that is not checked anywhere, could be done with:
```python
assert not list(rstcheck.check("my rst string here"))
```

tests/jobs/replace_string/test_integration.py
ManInFez: This is due to the general problem with ERT and how it escapes strings and how it is parsing arguments for forward model jobs.

tests/jobs/replace_string/test_integration.py
ManInFez: :+1:

ManInFez: I changed the first example to contain multiple. The integration test is pretty slow

tests/jobs/replace_string/test_integration.py
oyvindeide: Would be great to have that in the documentation as it is a common question.

ManInFez: done!

ManInFez:
Test this please!

PR #403, opened 2022-02-18
Extend functionality for gaussian_decay and exponential_decay functions in localisation job

oddvarlia:
Extend the gaussian_decay and exponential_decay functions in localisation to be more flexible.
Want to be able to have scaling factor equal to 1 within the range and a decaying scaling factor at distance larger than 1 range away from the reference point. Also introduce the possibility to set the scaling function to 0 outside the range.
New alternative versions:
f(d) = 1 if d <= 1
f(d) = exp(-3(d-1)/(dt-1))  for d > 1
and optionally 
f(d) = 0 if d > dt

Here d is normalised distance d= sqrt((dx/rx)^2 + (dy/ry)^2) where rx and ry are two range parameters,
dt is a user specified parameter for normalised distance.
The new alternative function is 1  for normalised distance up to d= 1, decays down to about 0.05 at normalised distance dt > 1 where the value of the function is about 0.05 and decays further to 0 at infinite distance unless the cutoff is activated.


semeio/workflows/localisation/local_script_lib.py
```diff
@@ -788,19 +823,52 @@ def get_dx_dy(self, data_index):
         ) / self.perp_range
         return dx, dy
 
+    def norm_dist_square(self, data_index):
+        dx, dy = self.get_dx_dy(data_index)
+        d2 = dx * dx + dy * dy
```
oyvindeide, line 828:
```suggestion
        d2 = dx**2 + dy**2
```

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -788,19 +823,52 @@ def get_dx_dy(self, data_index):
...
+
+            distance_from_inner_ellipse = (d - 1) / (self.normalised_tapering_range - 1)
+            exp_arg = -3 * distance_from_inner_ellipse * distance_from_inner_ellipse
+            return math.exp(exp_arg)
+
+        else:
+            if self.use_cutoff and d2 > 1.0:
+                return 0.0
+            exp_arg = -3.0 * d2
+            return math.exp(exp_arg)
```
oyvindeide, line 850: The functionality here looks good to me, but could it be an idea to introduce it as separate classes instead? For example:
```
class CutoffExpontentialDecay(GaussianDecay)
```
my main reason for suggesting that is that this has a lot of conditionals, which complicates things quite a bit for the user, for example if `use_one_in_range` if `False`, `normalised_tapering_range` is not used. By splitting this in a few different data models we can make it clearer which options do what.

oyvindeide (commented):
I think the changes look good, but have a question that relates to user configuration.

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Rewritten using one class per method.

oddvarlia:
Have updated with suggestions from code review

oyvindeide (approved):

PR #412, opened 2022-03-18
Remove non-functional test giving pytest warning

berland:
The intention of the removed test is already covered by
test_semeio_script.py::test_semeio_script_concurrent_logging


berland:
The file is indeed in use, and the warning seems gone.

PR #415, opened 2022-03-22
Rename file to avoid being collected by pytest

berland:
Pytest complains when it collects it, and it is not supposed to be collected.


berland:
Warning was:
```
tests/communication/test-data/test_workflow_job.py:7
  /Users/HAVB/projects/semeio/tests/communication/test-data/test_workflow_job.py:7: PytestCollectionWarning: cannot collect test class 'TestWorkflowJob' because it has a __init__ constructor (from: tests/communication/test-data/test_workflow_job.py)
    class TestWorkflowJob(SemeioScript):
```

PR #416, opened 2022-03-28
Update decay functions

oyvindeide:


oddvarlia:
This update is already included by another PR and can be removed.

PR #418, opened 2022-04-05
Remove skipping tests on darwin, bug is fixed

berland:


sondreso (approved):
Lint fails, but LGTM otherwise!

PR #420, opened 2022-04-29
Update test

lars-petter-hauge:
Ert has recently changed from continuing a run when there is no ensemble variations, to stopping (as it should..).

We are mostly concerned about checking that the output from gen_data_rft does not generate an error when it's loaded. It would be better to check that it is properly included though..


berland (approved):
ok!

PR #423, opened 2022-05-03
Update design_kw to accept template files of type xml

oddvarlia:
Purpose: Sometimes it is useful to apply scripts that have xml files as model files. A template file for an xml model file has lots of keywords in <> parenthesis. But validation check in design_kw.py will fail for template files for xml model files. An exception where errors are replaced by warnings is implemented for template files of type xml.

### DIFF
--- a/semeio/jobs/design_kw/design_kw.py
+++ b/semeio/jobs/design_kw/design_kw.py
@@ -3,6 +3,8 @@ import logging
 import shlex
 import re
 
+from typing import List
+
 
 _STATUS_FILE_NAME = "DESIGN_KW.OK"
 
@@ -61,6 +63,13 @@ def all_matched(line, template_file_name, template):
                     f"but this is probably a Perl file"
                 )
             )
+        elif is_xml(template_file_name, template):
+            _logger.warning(  # pylint: disable=logging-fstring-interpolation
+                (
+                    f"{unmatched} not found in design matrix, "
+                    f"but this is probably an xml file"
+                )
+            )
         else:
             _logger.error(  # pylint: disable=logging-fstring-interpolation
                 f"{unmatched} not found in design matrix"
@@ -71,6 +80,10 @@ def all_matched(line, template_file_name, template):
 
 def is_perl(file_name, template):
     return file_name.endswith(".pl") or template[0].find("perl") != -1
+
+
+def is_xml(file_name: str, template: List[str]):
+    return file_name.endswith(".xml") or template[0].find("?xml") != -1
 
 
 def unmatched_templates(line):
--- a/tests/jobs/design_kw/test_design_kw.py
+++ b/tests/jobs/design_kw/test_design_kw.py
@@ -190,19 +190,54 @@ def test_is_perl():
     assert not design_kw.is_perl(file_name, template)
 
 
+def test_is_xml():
+    file_name = "filename.yml"
+    template = ["?xml", "  ECLIPSE_INIT_DATE: <DATE>"]
+    assert design_kw.is_xml(file_name, template)
+
+    file_name = "filename.xml"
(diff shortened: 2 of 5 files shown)


semeio/jobs/design_kw/design_kw.py
```diff
@@ -61,6 +61,13 @@ def all_matched(line, template_file_name, template):
                     f"but this is probably a Perl file"
                 )
             )
+        elif is_xml(template_file_name, template):
+            _logger.warning(  # pylint: disable=logging-fstring-interpolation
+                (
+                    f"{unmatched} not found in design matrix, "
+                    f"but this is probably a xml file"
```
berland, line 68: (I think I would say `an xml file`)

berland:
There should be a test analog to the perl test as well.

semeio/jobs/design_kw/design_kw.py
```diff
@@ -73,6 +80,10 @@ def is_perl(file_name, template):
     return file_name.endswith(".pl") or template[0].find("perl") != -1
 
 
+def is_xml(file_name, template):
+    return file_name.endswith(".xml") or template[0].find("xml") != -1
```
berland, line 84: the `template[0].find("xml")` I guess is copied from `is_perl()` where it is used as a heuristic to try to catch  for example
the shebang in a perl file (that would have made a better example in the test code for perl). Is there a similar probability for the string "xml" in the first line in an xml-file?  Could we risk ruining something where the letters `xml` is found on the first line? (no not really, since this is only changing from error to warning)

semeio/jobs/design_kw/design_kw.py
oddvarlia: OK. I also added some tests.

semeio/jobs/design_kw/design_kw.py
oddvarlia: Usually there is a string of type <?xml version="1.0" ?> in the heading. Should I check on ?xml instead maybe ? It could be useful that the code identify a file as xml file also from the top string and not only from the extension .xml What do you think? Use ?xml as search word instead of xml in the heading?

oddvarlia: It works with '?xml' as search word also to identify a file as xml format.

semeio/jobs/design_kw/design_kw.py
```diff
@@ -61,6 +61,13 @@ def all_matched(line, template_file_name, template):
                     f"but this is probably a Perl file"
                 )
             )
+        elif is_xml(template_file_name, template):
```
berland, line 64: Some day in the future, this code should be typed. Can you add simple typing to the function you added, is it 
`template_file_name: str, template: List[str]` ?

berland: (this comment applies to line 83, not this one ...)

oddvarlia: OK fixed

tests/jobs/design_kw/test_design_kw.py
```diff
@@ -190,6 +190,24 @@ def test_is_perl():
     assert not design_kw.is_perl(file_name, template)
 
 
+def test_is_xml():
```
berland, line 193: This only unit-tests the `is_xml` function I see, there should perhaps be some asserts on  `all_matched(...) == True`

oddvarlia: Added xml format files as template files in test

berland (approved):
Good!

tests/jobs/design_kw/test_design_kw.py
oddvarlia: I have added a test that run a case using template model file  where the model file is in xml format.

PR #434, opened 2022-06-13
Remove TARGET_FILE from forward model configs

### LINKED ISSUE #431: Remove usage of TARGET_FILE in job configurations
The TARGET_FILE mechanism for telling the success state of a forward model should be regarded as inferior (deprecate?) to the return code of the shell command that a particular forward model runs.

TARGET_FILE should thus be removed from all semeio jobs. Preferably with tests asserting that ERT is able to pick up return values in a proper manner. 

Relying on TARGET_FILE causes bugs like https://github.com/equinor/ert/issues/3412

berland:
ERT is able to determine failure in a forward model from
the shell exit code. Relying on exit code is more robust
than relying on last modification time of a file on disk, and
no TARGET_FILE is therefore the recommended configuration
of ERT forward models.

Resolves #431


sondreso (approved):
LGTM

PR #436, opened 2022-06-14
Solve Python 3.10 incompatibilites, fix warnings

berland:
A selection of DeprecationWarnings (not only Py310-related)
are also solved.


sondreso (approved):
:three:.:one::zero:

PR #437, opened 2022-06-14
Drop support for Python 3.6 and 3.7

berland:


.github/workflows/publish_to_pypi.yml
```diff
@@ -3,26 +3,26 @@ name: Publish Python 🐍 distributions 📦 to PyPI
...
-    - name: Install dependencies
-      run: pip install --upgrade setuptools setuptools_scm wheel twine
-    - name: Build package
-      run: python setup.py sdist bdist_wheel
-    - name: Upload deploy
-      env:
-        TWINE_USERNAME: __token__
-        TWINE_PASSWORD: ${{ secrets.PYPI_SECRET }}
-      run: python -m twine upload dist/*
+      - uses: actions/checkout@v2
```
kvashchuka, line 13: Not that it is a big of a deal, but why did you change indentation?

semeio/jobs/design2params/design2params.py
```diff
@@ -5,7 +5,6 @@
 
 import numpy as np
 import pandas as pd
-from xlrd.biffh import XLRDError
```
kvashchuka, line 8: It should be deleted in `setup.py` as well, right?

kvashchuka: There is also a comment about `xlrd` here:
https://github.com/equinor/semeio/blob/24bc00bb1755ab420bc4b8e1f1be57c7f4096086/tests/jobs/pyscal/test_pyscal.py#L148

but I am not sure I understand what is meant here. Should it be deleted as well?

.github/workflows/publish_to_pypi.yml
berland: I will blame vscode, which was insistent on fixing this indentation..

setup.py
```diff
@@ -57,7 +55,7 @@
         "ert >= 2.36.0b0",
         "configsuite>=0.6",
         "numpy",
-        "pandas>0.24",
+        "pandas>1.3.0",
```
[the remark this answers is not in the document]
berland: There is an actual dependency in `design2parameters.py` on this.

tests/jobs/design2params/test_design2params.py
```diff
@@ -1,20 +1,20 @@
 import filecmp
 import logging
 import os
-from distutils.dir_util import copy_tree
+import shutil
```
kvashchuka, line 4: Why this changes are part of this PR? I think they both work with Pythin3.6+, do they not?

berland: `distutils` is deprecated in Python 3.10

semeio/jobs/design2params/design2params.py
berland: No, we still need xlrd in order to support old Excel file formats. But with newer Pandas versions we do not need workarounds in semeio in order to support it. If we remove xlrd, the `test_support_multiple_format` for design2params will fail.

The pyscal test will also fail if it is missing, but pyscal declares xlrd as its own dependency.

berland: So we should declare `xlrd` as a direct dependency of `semeio` even though we only use it indirectly through Pandas, but Pandas declares it as optional.

kvashchuka (approved):
LGTM :)

PR #441, opened 2022-07-29
Import MeasuredData through ert

eivindjahren:
The ert import structure is going to change to conform to pep8's
"modules should explicitly declare the names in their public API using
the `__all__` attribute. [...] Other modules must not rely on indirect
access to such imported names unless they are an explicitly documented
part of the containing module's API, such as os.path".


eivindjahren:
Jenkins, test this please!

dafeda (approved):

eivindjahren:
test this please

eivindjahren:
Test this please

PR #451, opened 2022-08-15
Support YYYY-MM-DD in gendata_rft input

berland:
The old format is still supported but a FutureWarning is issued


xjules (approved):
Seems valid! :rocket:

PR #452, opened 2022-08-15
Use ert.shared instead of ert_shared

eivindjahren:


sondreso (approved):
LGTM, just needs the updated ERT version

PR #453, opened 2022-08-16
Move to new ert import structure

eivindjahren:
This uses private import ert._c_wrappers, which should be fixed asap. However, this temporary fix should do for now.


eivindjahren:
Jenkins, test this please!

oyvindeide (approved):
LGTM

PR #459, opened 2022-09-06
Solve and enforce pylint redefined-outer-name

berland:


hnformentin (approved):
LGTM, but I could not understand from fixture documentation why 

```
@pytest.fixture(name="setup_tmpdir")
def fixture_setup_tmpdir(tmpdir):

```
is preferred over

```
@pytest.fixture()
def setup_tmpdir(tmpdir):

```
For me the first one seems to be just more repetition.
Do you understand why lint suggests it?

berland:
This is the basis for the change to how fixtures are named: https://stackoverflow.com/questions/46089480/pytest-fixtures-redefining-name-from-outer-scope-pylint

PR #460, opened 2022-09-06
wip: Solve pylint duplicate-code

berland:
Bump the requirement slightly (from 4 to 6 lines)
and put in exceptions elsewhere.


tests/jobs/localisation/test_methods.py
```diff
@@ -15,6 +15,8 @@
     smooth_parameter,
 )
 
+# pylint: disable=duplicate-code
+
```
hnformentin, line 19: Will those exceptions be applied to the complete file?
If so, would it not be cleaner to add `# pylint: disable=duplicate-code` closer to where the exception should be applied?
Comment valid to the other instances in this PR :-)

berland: Pylint complained about duplicate-code multiple places, then I figured the exceptions I had to add to the code would be less noisy if just placed at the top (even though it then ignores the whole file).

hnformentin: I see, that should not be a big problem for a single file, but let's keep attention if we repeat too much this pattern though.

hnformentin:
pylint error seems to be only an extra line in one of the files.

hnformentin (approved):
LGTM!

PR #461, opened 2022-09-12
Solve pylint unspecified encoding

berland:


hnformentin (approved):
LGTM! I could learn a new way to read data which does not need `with` as context manager: `pathlib.Path`. Very nice 👍

Issue #474, opened 2022-10-31
csv_export2 does not work with dates beyond year 2262


berland:
**Describe the bug**
csv_export2 will crash if it encounters summary files containing dates beyond the year 2262. This is a limitation from fmu-ensemble which again inherits the limitation from pandas.

The bug is part of the documentation, remove when resolved (inside `csv_export2.py`)

**Expected behavior**
No error.

**Workaround** 
Users can use `ECL2CSV` (from ecl2df) as a forward model, and `merge_csv` (subscript) as a workflow.

**Possible resolution**
fmu-ensemble can use  `ecl2df` to parse the summary files, which is able to workaround the 2262-issue in pandas:
See https://github.com/equinor/fmu-ensemble/pull/182 (stale)

Alternatively, fixing https://github.com/equinor/ecl/pull/804 might also suffice. EDIT: It did not.

PR #475, opened 2022-10-31
Document the bug in csv_export2 on year 2262

### LINKED ISSUE #473: Update csv_export2 documentation on dates limitation
https://github.com/equinor/semeio/blob/999b075c26d7494dea41c60b1fd6b9023d3448d4/semeio/workflows/csv_export2/csv_export2.py#L39. Add a comment: Do not support Dates beyond year 2262.

berland:
Resolves #473 

See also issue #474.

### DIFF
--- a/semeio/workflows/csv_export2/csv_export2.py
+++ b/semeio/workflows/csv_export2/csv_export2.py
@@ -42,6 +42,9 @@ combination of wildcards used in this workflow and the actual vectors that are
 requested in the Eclipse DATA file. A wildcard like ``W*`` can in certain cases
 (e.g. Eclipse simulations with 100+ wells) produce thousands of vectors, and can
 then be replaced by something more explicit like ``WOPT* WGPT* WWPT*``.
+
+Dates beyond year 2262 are not supported. Use ecl2csv in combination with
+merge_csv in such scenarios.
 """  # noqa
 
 EXAMPLES = """


xjules (approved):
👍

semeio/workflows/csv_export2/csv_export2.py
```diff
@@ -42,6 +42,9 @@
 requested in the Eclipse DATA file. A wildcard like ``W*`` can in certain cases
 (e.g. Eclipse simulations with 100+ wells) produce thousands of vectors, and can
 then be replaced by something more explicit like ``WOPT* WGPT* WWPT*``.
+
+Dates beyond year 2262 are not supported. Use ecl2csv in combination with
+merge_csv in such scenarios.
```
hnformentin, line 47: Would it be possible to add an example of this use?

hnformentin:
After commenting I noticed it is actually already approved,  but tests are not passing. 
Could you please take a look in my comment and the tests?

semeio/workflows/csv_export2/csv_export2.py
berland: I think it is covered by the examples provided by ecl2df and subscript (as forward models and workflows injected into the ert documentation in fmu-docs.

berland:
I think a new release of ert is needed on pypi for the tests to pass.

PR #477, opened 2022-11-09
Use release candidates of ert in CI testing

berland:
This will allow PR's that depend on release candidates of ert to pass testing.


berland:
Wontfix.

PR #478, opened 2022-11-14
Return correct number of principal components

dafeda:
Code returns one more principal component than it should.
I've added a property test and fixed failing tests.


oyvindeide (approved):
LGTM! Just a conflict that needs to be resolved.

PR #480, opened 2022-11-18
Fix and bump pre-commit configuration

berland:
The url for flake8 now requires login at gitlab. Moved to github


xjules (approved):
LGTM!

Issue #368, opened 2021-11-09
Make gendata_rft find i,j,k faster


berland:
The ecl API for obtaining i,j,k from x,y,z allows for an initial guess when searching for i,j,k. This makes sense to utilize when traversing along a wellpath. The current code performs a full linear search on the grid cells for every trajectory point.

https://github.com/equinor/semeio/blob/69230292b27acb60ef4c65fd2e18a5c365b89ae0/semeio/jobs/rft/gendata_rft.py#L144-L147

https://github.com/equinor/ecl/blob/6296766f0f21238a8d8e434b0aec7adb4c2fd364/python/ecl/grid/ecl_grid.py#L876

PR #483, opened 2022-12-15
Optimize finding ijk from xyz in grid

berland:
Solves #368


berland:
On the dataset in the tests, the modified loop is roughly an order of magnitude faster with this patch.

xjules (approved):
LGTM!

PR #493, opened 2023-01-10
Remove checked in storage

### LINKED ISSUE #492: Dont rely on checked in storage from ERT
This is very susceptible to changes in ERT, and we should perhaps create it on the fly instead.

oyvindeide:
Closes #492


tests/jobs/conftest.py
```diff
@@ -6,17 +6,9 @@
...
 
 @pytest.fixture()
-def setup_ert(tmpdir, test_data_root):
-    cwd = os.getcwd()
-    tmpdir.chdir()
-    test_data_dir = os.path.join(test_data_root, "snake_oil")
-    shutil.copytree(test_data_dir, "test_data")
-    os.chdir(os.path.join("test_data"))
-
+def setup_ert(copy_snake_oil_case_storage):  # pylint: disable=unused-argument
```
ManInFez, line 9: setup_ert sounds like it should at least return ert and not a libresfacade :shrug:

oyvindeide: True, did not change the name though, but added another commit to do that

ManInFez (approved):

PR #494, opened 2023-01-17
Remove check for empty observation vectors

oyvindeide:
If no response can be loaded an error message will be raised from ert, meaning that this check is redundant.


lars-petter-hauge:
So one would never get to this point in the workflow. Quick question, what if one would run the workflow manually before an ensemble has been run?

oyvindeide:
The workflow would fail, but it would do that previously as well, just with a slightly different message, though that message was not really any better.

lars-petter-hauge (approved):

Issue #488, opened 2022-12-22
Non-adaptive localisation fails when using FIELD in ERT config and there are no specified keyword for field_scale for the same correlation group


oddvarlia:
**Describe the bug**
The non-adaptive localisation implemented in the job LOCALISATION_JOB in semeio fails if the ERT model contains fields and localisation config specify  correlation group containing the field parameter but without specifying field_scale keyword in the localisation config. What happens now is that missing specification of row_scaling input for this case makes localisation fail.

**To Reproduce**
Run  the localisation job using a correlation group which uses a field parameter in the parameter group and where no specification of scaling was specified (e.g. missing the keyword **field_scale** in the localisation config file for that correlation group.

**Expected behavior**
The localisation should work also in this case, but with scaling factor equal to 1 for the fields in the correlation group. This means that original estimated correlations between the field parameter values and the observations  in the correlation group should be kept unmodified. What A clear and concise description of what you expected to happen.

oddvarlia:
The purpose was to run with field without having to use any method to modify the correlation by distance or region. By introducing a method **constant** we can achieve this with default scaling = 1.

PR #496, opened 2023-01-19
Fix non-adaptive localisation if keyword field_scale is missing when using fields

oddvarlia:
Fix the issue #488


semeio/workflows/localisation/local_script_lib.py
```diff
@@ -1,4 +1,5 @@
 # pxylint: disable=attribute-defined-outside-init
+# pylint: disable=C0302
```
berland, line 2: can you spell this one out so we don´t have to look it up?

oddvarlia: Will try.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -294,6 +296,42 @@ def apply_decay(
...
+    # pylint: disable=too-many-arguments,too-many-locals
+    """Assign constant value to the scaling factor,
+    assign it to ERT instance by row_scaling
+    and returns a full sized grid parameter with scaling factors for active
+    grid cells and 0 elsewhere to be used for QC purpose.
+    """
+    debug_print(f"Scaling factor is constant: {value}", LogLevel.LEVEL3, log_level)
+    decay_obj = ConstantScalingFactor(value)
+
+    scaling_vector = np.zeros(data_size, dtype=np.float32)
```
dafeda, line 316: This is unfamiliar territory for me, so I'll just start poking around and hopefully eventually understand whats going on.

First question, why do we use np.float32 here?

dafeda (approved):
Do not understand all the details here, but in general it looks fine.

PR #499, opened 2023-02-23
Fix bug where AHM_ANALYSIS would fail if all observations were deactivated

oyvindeide:


berland:
Switch the order of the commits, so that blackening is done first (the second commit touches the lines of the first)

berland (approved):
good to go

PR #501, opened 2023-03-12
Update to new ert api

oyvindeide:


oyvindeide:
The failing test is due to the latest version of ert not being available in `bleeding`, but it is on pypi.

dafeda (approved):

PR #509, opened 2023-04-14
Use GenKwConfig in place of EnkfConfigNode

andreas-el:
Update Semeio to use refactored gen_kw_nodes in Ert.

Ref.
https://github.com/equinor/ert/issues/5267

### DIFF
--- a/semeio/workflows/localisation/local_script_lib.py
+++ b/semeio/workflows/localisation/local_script_lib.py
@@ -14,6 +14,7 @@ from ecl.eclfile import Ecl3DKW
 from ecl.grid.ecl_grid import EclGrid
 from ecl.util.geometry import Surface
 from ert._c_wrappers.enkf.config.field_config import Field
+from ert._c_wrappers.enkf.config.gen_kw_config import GenKwConfig
 from ert._c_wrappers.enkf.config.surface_config import SurfaceConfig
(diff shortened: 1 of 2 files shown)


ertomatic:

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -98,10 +99,8 @@ def get_param_from_ert(ens_config):
         impl_type = node.getImplementationType()
         my_param = Parameter(key, impl_type)
         new_params.append(my_param)
-        if impl_type == ErtImplType.GEN_KW:
-            # Node contains scalar parameters defined by GEN_KW
-            kw_config_model = node.getKeywordModelConfig()
-            my_param.parameters = kw_config_model.getKeyWords().strings
+        if isinstance(node, GenKwConfig):
```
hnformentin, line 102: Why this substitution is not needed in line 675?

hnformentin (approved):
As far as I understood this PR, it looks good to me!

semeio/workflows/localisation/local_script_lib.py
andreas-el: This is because Gen_data is of type EnkfConfigNode, and not GenKwConfig.
They were both EnkfConfigNodes at some point though, but the refactoring separated GEN_KW into its own entity.

```
      elif impl_type == ErtImplType.GEN_DATA:
```

PR #510, opened 2023-05-10
Use IENS and ITER for runpaths

berland:
Also conform to ERT standard for directory naming

Avoids warnings from ERT on usage of %d.


xjules (approved):
Good catch!

PR #512, opened 2023-05-10
Use 32-bit integers due to xtgeo

berland:
Usage of uint16 is a valid optimization for the region parameter, but xtgeo is using numpy with a fill value that overflows in uint16.

As long as xtgeo does not support uint16 explicitly, we need to resort to 32-bits.

https://github.com/equinor/xtgeo/blob/69becba852ace157c8610f9b2edcfb582e7d96f9/src/xtgeo/common/constants.py#L13

This triggers a warning from numpy stating that this condition will lead to an error in the future

### DIFF
--- a/semeio/workflows/localisation/local_script_lib.py
+++ b/semeio/workflows/localisation/local_script_lib.py
@@ -304,7 +304,7 @@ def define_look_up_index(user_defined_active_region_list, max_region_number):
     corresponding to the region number.
     """
     active_segment_array = np.array(user_defined_active_region_list)
-    index_per_used_region = ma.zeros((max_region_number + 1), dtype=np.uint16)
+    index_per_used_region = ma.zeros((max_region_number + 1), dtype=np.int32)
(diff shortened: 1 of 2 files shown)


berland:
```
tests/jobs/localisation/test_integration.py::test_localisation_field2
  /prog/komodo/bleeding-py38-rhel7/root/lib64/python3.8/site-packages/numpy/ma/core.py:467: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of 2000000000 to uint16 will fail in the future.
  For the old behavior, usually:
      np.array(value).astype(dtype)`
  will give the desired result (the cast overflows).
    fill_value = np.array(fill_value, copy=False, dtype=ndtype)
```

kwinkunks (approved):
I take it this is about optimizing memory? Seems reasonable.

These fill values always seem so hacky but I guess they are sometimes necessary.

PR #513, opened 2023-05-10
Resolve pytest warning on PytestUnknownMarkWarning

berland:
When running tests through komodo, pytest claims ert_integration is an unknown mark. This is probably due to pyproject.toml, which defines it, is not present where it is running the tests

### DIFF
--- a/ci/testkomodo.sh
+++ b/ci/testkomodo.sh
@@ -1,6 +1,7 @@
 
 copy_test_files () {
     cp -r $CI_SOURCE_ROOT/tests $CI_TEST_ROOT/tests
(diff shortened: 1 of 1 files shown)


berland:
This commit is educated guesswork. I can't reproduce the issue locally without removing pyproject.toml, so when looking at the testkomodo script I assume the problem must be due to pytest running in CI_TEST_ROOT

hnformentin (approved):
Sorry for coming in your review @kwinkunks 

It seems there is no risk in trying this educated guess. 
Need a rebase though. 

After merged, @berland could you please comment here if the warning is solved?

PR #518, opened 2023-05-19
Solve pylint invalid-name

berland:


berland:
Left as three commits for the sake of review.

The three remaining exceptions in .pylintrc are not something that should be fixed, thus closing the pylinting issue.

.pylintrc
```diff
@@ -135,7 +132,7 @@ include-naming-hint=no
 inlinevar-rgx=[A-Za-z_][A-Za-z0-9_]*$
 
 # Regular expression matching correct method names
-method-rgx=(([a-z][a-z0-9_]{2,30})|(_[a-z0-9_]*))$
+method-rgx=(([a-z][a-z0-9_]{2,50})|(_[a-z0-9_]*))$
```
hnformentin, line 135: Why need to increase from 30 to 50?
`{2,50}` means match at least two times but no more than 50 times, is it correct?
I see that [below](https://github.com/equinor/semeio/pull/518/files#diff-ba293004ae8f18975051e825bd8d5c70137a8b5d42d16e0f7c9354aa634c446cL156) we keep 30, so wondered the reason behind this change.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -1,4 +1,4 @@
-# pxylint: disable=attribute-defined-outside-init
+# pylint: disable=attribute-defined-outside-init,invalid-name
```
hnformentin, line 1: Was `pxylint` only misspelled?
Why include `invalid-name`?

semeio/workflows/localisation/localisation_config.py
```diff
@@ -216,10 +216,9 @@ class ScalingForSegmentsConfig(BaseModel):
     ] = [0, 0]
 
     @validator("scalingfactors")
-    def check_length_consistency(cls, v, values):
+    def check_length_consistency(cls, scalingfactors, values):
```
hnformentin, line 219: Any reason not to be `scaling_factors`?

tests/hook_implementations/test_hook_implementations.py
```diff
@@ -12,7 +12,7 @@
 
 
 def test_hook_implementations():
-    pm = ErtPluginManager(
+    pluginmanager = ErtPluginManager(
```
hnformentin, line 15: plugin_manager?

.pylintrc
berland: There are some test function names that are longer than 30 characters. I think those test function names are legitimate and not something I want to disallow, so therefore I propose allowing up to 50 at least. One could think of having different requirements for test function names and production function names, but I think that just adds unnecessary complications to the CI.

tests/hook_implementations/test_hook_implementations.py
berland: Fixed. `plugin_manager` is in line with the spelling inside the module `ert.shared.plugins`. Hopefully that will not lead to more confusion.

hnformentin (approved):
Thanks for the replies 🚀

PR #519, opened 2023-05-23
Replace ErtImplType

andreas-el:
This PR was triggered by a change in Ert, -- namely the removal of ErtImplType.


DanSava (approved):
Onwards and upwards :rocket:

PR #529, opened 2023-07-13
Fix imports for configs

frode-aarstad:


frode-aarstad:
> i don't know if we have a system for dependency management that we use, but i feel like this is an API change of ert, and hence we should bump the version of ert as a dependency when updating these import lines - which of course requires merging ert first, and making a new tag / release, and us having ert as a versioned requirement in this project... yes i see we have ert with a min version in setup.py i suggest we wait for the ert PR to be merged, make a new tag, and then bump the min version here before merging

Not sure there is a formal process around this until we propose releases for komodo

valentin-krasontovitsch:
in my time as a release manager, i did not see this mentioned or formalized anywhere. but i guess it doesn't need to be formal - would you agree with the following steps:

- merge the ert PR
- make a new ert tag
- bump the `>=` requirement of ert to that new tag as a part of this PR ?

valentin-krasontovitsch:
the [test is failing](https://github.com/equinor/semeio/actions/runs/5539521859/jobs/10137077210) with
> E   ModuleNotFoundError: No module named 'ert.config'
from the logs, we can see it installs
> ert==5.1.0b3
so it looks like we have to bump the dependencies before the tests can pass - i guess this is a good way of formalizing it :sweat_smile:

oysteoh:
I guess a new tag on ert will do it as semeio then downloads the latest ?

oysteoh (approved):

valentin-krasontovitsch (approved):

Issue #532, opened 2023-08-02
Integration test for gendata-rft fails after updated ert gui


jonathan-eq:
**Issue**
Integration test for gendata-rft fails after updated ert gui

**Background**
The update in question is the rephrasing of some buttons and labels from "simulation" to "experiment"

PR #533, opened 2023-08-02
Fix integration tests for gendata_rft

jonathan-eq:
Recent changes to gui requires integrations tests to be updated. (certain buttons and labels have been rephrased from simulation to experiment)
Links to #532


hnformentin (approved):
Very good Jonathan 🚀

PR #534, opened 2023-08-11
Update to new ert api for no refcase

oyvindeide:


tests/communication/unit/test_file_reporter.py
```diff
@@ -146,17 +146,6 @@ def test_file_reporter_publish_valid_json(data, tmpdir):
...
-    tmpdir.chdir()
-    namespace = "data"
-    data = json  # The json module is not JSON serializable...
-
-    reporter = FileReporter(os.getcwd())
-
-    with pytest.raises(TypeError):
-        reporter.publish(namespace, data)
-
-
```
sondreso, line 159: Why are we removing this test?

oyvindeide: Because this is no longer invalid json after: `json.dump(all_data, f_handle, default=str)`, which was required when dumping `datetime` to json

sondreso: Is this file report used for anything? Because this means that it will be difficult to deserialize it again

oyvindeide: No. Well, that is not the _intention_, but what people are doing, I do not know. However I doubt nobody really knows about this.

oyvindeide: This is another one of those things that should never have been done in the first place, at least not externally from ert

sondreso (approved):

PR #537, opened 2023-09-18
Pin rstcheck-core because of pydantic issues

eivindjahren:
Although ert requires pydantic<2. Installing test-requirements.txt after installing ert means you get pydantic v2 due to rstcheck-core.


yngve-sk (approved):
LGTM

PR #541, opened 2023-10-03
Remove redundant log validator

dafeda:
Pydantic automatically handles validation based on type annotations.

This can be removed as long as `log_level` is never explicitly set to `None`.


kvashchuka:
It is set to None here:
https://github.com/equinor/semeio/blob/65b72b6eb063ef232822dc19bc4a09c79a0d8ce9/semeio/jobs/design2params/design2params.py#L33

kvashchuka:
Something funky is going on - I was testing this behaviour and I cannot set `log_level` to not integer at all. So, on the one hand, this check is definitely not needed, on the other, I don't understand what is going on 🤦‍♀️ Could we take a look together?

kvashchuka (approved):
As we learned today, pydantic enforces the type, therefore validator is redundant

PR #543, opened 2023-10-11
Integration test for non-adaptive localisation to check rowscaling

oddvarlia:


oddvarlia:
This is replaced by another PR and can be closed.

PR #545, opened 2023-10-20
Add non adaptive test case

oddvarlia:
### Test case for non-adaptive localisation. 

Need installation of **gaussianfft** or **gstools** to simulate gaussian fields.
gstools is per today preferred due to simpler installation while **gaussianfft** depends on **xtgeo** indirectly through the rms library **roxar** and their use of the Boost library. In future when roxar and gaussianfft uses identical version of the Boost library, the use of gaussianfft will work independent of import sequence of xtgeo and gaussianfft.

The following test are included:

- Use fields from files in ROFF format, Eclipse grid index origin. Localisation: decay function
- Use fields from files in GRDECL format, Eclipse grid index origin. Localistion decay function
- Use fields from files in ROFF format, RMS grid index origin. Localisation: decay function
- Use grid with some inactive grid cells. Localisation: decay function
- Use fields from files in ROFF format. Localisation: regions
- Use fields from files in ROFF format: Localisation: import scaling factor from file and eclipse grid index origin
- Use fields from files in ROFF format: Localisation: import scaling factor from file and RMS grid index origo

Closes: #547


tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,548 @@
...
+Common functions used by the scripts: init_test_case.py and sim_field.py
+"""
+import copy
+import math
+
+import gstools as gs
+import numpy as np
+import yaml
+
+import xtgeo  # isort: skip
```
dafeda, line 11: # isort: skip is no longer needed.

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,548 @@
...
+    upscaled_values = upscaling(
+        field_values,
+        settings,
+        write_field=True,
+        iteration=0,
+    )
+    return upscaled_values
+
+
+def get_seed(seed_file_name, r_number):
```
dafeda, line 186: The #noqa: E722 directive is used to tell linters like flake8 to ignore the bare except clause, which is not a recommended practice in Python.

To be more specific in your exception handling, catch only the exceptions you expect, in this case, the IndexError when trying to access an element from the lines list that doesn't exist and the ValueError when trying to convert a non-integer string to an integer.

```python
def get_seed(seed_file_name, r_number):
    with open(seed_file_name, "r", encoding="utf8") as file:
        lines = file.readlines()
        try:
            seed_value = int(lines[r_number - 1])
        except IndexError:
            raise IOError(f"Seed value not found for realization {r_number}")
        except ValueError:
            raise IOError(f"Invalid seed value in file for realization {r_number}")
    return seed_value
```

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,548 @@
...
+import math
+
+import gstools as gs
+import numpy as np
+import yaml
+
+import xtgeo  # isort: skip
+
+# pylint: disable=missing-function-docstring, too-many-locals, invalid-name
+# pylint: disable=bare-except, raise-missing-from
```
dafeda, line 14: This is not needed if `get_seed` is fixed as suggested below.

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,548 @@
...
+# pylint: disable= redefined-outer-name, too-many-nested-blocks
```
dafeda, line 15: The `redefined-outer-name` can be enabled it seems.

tests/jobs/localisation/example_case/scripts/sim_fields.py
```diff
@@ -0,0 +1,128 @@
...
+            raise IOError(
+                "Missing real_number (argv[2]) when running this script manually"
+            )
+        real_number = int(arg2)
+    print(f"ERT realization: {real_number}")
+    return iteration, real_number
+
+
+def main(config_file_name, iteration, real_number):
+    # pylint: disable=too-many-arguments
```
dafeda, line 68: pylint disable not needed.

tests/jobs/localisation/example_case/scripts/init_test_case.py
```diff
@@ -0,0 +1,281 @@
...
+    grid_object.to_file(grid_file_name, fformat="egrid")
+    return grid_object
+
+
+def main(config_file_name):
+    """
+    Initialize seed file, grid files, observation files and localisation config file
+    """
+    # pylint: disable=import-outside-toplevel
+    from common_functions import (
```
dafeda, line 228: Is there a good reason for importing here instead of on toplevel?

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,548 @@
...
+                        )
+            else:
+                raise KeyError(
+                    f"Unknown keyword '{key1}' specified under keyword '{main_key}'"
+                    f" in {yml_file_name}"
+                )
+    return settings
+
+
+def read_test_config(config_file_name):
```
dafeda, line 137: I think this can be improved somewhat by 1. early return and 2. using try and catch to get rid of the if-statement.
Something like this:

```python
import yaml

def read_test_config(config_file_name):
    if config_file_name is None:
        return specify_settings(None, None)

    try:
        with open(config_file_name, "r", encoding="utf-8") as yml_file:
            spec_dict = yaml.safe_load(yml_file)
    except FileNotFoundError:
        raise IOError(f"File {config_file_name} not found.")
    except yaml.YAMLError as e:
        raise ValueError(f"Failed to parse YAML file {config_file_name}. Ensure it's a valid .yml or .yaml file. Error: {e}")

    return specify_settings(spec_dict, config_file_name)

```

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: OK

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: OK

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: OK. Fixed

tests/jobs/localisation/example_case/scripts/sim_fields.py
oddvarlia: OK fixed

tests/jobs/localisation/example_case/scripts/init_test_case.py
oddvarlia: No not any longer. Fixed

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: This function is removed in the updated version. Not used.

tests/jobs/localisation/example_case/README
```diff
@@ -0,0 +1,44 @@
...
+​
+- **Other Files:**
+  - `randomseeds.txt`: Not used by ERT, but by `sim_fields.py`. This file can be generated by `scripts/init_test_case.py`
+  - `UpscaledGrid.EGRID`: Not used by ERT, but by `sim_fields.py`. This file can be generated by `scripts/init_test_case.py`
+​
+### Typical Workflow:
+​
+1. **Preparation:** Prepare ERT config input by running `scripts/init_test_case.py`.
+2. Directories for observations are created automatically according to the default settings.
+3. Make the directory `init_files` if not existing.
```
dafeda, line 30: Is step 3. needed given that we ran step 1.?

Also, we should give the `README` files the markdown extension `.md`.

oddvarlia: No, init_files directory is checked into the repo and has initially only the README file.
Fixed

tests/jobs/localisation/example_case/scripts/init_test_case.py
```diff
@@ -0,0 +1,262 @@
...
+
+# pylint: disable=too-many-arguments,invalid-name,missing-function-docstring
+# pylint: disable=too-many-locals,redefined-outer-name
+
+
+def generate_seed_file(
+    start_seed: int = 9828862224,
+    number_of_seeds: int = 1000,
+):
+    # pylint: disable=unused-variable
```
dafeda, line 22: Instead of disabling pylint here, we can rewrite the for-loop as follows:

`for _ in range(number_of_seeds):`.

The underscore indicates a variable that's not meant to be used.

tests/jobs/localisation/example_case/scripts/init_test_case.py
```diff
@@ -0,0 +1,262 @@
...
+import math
+import os
+import random
+
+# pylint: disable=import-error
+import xtgeo
+from common_functions import generate_field_and_upscale, settings, write_upscaled_field
+
+# pylint: disable=too-many-arguments,invalid-name,missing-function-docstring
+# pylint: disable=too-many-locals,redefined-outer-name
```
dafeda, line 15: Not necessary to ignore `redefined-outer-name` here I think.

tests/jobs/localisation/example_case/scripts/init_test_case.py
```diff
@@ -0,0 +1,262 @@
...
+            obs_file.write(f"OBS_FILE = ./{obs_data_relative_file_name} ; ")
+            obs_file.write(" };\n")
+            number += 1
+            data_file_name = observation_dir + "/" + obs_data_relative_file_name
+            print(f"Write file: {data_file_name} ")
+            with open(data_file_name, "w", encoding="utf8") as data_file:
+                data_file.write(f"{value}  {value_err}\n")
+
+
+def create_grid():
```
dafeda, line 152: The `create_grid` and other functions rely on a global settings object for their configuration. It makes the code harder to understand I think.
Could we instead consider parameterizing the functions?
For `create_grid` in particular, we could perhaps do something like this:

```python
def create_grid(grid_file_name, grid_dimension, grid_size, use_eclipse_grid_index_origo=False):
```

tests/jobs/localisation/example_case/scripts/init_test_case.py
```diff
@@ -0,0 +1,262 @@
...
+        rotation=0.0,
+        flip=flip,
+    )
+
+    print(f"Write grid file: {grid_file_name} ")
+    grid_object.to_file(grid_file_name, fformat="egrid")
+    return grid_object
+
+
+def create_upscaled_grid():
```
dafeda, line 186: If we parametrize like I suggested in another comment, I think we could just have one function to create grids, as the only differences between `create_upscaled_grid` and `create_grid` seem to be `grid_file_name` and `grid_dimension`.

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,526 @@
...
+    )
+
+    print(f"Simulate field with size: nx={nx},ny={ny} ")
+    field1D = sim.simulate(variogram, nx, dx, ny, dy, nz, dz)
+    field = field1D.reshape((nx, ny, nz), order="F")
+    return field
+
+
+def simulate_field_using_gstools(start_seed):
+    # pylint: disable=no-member,
```
dafeda, line 380: Disabling `pylint` here is not necessary I think.

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,526 @@
...
+    dy = ysize / ny
+    dz = zsize / nz
+
+    x = np.arange(0.5 * dx, xsize, dx)
+    y = np.arange(0.5 * dy, ysize, dy)
+    z = np.arange(0.5 * dz, zsize, dz)
+    # Rescale factor is set to:
+    #    sqrt(3.0) for gaussian correlation functions,
+    #    3.0 for exponetial correlation function,
+    #    pow(3.0, 1/alpha) for general exponential correlation function
```
dafeda, line 404: The `pow(3.0, 1/alpha)` rescale factor seems not to be used, in which case it should be removed from the comment.

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,571 @@
...
+    initial_file_name: str = "init_files/FieldParam.roff"
+    updated_file_name: str = "FieldParam.roff"
+    seed_file: str = "randomseeds.txt"
+    variogram: str = "exponential"
+    correlation_range: Tuple[float] = (3000.0, 1000.0, 2.0)
+    correlation_azimuth: float = 45.0
+    correlation_dip: float = 0.0
+    correlation_exponent: float = 1.9
+    trend_use: bool = False
+    trend_params: Tuple[float] = (1.0, -1.0)
```
dafeda, line 58: I think we should get rid of these type-hints for now as `semeio` does not use `mypy` and we'll probably just end up with a bunch of wrongly typed variables.
We should definitely start using `mypy` in `semeio` but can save that for another day.

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,604 @@
...
+            discrete=False,
+            name=field_name_selected,
+        )
+        print(f"Write conditioned cell values as field: {file_name_selected}")
+        cond_field_object.to_file(file_name_selected, fformat="roff")
+
+    return field_object
+
+
+def upscale_average(
```
dafeda, line 257: This function is quite complex and should really have a test of its own. Could we leverage some third-party library that already has upscaling implemented? I did a search for upscaling in `xtgeo` but came up short.

If we really need to write this our selves, we could perhaps at least try and leverage `np.mean`. Something like this:

```python
import numpy as np

def upscale_average(field_values, coarse_cell_index_list, upscaled_values, use_all=False):
    nx, ny, nz = field_values.shape
    NX, NY, NZ = upscaled_values.shape

    print(f"Number of fine scale grid cells:   (nx,ny,nz): ({nx},{ny},{nz})")
    print(f"Number of coarse scale grid cells: (NX,NY,NZ): ({NX},{NY},{NZ})  ")

    mx, my, mz = nx // NX, ny // NY, nz // NZ
    print(
        "Number of fine scale grid cells per coarse grid cell: "
        f"(mx,my,mz): ({mx},{my},{mz})"
    )

    if use_all:
        print("Calculate upscaled values for all grid cells")
        reshaped_field = field_values.reshape(NX, mx, NY, my, NZ, mz)
        avg_values = np.mean(reshaped_field, axis=(1, 3, 5))
        np.copyto(upscaled_values, avg_values)
    else:
        print("Calculate upscaled values for selected grid cells")
        for Iindx, Jindx, Kindx in coarse_cell_index_list:
            Iindx, Jindx, Kindx = Iindx - 1, Jindx - 1, Kindx - 1  # adjust for 1-based index

            i_slice = slice(Iindx * mx, (Iindx + 1) * mx)
            j_slice = slice(Jindx * my, (Jindx + 1) * my)
            k_slice = slice(Kindx * mz, (Kindx + 1) * mz)

            avg_value = np.mean(field_values[i_slice, j_slice, k_slice])
            upscaled_values[Iindx, Jindx, Kindx] = avg_value

    return upscaled_values

```

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,604 @@
...
+    use_trend = settings.field.trend_use
+    algorithm_method = settings.field.algorithm
+    start_seed = get_seed(seed_file_name, real_number)
+    if algorithm_method == "gstools":
+        print(f"Use algorithm: {algorithm_method}")
+        residual_field = simulate_field_using_gstools(start_seed)
+    else:
+        print("Use algorithm: gaussianfft")
+        residual_field = simulate_field(start_seed)
+    if use_trend == 1:
```
dafeda, line 151: I think this should be `if use_trend` since `use_trend` is a `bool`.

tests/jobs/localisation/example_case/scripts/init_test_case.py
oddvarlia: fixed

tests/jobs/localisation/example_case/scripts/init_test_case.py
oddvarlia: Need it after my last updated

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: Fixed

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: yes

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: I have done some update using numpy slice and mean and simplified the function

tests/jobs/localisation/example_case/scripts/init_test_case.py
```diff
@@ -0,0 +1,302 @@
...
+        rotation=0.0,
+        flip=flip,
+    )
+
+    print(f"Write grid file: {grid_file_name} ")
+    grid_object.to_file(grid_file_name, fformat="egrid")
+    return grid_object
+
+
+def main(config_file_name=None):
```
dafeda, line 238: Seems like a `yaml` config-file is now required in order to run `init_test_case.py`, which mean we can specify the input parameter to `main` as non-optional, i.e., remove the `=None`.
Also, the `README.md` could perhaps also reflect the fact that `init_test_case.py` needs a `yaml` config-file.

tests/jobs/localisation/example_case/scripts/common_functions.py
```diff
@@ -0,0 +1,916 @@
...
+        for sub_key in grid_size_dict:
+            if sub_key not in valid_keys:
+                err_msg.append(f"  {sub_key}")
+        if len(err_msg) > 0:
+            print(f"Unknown keywords in config file under keyword {key}: ")
+            for text in err_msg:
+                print(text)
+            raise KeyError("Unknown keywords")
+
+        grid_size_object = settings.grid_size
```
dafeda, line 202: I find this function a bit hard to read.
If I understand this correctly, it modifies parts of the global `settings` object.
I think we should try splitting it up somehow such that it does not modify global state.

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: Fixed

tests/jobs/localisation/example_case/scripts/init_test_case.py
oddvarlia: Done

tests/jobs/localisation/example_case/scripts/init_test_case.py
oddvarlia: Done

tests/jobs/localisation/example_case/scripts/init_test_case.py
oddvarlia: Done

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: The purpose is to modify hardcoded default settings with settings from the config file.
The settings object defined by settings = Settings() is now a global object used everywhere where necessary.
I can maybe let it be instantiated in the main function instead just after reading the yaml file and update it by the modifed settings from the yaml file and let the read_config_file return the settings object and use it only in the main function. I will try that, but it requires that all functions depending on settings will need a bunch of parameters. But as I understood that was preferred over a single global object containing settings.

and call the function update_setting

tests/jobs/localisation/example_case/test_full.py
```diff
@@ -0,0 +1,385 @@
...
+    monkeypatch.chdir(tmp_path)
+    print(f"Setup test case on tmp_path:  {tmp_path}")
+    settings = Settings()
+    settings.update(new_settings)
+
+    generate_seed_file(settings.field.seed_file)
+
+    # Make available data to generate grid
+    polygon_file = None
+    if settings.case_name == "A2":
```
dafeda, line 157: Can we avoid loading "Polygons.txt" and somehow re-create it?
Also, could we use a more descriptive case name than "A2"?

tests/jobs/localisation/example_case/test_full.py
```diff
@@ -0,0 +1,521 @@
+"""
+Test of full workflow using non-adaptive localisation
+- Gaussian fields are simulated
+- Response variables are upscaled values of simulated gaussian fields
+- Some selected positions within the upscaled grid is used to extract
```
dafeda, line 5:
```suggestion
- Some selected positions within the upscaled grid are used to extract
```

tests/jobs/localisation/example_case/test_full.py
```diff
@@ -0,0 +1,521 @@
...
+  synthetic observations.
+- Localisation config file is generated.
+- Running ERT using the generated localisation file.
+- The ERT forward model (simulating gaussian fields,
```
dafeda, line 9:
```suggestion
- The ERT forward model (simulating and upscaling gaussian fields is done by `sim_fields.py`).
```

tests/jobs/localisation/example_case/test_full.py
```diff
@@ -0,0 +1,521 @@
...
+  and upscale it is done using the script sim_fields.py)
+- For iteration > 0, the updated field parameter is imported from ERT
```
dafeda, line 11:
```suggestion
- For iteration > 0, the updated field parameter is imported from ERT and upscaled.
  Predictions of observed values are extracted and saved to ERT using the GEN_DATA keyword.
```

tests/jobs/localisation/example_case/test_full.py
oddvarlia: Fixed

tests/jobs/localisation/example_case/test_full.py
oddvarlia: Fixed

tests/jobs/localisation/example_case/test_full.py
oddvarlia: fixed

tests/jobs/localisation/example_case/scripts/common_functions.py
oddvarlia: Is handled in another way now in test_full.py The example config file is still read by the forward model, but generated automatically to avoid manual editing and errors.

oyvindeide:
Closing this PR as the underlying functionality in ert was removed.

PR #552, opened 2023-11-06
Build decay object in separate function

dafeda:
Pylint complains about duplicate code.

### DIFF
--- a/tests/jobs/localisation/test_methods.py
+++ b/tests/jobs/localisation/test_methods.py
@@ -8,10 +8,9 @@ from numpy import ma
 
 from semeio.workflows.localisation.local_script_lib import (
     ConstantScalingFactor,
-    ConstExponentialDecay,
-    ConstGaussianDecay,
     ExponentialDecay,
     GaussianDecay,
+    build_decay_object,
     calculate_scaling_factors_in_regions,
     smooth_parameter,
 )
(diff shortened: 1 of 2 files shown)


tests/jobs/localisation/test_methods.py
```diff
@@ -103,18 +102,11 @@ def create_parameter_from_decay_functions(method_name, grid):
...
-            main_range,
-            perp_range,
-            azimuth,
-            grid,
-            tapering_range,
-            use_cutoff,
-        )
-    elif method_name == "const_exponential_decay":
-        decay_obj = ConstExponentialDecay(
+    if method_name == "constant":
```
kvashchuka, line 105: Why this if cannot be part of `build_decay_object` function?

kvashchuka: Sorry for starting with "why":) I mean it in "I am curious why"-way

dafeda: Good question. `ConstantScalingFactor` is used only in `apply_constant` while the other decay-functions that get built in `build_decay_object` are for use in `apply_decay`.
This is not great and I think we can combine `apply_decay` and `apply_constant` into a single function, but perhaps in another PR.
What I really wanted to accomplish in this PR is to make `pylint` happy so that I can continue working on another PR that gets rid of `setup.py`.

PR #558, opened 2023-11-13
Introduce mypy

dafeda:
Replace conlist etc with Annotated and Field
to make mypy happy.

**NB**
Notice that I have set quite a few of the checks to `False` in `mypy.ini`, which means that this does not do a lot of type-checking yet.
The goal is to eventually set all of them to `True`.

### DIFF
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -1,0 +1,32 @@
+name: Type checking
+
+on:
+ push:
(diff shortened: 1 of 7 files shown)


kvashchuka (approved):
Great initiative! I have checked and removed mypy options that do not throw error with default values. Feel free to squash commit and merge :)

PR #561, opened 2023-11-21
Use returned snapshot instead of log file on disk

oyvindeide:


semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -293,20 +292,16 @@ def run(
...
         "name": "MINISTEP",
         "observations": obs_group,
         "parameters": data_parameters,
     }
     facade.update_configuration = [update_step]
-    facade.set_log_path(output_path)
-    # Perform update analysis
-
     facade.smoother_update(prior_storage, target_storage, target_storage.name)
+    return facade.update_snapshots[target_storage.name]
```
sondreso, line 304: Is this not returned from the function, such that we can avoid accessing this property of the facade (which I think we should get rid of over time)?

oyvindeide: Yes, it is, but there is no tagged version of that (yet), this way is backwards compatible. But yes, can do the tagging.

sondreso (approved):

PR #575, opened 2023-12-18
Fixed index order bugs

oddvarlia:
Fix grid index order bug in local_script_lib.py

Close issue #576


tests/jobs/localisation/test_methods.py
```diff
@@ -84,7 +84,8 @@ def create_region_parameter(
         mask_used = np.isin(region_param, used_regions, invert=True)
         region_param[np.logical_and(mask_used, ~inactive)] = ma.masked
 
-    return region_param.ravel(order="F"), scaling_param.ravel(order="F")
+    # return region_param.ravel(order="F"), scaling_param.ravel(order="F")
```
oyvindeide, line 87: Extra comment

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -556,9 +665,10 @@ def smooth_parameter(
     nz = grid.get_nz()
     di = smooth_range_list[0]
     dj = smooth_range_list[1]
-    scaling_values_smooth = np.zeros(nx * ny * nz, dtype=np.float32)
+    scaling_values_smooth = np.ma.zeros(nx * ny * nz, dtype=np.float32)
     for k, j0, i0 in itertools.product(range(nz), range(ny), range(nx)):
-        index0 = i0 + j0 * nx + k * nx * ny
+        # index0 = i0 + j0 * nx + k * nx * ny
```
oyvindeide, line 682: Commented out code

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -568,7 +678,8 @@ def smooth_parameter(
             jhigh = min(j0 + dj + 1, ny)
             for i in range(ilow, ihigh):
                 for j in range(jlow, jhigh):
-                    index = i + j * nx + k * nx * ny
+                    # index = i + j * nx + k * nx * ny
```
oyvindeide: Commented out code

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -627,21 +738,21 @@ def apply_segment(
...
 
     # Assign values to row_scaling object
-    for index in range(data_size):
-        global_index = grid.global_index(active_index=index)
-        row_scaling[index] = scaling_values[global_index]
+    row_scaling.assign_vector(scaling_values)
+
+    #    for index in range(data_size):
+    #        global_index = grid.global_index(active_index=index)
+    #        row_scaling[index] = scaling_values[global_index]
```
oyvindeide: Commented out code

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -697,7 +810,9 @@ def read_region_files_for_all_correlation_groups(user_config, grid):
                     region_parameter = np.zeros(nx * ny * nz, dtype=np.int32)
                     not_active = np.zeros(nx * ny * nz, dtype=np.int32)
                     for k, j, i in itertools.product(range(nz), range(ny), range(nx)):
-                        index = i + j * nx + k * nx * ny
+                        # c-index order
+                        index = k + j * nz + i * nz * ny
+                        # index = i + j * nx + k * nx * ny
```
oyvindeide, line 822: Commented out code

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -220,33 +227,85 @@ def write_qc_parameter(
...
+            f"{scaling_kw_name} to file: {filename}",
+            LogLevel.LEVEL3,
+            log_level,
         )
 
+        # Increase parameter number to define unique parameter name
+        cls.scaling_param_number = cls.scaling_param_number + 1
+
+    @classmethod
+    def write_qc_parameter_surface(
```
oyvindeide, line 260: Seems this is never called. Understand that this was probably used for debugging, but if we are adding it we should also add some tests for the functionality.

oyvindeide (commented):
Looks like a good improvement! Seems there are some stray comments. There are also some added features that are not tested. Optionally they could be added in another PR. Seems like the bug which is being fixed here has test coverage already in the form of a snapshot test?

tests/jobs/localisation/test_methods.py
oddvarlia: fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: Fixed

semeio/workflows/localisation/local_script_lib.py
oddvarlia: image

It is used when the user specify to get QC files for scaling factor

oddvarlia: image
 
This map is now output from the test_integration.py for test_localisation_surf() It is activated by using the keyword:
write_scaling_factors = True in localisation config file.
For QC compare with the specification of ranges and anisotropy:
image

oddvarlia: So the already existing test can be used.

semeio/workflows/localisation/local_script_lib.py
```diff
@@ -220,33 +227,85 @@ def write_qc_parameter(
...
-                log_level,
-            )
-            with cwrap.open(filename, "w") as file:
-                grid.write_grdecl(scaling_kw, file)
-            # Increase parameter number to define unique parameter name
-            cls.scaling_param_number = cls.scaling_param_number + 1
+        if corr_name == cls.corr_name:
+            return
+
+        cls.corr_name = corr_name
```
oyvindeide, line 279: This is probably not a good idea, as it has some wierd side effects, take for example:
```python
class MyClass:
    foo = 1

    @classmethod
    def bar(cls):
        if cls.foo == 1:
            print(1)
        else:
            print(1)
        cls.foo = 2

test = MyClass()
print(test.foo)
>>> 1

MyClass.bar()

print(test.foo)
>>> 2
```

Notice that this last print is 2, even though `test` has never called `bar`

it even means that if we later do:

```python
test_2 = MyClass()

print(test_2.foo)
>>> 2
```

which will be very hard to track down if we ever need to debug it.

oddvarlia: It is possible to remove the class ScalingValues and let the function add_ministeps count the number of fields and surface parameters used in a ministep to avoid writing the qc data more than once per ministep per type (field, surface)

oddvarlia: I will propose an new solution without the ScalingValues class.

semeio/workflows/localisation/local_script_lib.py
oddvarlia: A test is now implemented comparing generated qc file with scaling factor with a reference.

oyvindeide:
Closing this PR as the underlying functionality in ert was removed.

PR #595, opened 2024-02-19
Add TMPDIR to environment passed to ERT in integration test

larsevj:
If TMPDIR is set to anything but /tmp, then ert will currently leave behind a folder in /tmp due to not beeing passed the TMPDIR environment variable.


berland:
LGTM

berland (approved):

PR #600, opened 2024-03-01
Update to new ert api

oyvindeide:


sondreso (approved):
Will you create an issue for bumping ert to a no-rc tag?

Issue #603, opened 2024-03-08
AHM_ANALYSIS fails with key error


oyvindeide:
**Describe the bug**
Running ahm_analysis on a case fails with a key error, though the ensemble and parameter are present in storage

Issue #498, opened 2023-01-31
AHM_ANALYSIS: Workflow fails if all observations are deactivated during update


oyvindeide:
**Describe the bug**
Run the AHM_ANALYSIS report with a case where all the observations in a group are deactivated because of outlier detection. The analysis will fail, causing the whole job to fail. Instead we should catch the exception and make a reasonable update report instead, which highlights which observations are excluded.

oddvarlia:
This error also occurs when running the LOCALISATION_JOB. The localisation code should check if some observations are active before creating a ministep and just skip the ministep (not create it) if all observations are deactivated, BUT should report this with a warning.

PR #605, opened 2024-03-20
Fix bug where ahm_analysis failed when observation groups were disabled

oyvindeide:
Closes: #603 
Closes: #601
Closes: #498


src/semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -164,18 +171,24 @@ def run(
         # loop over keys and calculate the KS matrix,
         # conditioning one parameter at the time.
         field_output = {}
+        updated_combinations = deepcopy(combinations)
         for group_name, obs_group in combinations.items():
             print("Processing:", group_name)
 
             #  Use localization to evaluate change of parameters for each observation
-            with tempfile.TemporaryDirectory():
+            with tempfile.TemporaryDirectory(), open_storage("storage", "w") as storage:
```
sondreso, line 179: Maybe deserves a comment? Something like this
```suggestion
            # The ordering of the context managers matters, we want to open a new storage in the
            # temporary directory
            with tempfile.TemporaryDirectory(), open_storage("storage", "w") as storage:
```

src/semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -292,8 +298,14 @@ def run(
 
 def _run_ministep(facade, prior_storage, target_storage, obs_group, data_parameters):
     # pylint: disable=too-many-arguments
+    rng = np.random.default_rng(seed=facade.config.random_seed)
```
sondreso, line 301: This makes me slightly nervous, what is the reasoning?

src/semeio/workflows/ahm_analysis/ahmanalysis.py
sondreso: RIP formatting...

src/semeio/workflows/ahm_analysis/ahmanalysis.py
oyvindeide: Why does that make you nervous? It is just so it is reproducible. It was primarily for my snapshot test, but it is just using the seed the user has (optionally) provided.

oyvindeide: Created issue for saving random seed in ert: https://github.com/equinor/ert/issues/7499

PR #606, opened 2024-03-21
Remove fields from AHM_ANALYSIS

oyvindeide:


sondreso (approved):
The PR description/commit body needs an update to explain why this feature is removed. 
We should also create an issue (pointing to this commit) to optimize the code to not load all field parameters into memory

PR #607, opened 2024-03-22
Sort keys on merge in design2param

larsevj:
The sort behaviour in pandas merge was "fixed" in pandas 2.2: https://pandas.pydata.org/docs/dev/whatsnew/v2.2.0.html

An alternative solution is to relax test requirements somehow?

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -14,7 +14,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python 3.8
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: 3.8
       - name: Install dependencies
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -9,9 +9,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Setup Python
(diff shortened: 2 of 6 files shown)


berland:
We want to preserve the order of parameters in the XLS file in the appended parameters.txt.

src/semeio/forward_models/design2params/design2params.py
```diff
@@ -98,13 +98,24 @@ def _complete_parameters_file(
...
+            on="keys",
+            how="outer",
+            copy=False,
+        ),
         on="keys",
         how="outer",
+        copy=False,
     )
-
+    merged.sort_values(["order_param", "order_real", "order_default"], inplace=True)
```
berland, line 118: If I change the order of "order_param" and "order_real" on this line, I will not fail the tests. Can you add sufficient tests to assert that what this line specifies cannot be changed?

berland: It might not be important, but it would be nice to have it specified by a test so we would know if it changes later. The process of writing the test is probably the way to figure out if it is important or not :)

berland: Nice. It only applies to logging then I reckon? (which one could argue is not as important..)

berland (approved):
👍🏻

PR #609, opened 2024-04-05
Adapt to combined observations in ert

yngve-sk:
Uses `experiment.observation_keys` instead of `experiment.observations.keys()` (which now returns `["gen_data", "summary"]`)
Should be merged in along with:
https://github.com/equinor/ert/pull/7563


oyvindeide (approved):
Just split style changes into a separate commit or drop them

PR #613, opened 2024-05-23
Forward models as ert plugins inheriting from ForwardModelStep

yngve-sk:
Requires https://github.com/equinor/ert/pull/7743


src/semeio/forward_models/__init__.py
```diff
@@ -0,0 +1,160 @@
+# class notation
```
eivindjahren, line 1: superflous comment

src/semeio/forward_models/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+from ert.config.forward_model_step import ForwardModelStepJSON, ForwardModelStep
```
eivindjahren, line 2: We should have `from ert import` for the final interface.

src/semeio/forward_models/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class Design2Params(ForwardModelStep):
+    def __init__(self):
+        super().__init__(
+            name="DESIGN2PARAMS",
+            executable="design2params",
+            arglist=["<IENS>", "<xls_filename>", "<designsheet>", "<defaultssheet>"],
```
eivindjahren, line 10: We should probably use python types here

```
arglist[("<IENS", int), ("<xls_filename>", Annotated[Path, PathType('file')]) ...
```

That way we can use standard functions for checking types. Not sure exactly which one, but `beartypes` should have something for this.

src/semeio/forward_models/__init__.py
```diff
@@ -0,0 +1,160 @@
+# class notation
+from ert.config.forward_model_step import ForwardModelStepJSON, ForwardModelStep
+
+
+class Design2Params(ForwardModelStep):
+    def __init__(self):
+        super().__init__(
+            name="DESIGN2PARAMS",
+            executable="design2params",
```
eivindjahren, line 9: Potentially, in order to mimic `subprocess.run` it would be convenient to specify the cmd_line args:

```python
cmd=["design2params", "<IENS>", "<xls_filename>", "<designsheet>", "<defaultssheet>"],
```

eivindjahren: I believe subprocess allows both list and str.

tests/forward_models/rft/test_gendata_rft.py
```diff
@@ -466,7 +466,7 @@ def test_defaults():
 
     fm_description_file = os.path.join(
         os.path.dirname(semeio.forward_models.__file__),
-        "config",
+        "config_deprecated",
```
eivindjahren, line 469: debugging stuff left in here.

tests/test_ert_integration.py
```diff
@@ -77,7 +79,7 @@ def test_forward_model_error_propagation(forward_model, configuration, expected_
                 semeio.hook_implementations.forward_models,
                 ert.shared.hook_implementations,
             ]
-        ):
+        ) as ctx:
```
eivindjahren, line 82: this change should not be necessary.

tests/forward_models/rft/test_gendata_rft.py
yngve-sk: Removed the test, as it asserts a relationship between the previously used config files filenames and the names output by ERT

src/semeio/forward_models/__init__.py
yngve-sk: I think we can leave types/validation up to "users" implementing this for a simpler interface on our end

src/semeio/forward_models/__init__.py
yngve-sk: Did this, stuck to only list syntax instead of str, even though both are possible

tests/hook_implementations/test_hook_implementations.py
```diff
@@ -3,11 +3,23 @@
...
 
+def test_that_installable_fm_steps_work_as_plugins():
+    plugin_manager = ErtPluginManager()
+    fms = plugin_manager.forward_model_steps
+
+    assert Design2Params in fms
+    assert DesignKW in fms
+    assert Pyscal in fms
+    assert GenDataRFT in fms
+    assert OTS in fms
```
andreas-el, line 20: I assume it was done by design to omit `InsertNoSim`, `RemoveNoSim` & `ReplaceStringConfig` ?

yngve-sk: Actually nice catch, I think these omited values are actually meant to be included

andreas-el (approved):
I assume you will rebase and remove the commit containing the debug repository.
Otherwise this looks proper. 💯

src/semeio/forward_models/__init__.py
```diff
@@ -0,0 +1,135 @@
+from ert.config import ForwardModelStepJSON, ForwardModelStepPlugin
```
eivindjahren, line 1: These two symbols should be exposed through `ert/__init__.py` so that we can import it like this:
```
from ert import ForwardModelStepJSON, ForwardModelStepPlugin
```

eivindjahren: This way we hide detail about internal structure in ert.

PR #620, opened 2024-06-10
Use macos-latest in github workflow tests

### LINKED ISSUE #616: python3.12 tests fail with macos-latest but not macos-12
Should use macos-12

larsevj:
Resolves #616


xjules (approved):
Yes, let's skip it :smile:  :rocket:

PR #635, opened 2024-07-30
Move csv_export2 into ert

eivindjahren:


yngve-sk (approved):
LGTM given that it now works in ert

PR #640, opened 2024-09-05
Dont use get_ensemble_by_name from LocalStorage

dafeda:
erts api has been updated such that get_ensemble_by_name is only available via LocalExperiment.


oyvindeide (approved):
LGTM!

PR #646, opened 2024-09-17
Add `validate_pre_experiment`-method to DESIGN_KW forward model

### LINKED ISSUE #641: Validate DESIGN_KW configuration at start up instead of when running forward model
The following list of errors can happen on running the forward model step. It should be validated at startup.

* `WARNING - Key(s) [...] can only be used with prefix.`
* `ERROR - <...> not found in design matrix`
* `design_kw: error: argument templatefile: /..../input/templates/plate.tmpl is not an existing file!`
* `MULTFLT is defined multiple times`

Also the following uncaught exceptions from `extract_key_value` can occurr which should be result in a better error message at startup.

```
File ".../python3.8/site-packages/semeio/forward_models/design_kw/design_kw.py", line 113, in extract_key_value
line_parts = shlex.split(line)
File "/usr/lib64/python3.8/shlex.py", line 311, in split
return list(lex)
File "/usr/lib64/python3.8/shlex.py", line 300, in __next__
token = self.get_token()
File "/usr/lib64/python3.8/shlex.py", line 109, in get_token
raw = self.read_token()
File "/usr/lib64/python3.8/shlex.py", line 191, in read_token
raise ValueError("No closing quotation")
ValueError: No closing quotation
```
```
File ".../python3.8/site-packages/semeio/forward_models/design_kw/design_kw.py", line 135, in extract_key_value
raise ValueError("\n".join(errors))
ValueError: No value found in line ...
```

jonathan-eq:
Resolves #641

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,8 @@ types = [
 addopts = "-ra --strict-markers"
 markers = [
     "ert_integration: Involves running the ert application explicitly",
-    "equinor_test: Requires presences of test data in specified location"
+    "equinor_test: Requires presences of test data in specified location",
+    "integration_test: Not a unit test",
 ]
 
 [tool.ruff]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     "resdata",
-    "ert>=11.0.0-b0",
+    "ert>=11.1.0-b1",
     "importlib_metadata",
     "importlib_resources",
     "numpy",
--- a/src/semeio/forward_models/__init__.py
+++ b/src/semeio/forward_models/__init__.py
@@ -4,10 +4,12 @@ from ert import (
     ForwardModelStepDocumentation,
     ForwardModelStepJSON,
     ForwardModelStepPlugin,
+    ForwardModelStepValidationError,
 )
 
 from .scripts.design2params import description as design2params_description
 from .scripts.design_kw import description as design_kw_description
+from .scripts.design_kw import design_kw as DesignKWScript
 
 
 class Design2Params(ForwardModelStepPlugin):
@@ -45,7 +47,8 @@ class DesignKW(ForwardModelStepPlugin):
     def __init__(self):
         super().__init__(
             name="DESIGN_KW",
-            command=["design_kw", "<template_file>", "<result_file>"],
+            command=["design_kw", "<template_file>", "<result_file>", "<VALIDATE>"],
+            default_mapping={"<VALIDATE>": "false"},
         )
 
     def validate_pre_realization_run(
@@ -54,7 +57,15 @@ class DesignKW(ForwardModelStepPlugin):
         return fm_step_json
 
     def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
-        pass
+        if self.private_args.get("<VALIDATE>", "").lower() == "true":
+            try:
+                template_file_name = self.private_args["<template_file>"]
+            except KeyError:
+                raise ForwardModelStepValidationError(
+                    "Argument list is missing <template_file>"
+                ) from None
+            DesignKWScript.validate_configuration(template_file_name=template_file_name)
+        return fm_step_json
 
     @staticmethod
     def documentation() -> Optional[ForwardModelStepDocumentation]:
--- a/src/semeio/forward_models/design_kw/design_kw.py
+++ b/src/semeio/forward_models/design_kw/design_kw.py
@@ -1,11 +1,17 @@
 import logging
 import re
 import shlex
-from typing import List
+from typing import Dict, List, Mapping
+
+from ert import ForwardModelStepWarning
 
 _STATUS_FILE_NAME = "DESIGN_KW.OK"
 
 _logger = logging.getLogger(__name__)
+
+
+class ParameterSyntaxError(ValueError):
+    pass
 
 
 def run(
@@ -39,7 +45,7 @@ def run(
                 for key, value in key_vals.items():
                     line = line.replace(f"<{key}>", str(value))
 
-                if not all_matched(line, template_file_name, template):
+                if find_matching_errors(line, template_file_name, template):
                     valid = False
 
             result_file.write(line)
@@ -51,8 +57,10 @@ def run(
     return valid
 
 
-def all_matched(line, template_file_name, template):
-    valid = True
+def find_matching_errors(
+    line: str, template_file_name: str, template: List[str]
+) -> List[str]:
+    errors = []
     for unmatched in unmatched_templates(line):
         if is_perl(template_file_name, template):
             _logger.warning(
@@ -65,9 +73,10 @@ def all_matched(line, template_file_name, template):
                 f"but this is probably an xml file"
             )
         else:
-            _logger.error(f"{unmatched} not found in design matrix")
-            valid = False
-    return valid
+            error_msg = f"{unmatched} not found in design matrix"
+            _logger.error(error_msg)
+            errors.append(error_msg)
+    return errors
 
 
 def is_perl(file_name, template):
@@ -91,7 +100,7 @@ def is_comment(line):
     return ecl_comment_pattern.search(line) or std_comment_pattern.search(line)
 
 
-def extract_key_value(parameters):
+def extract_key_value(parameters: List[str]) -> Dict[str, str]:
     """Parses a list of strings, looking for key-value pairs pr. line
     separated by whitespace, into a dictionary.
 
@@ -110,7 +119,11 @@ def extract_key_value(parameters):
     res = {}
     errors = []
     for line in parameters:
-        line_parts = shlex.split(line)
+        try:
+            line_parts = shlex.split(line)
+        except ValueError as e:
+            errors.append(f"Line '{line}' failed with '{str(e)}'")
+            continue
         if not line_parts:
             continue
         if len(line_parts) == 1:
@@ -125,7 +138,7 @@ def extract_key_value(parameters):
             continue
         res[key] = value
     if errors:
-        raise ValueError("\n".join(errors))
+        raise ParameterSyntaxError("\n".join(errors))
     return res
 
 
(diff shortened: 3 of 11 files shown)


src/semeio/forward_models/__init__.py
```diff
@@ -54,7 +56,11 @@ def validate_pre_realization_run(
         return fm_step_json
 
     def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
-        pass
+        errors = DesignKWScript.validate_configuration(fm_step_json.get("argList")[0])
+        if errors:
+            raise ForwardModelStepValidationError(
```
eivindjahren, line 61: Updated the issue as we need this to be a warning first before we commit to having an error.

eivindjahren (changes requested):
Need this to be a warning instead of an error for now.

jonathan-eq:
Blocked by https://github.com/equinor/ert/pull/8734

src/semeio/forward_models/__init__.py
```diff
@@ -54,7 +58,15 @@ def validate_pre_realization_run(
         return fm_step_json
 
     def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
-        pass
+        errors = DesignKWScript.validate_configuration(fm_step_json.get("argList")[0])
+        should_fail_on_validation_error = (
+            fm_step_json.get("argList")[2].lower() == "true"
```
eivindjahren, line 63: For now, validate=False should hide the warning message, and the error never fail. The default value should be to warn.

jonathan-eq:
This is blocked until ert 11 is in stable

eivindjahren:
@jonathan-eq You can still create a beta release for ert and set semeio to depend on `ert>=11.0.0-b0`.

src/semeio/forward_models/__init__.py
```diff
@@ -54,7 +57,13 @@ def validate_pre_realization_run(
...
+        )
+        error_message = f"DESIGN_KW configuration validation failed:\n{errors}"
```
xjules, line 64: This should go inside the if:

pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers = [
 dynamic = ["version"]
 dependencies = [
     "resdata",
-    "ert>=11.0.0-b0",
+    "ert>=11.0.0-rc0",
```
xjules, line 32: Add this as a standalone commit.

src/semeio/forward_models/__init__.py
```diff
@@ -54,7 +57,13 @@ def validate_pre_realization_run(
         return fm_step_json
 
     def validate_pre_experiment(self, fm_step_json: ForwardModelStepJSON) -> None:
-        pass
+        errors = DesignKWScript.validate_configuration(fm_step_json.get("argList")[0])
```
xjules, line 60: well, this still does not make sense. Ie. when `should_fail_on_validation_error` we don't need to do any validation based on the code at least. Then it can all go into:
```python
if should_fail_on_validation_error:
      DesignKWScript.validate_configuration(fm_step_json.get("argList")[0])
      ...
```

jonathan-eq: I agree, but I do not know if I want to call a `ForwardModelStepWarning` method directly from  DesignKWScript. The call to `ForwardModelStepWarning.warn(...)` only makes sense when called from a `ForwardModelStepPlugin` context.

eivindjahren: Should definily only validate inside the if statement as there are at least two bugs that causes validation to crash, see https://github.com/equinor/semeio/issues/641. Should probably also fix those bugs and have a defensive try-catch around the validation.

src/semeio/forward_models/__init__.py
```diff
@@ -40,21 +41,32 @@ def documentation() -> Optional[ForwardModelStepDocumentation]:
...
+            command=["design_kw", "<template_file>", "<result_file>", "<VALIDATE>"],
+            default_mapping={"<VALIDATE>": "false"},
         )
 
     def validate_pre_realization_run(
         self, fm_step_json: ForwardModelStepJSON
     ) -> ForwardModelStepJSON:
         return fm_step_json
 
+    @validate
```
xjules, line 67: `@validate` is actually a bad name :smile: it's more like `@validation_enabled`

src/semeio/forward_models/__init__.py
```diff
@@ -40,21 +41,32 @@ def documentation() -> Optional[ForwardModelStepDocumentation]:
             description=design2params_description,
         )
 
+def validate(validation_func):
+    def wrapper(self, fm_step_json):
+        print(f"{fm_step_json=}")
+        should_fail_on_validation_error = (
+            fm_step_json.get("argList")[2].lower() == "true"
```
xjules, line 48: Either the index of the validate arg is sent as a decorator parameter or can we just loop over all parameters if one equals "validate"?

src/semeio/forward_models/__init__.py
jonathan-eq: We do not get the argument key, only the value from `ForwardModelStepJSON`. I think it is better to assume the final argument value is for validation instead of looping over and looking for a string of a boolean
```
fm_step_json={'name': 'DESIGN_KW', 'executable': 'design_kw', 'target_file': None, 'error_file': None, 'start_file': None, 'stdout': 'DESIGN_KW.stdout.0', 'stderr': 'DESIGN_KW.stderr.0', 'stdin': None, 'argList': ['template.yml.tmpl', '<result_file>', 'true'], 'environment': {'_ERT_ITERATION_NUMBER': '0', '_ERT_REALIZATION_NUMBER': '0', '_ERT_RUNPATH': '/Users/JONAK/Documents/FMU/SCOUT/ert/test-data/poly_example/poly_out/realization-0/iter-0'}, 'exec_env': None, 'max_running_minutes': None}
```

src/semeio/forward_models/__init__.py
```diff
@@ -8,6 +8,17 @@
...
 from .scripts.design2params import description as design2params_description
 from .scripts.design_kw import description as design_kw_description
+from .scripts.design_kw import design_kw as DesignKWScript
+
+
+def validation_enabled(validation_func: Callable) -> Callable[..., None]:
+    def wrapper(
+        self: ForwardModelStepPlugin, fm_step_json: ForwardModelStepJSON
+    ) -> None:
+        if "<VALIDATE>" in self.private_args.get("<VALIDATE>", "").lower() == "true":
```
xjules, line 18: I guess it should be:
```python
if self.private_args.get("<VALIDATE>", "").lower() == "true":
```

src/semeio/forward_models/design_kw/design_kw.py
```diff
@@ -178,3 +184,41 @@ def rm_genkw_prefix(paramsdict, ignoreprefixes="LOG10_"):
...
+
+        with open(template_file_name, encoding="utf-8") as template_file:
+            template = template_file.readlines()
+
+        for line in template:
+            if not is_comment(line):
+                for key, value in key_vals.items():
+                    line = line.replace(f"<{key}>", str(value))
+
+                all_matched(line, template_file_name, template)
```
xjules, line 217: in `all_matched` this one does not raise `ValueError` but just logger.error. Not sure how to propagate it here, caplog?

src/semeio/forward_models/scripts/design_kw.py
```diff
@@ -44,7 +46,7 @@ def create_parser():
         default="WARNING",
         type=logging.getLevelName,
     )
-
+    parser.add_argument("--validate", required=False, type=bool, default=True)
```
xjules, line 49: I thought default is `False`?

jonathan-eq: It is false in `ForwardModelStepPlugin`'s default_mapping, but I can have it default False here too :)

xjules: So why we need both? :)

jonathan-eq: I think it is so that someone running it directly as a script won't have to specify it every time. 
We do the same for `--log_level`

xjules: Clarified!

src/semeio/forward_models/__init__.py
```diff
@@ -45,16 +56,20 @@ class DesignKW(ForwardModelStepPlugin):
...
+            command=["design_kw", "<template_file>", "<result_file>", "<VALIDATE>"],
+            default_mapping={"<VALIDATE>": "false"},
         )
 
     def validate_pre_realization_run(
         self, fm_step_json: ForwardModelStepJSON
     ) -> ForwardModelStepJSON:
         return fm_step_json
 
+    @validation_enabled
```
xjules, line 68: Now when we won't do `DESIGN2PARAMS` validation maybe this can be simplified to the previous version? Although this looks quite nice :)

xjules (approved):
Nice job @jonathan-eq ! :rocket:

src/semeio/forward_models/design_kw/design_kw.py
```diff
@@ -178,3 +184,41 @@ def rm_genkw_prefix(paramsdict, ignoreprefixes="LOG10_"):
...
+        if not os.path.exists(template_file_name):
+            raise ValueError(
+                f"Argument template_file: '{template_file_name}' is not an existing file"
+            )
+
+        if not os.path.exists(parameters_file_name):
+            raise ValueError(
+                f"Could not find required parameters file from DESIGN2PARAMS: '{parameters_file_name}'"
+            )
+        with open(parameters_file_name, encoding="utf-8") as parameters_file:
```
eivindjahren, line 202: This can fail if you don't have read permissions for the parameters_file_name and crash ert instead of showing a warning in the GUI

src/semeio/forward_models/design_kw/design_kw.py
```diff
@@ -178,3 +184,41 @@ def rm_genkw_prefix(paramsdict, ignoreprefixes="LOG10_"):
...
+            parameters = parameters_file.readlines()
+
+        key_vals = extract_key_value(parameters)
+
+        key_vals.update(rm_genkw_prefix(key_vals))
+
+        with open(template_file_name, encoding="utf-8") as template_file:
```
eivindjahren, line 209: This can fail if you don't have read permissions for the template_file_name and crash ert instead of showing a warning in the GUI

src/semeio/forward_models/design_kw/design_kw.py
```diff
@@ -178,3 +184,41 @@ def rm_genkw_prefix(paramsdict, ignoreprefixes="LOG10_"):
...
+            )
+
+        if not os.path.exists(parameters_file_name):
+            raise ValueError(
+                f"Could not find required parameters file from DESIGN2PARAMS: '{parameters_file_name}'"
+            )
+        with open(parameters_file_name, encoding="utf-8") as parameters_file:
+            parameters = parameters_file.readlines()
+
+        key_vals = extract_key_value(parameters)
```
eivindjahren, line 205: extract_key_value currently raises `ValueError` from shlex with the message "No closing quotation" see https://github.com/equinor/semeio/issues/641 . `extract_key_value` needs to be fixed so that it doesn't raise in order for this design to be used. I would suggest both fixing the bug with `extract_key_value` so that it shows a reasonable error message rather than throwing from shlex, and not relying on catching an exception in order to indicate warnings.

src/semeio/forward_models/design_kw/design_kw.py
```diff
@@ -178,3 +184,41 @@ def rm_genkw_prefix(paramsdict, ignoreprefixes="LOG10_"):
...
         if keys.count(keyvalue[0]) == 1
     }
+
+
+def validate_configuration(
+    template_file_name: str, parameters_file_name: str = "parameters.txt"
+) -> None:
+    try:
+        if not os.path.exists(template_file_name):
+            raise ValueError(
```
eivindjahren, line 194: By raising and aborting at this point you only get the first warning that is encountered which can be frustrating for users. If you structure the validation like this:

```python
def validate_configuration(
    template_file_name: str, parameters_file_name: str = "parameters.txt"
) -> None:
    validate_template_file(template_file_name)
    validate_parameters_file(parameters_file_name)
```
you can create warnings both for the template file and the parameters file, when there are warnings for both.

xjules: Agree with this one :+1:

eivindjahren (approved):
:+1:

jonathan-eq:
Blocked by https://github.com/equinor/ert/issues/8868

eivindjahren:
This should be ready now

PR #649, opened 2024-10-07
Fix `DESIGN_KW` validation flag bug

jonathan-eq:
This commit fixes the issue where the `<VALIDATE>=true` argument would be parsed as a positional argument instead of a keyword argument, and the arg parsed would raise an exception for unexpected argument `false`. After this commit, the <VALIDATE> argument is only for ert validation purposes, and is not forwarded to the `DESIGN_KW` argparser.

### DIFF
--- a/src/semeio/forward_models/__init__.py
+++ b/src/semeio/forward_models/__init__.py
@@ -47,7 +47,12 @@ class DesignKW(ForwardModelStepPlugin):
     def __init__(self):
         super().__init__(
             name="DESIGN_KW",
-            command=["design_kw", "<template_file>", "<result_file>", "<VALIDATE>"],
+            command=[
+                "design_kw",
+                "<template_file>",
+                "<result_file>",
+                "--validate=<VALIDATE>",
+            ],
             default_mapping={"<VALIDATE>": "false"},
         )
 
--- a/src/semeio/forward_models/scripts/design_kw.py
+++ b/src/semeio/forward_models/scripts/design_kw.py
@@ -46,7 +46,7 @@ def create_parser():
         default="WARNING",
         type=logging.getLevelName,
     )
-    parser.add_argument("--validate", required=False, type=bool, default=False)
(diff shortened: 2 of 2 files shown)


eivindjahren:
It is really important one can set the `<VALIDATE>` keyword in `FORWARD_MODEL DESIGN_KW(...`, how is it that one does that now? Also, why didn't the default value work? It is supposed to work.

jonathan-eq:
> It is really important one can set the `<VALIDATE>` keyword in `FORWARD_MODEL DESIGN_KW(...`, how is it that one does that now? Also, why didn't the default value work? It is supposed to work.

It is still the same, but the value is never passed to the DESIGN_KW script, it is only retrieved in the validate_forward_model_config step. The default value was passed to the script as a positional argument (without the `--validate=`-part) ,  which is why we got the error

eivindjahren:
Then it should be possible to use 
```python
        super().__init__(
            name="DESIGN_KW",
            command=["design_kw", "<template_file>", "<result_file>", "--validate=<VALIDATE>"],
            default_mapping={"<VALIDATE>": "false"},
        )
```

jonathan-eq:
> Then it should be possible to use
> 
> ```python
>         super().__init__(
>             name="DESIGN_KW",
>             command=["design_kw", "<template_file>", "<result_file>", "--validate=<VALIDATE>"],
>             default_mapping={"<VALIDATE>": "false"},
>         )
> ```

Yes, but do we want to pass it to the script/have it as a script argument if it is unused?

eivindjahren:
> > Then it should be possible to use
> > ```python
> >         super().__init__(
> >             name="DESIGN_KW",
> >             command=["design_kw", "<template_file>", "<result_file>", "--validate=<VALIDATE>"],
> >             default_mapping={"<VALIDATE>": "false"},
> >         )
> > ```
> 
> Yes, but do we want to pass it to the script/have it as a script argument if it is unused?

Yes, that seems unproblematic.

eivindjahren (approved):
:+1:

PR #652, opened 2024-10-28
Avoid misuse of pytest.mark.xfail in pyscal test

berland:
Test this SystemExit explicitly instead as it is not going to change.


eivindjahren (approved):
:+1:

PR #654, opened 2024-10-28
Pin polars

berland:


yngve-sk (approved):
LGTM, consider putting the sorting of pyproject.toml to a separate commit for smaller diff

PR #655, opened 2024-11-21
Add fmudesign

dafeda:
Must be delete from `fmu-tools`: https://github.com/equinor/fmu-tools/pull/261


src/semeio/fmudesign/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+from semeio.fmudesign._tornado_onebyone import calc_tornadoinput
+from semeio.fmudesign.create_design import DesignMatrix
+
+__all__ = [
+    "summarize_design",
+    "calc_tornadoinput",
+    "DesignMatrix",
+    "excel2dict_design",
+    "inputdict_to_yaml",
+]
```
sondreso, line 19: I think overall the changes are good, my only worry is that people will have to change to `from semeio.fmudesign import ...` if they are importing this functionality in their scripts directly. Do we have any idea if people are doing this at all, and if yes to what extent? :thinking:

dafeda: @tralsos do you have any comments here?

sondreso: See this discussion: https://github.com/equinor/fmu-tools/pull/246#pullrequestreview-2413109007

PR #656, opened 2024-11-26
Add nearest correlation matrix function

tommyod:
- Implements the H-weighted (elementwise weighting) nearest correlation matrix in function `nearest_correlation_matrix()`
- Remove old function `_nearest_positive_definite()`
- Add back diagonal check == 1.0 in Iman Conover, since it now passes for inputs from this new function (whereas before it failed on output from `_nearest_positive_definite()`)
- Replace usage in the main function

I am unsure about these lines. I added them as an extra fail-safe, but adding some slack on the constraint `X >> 0` (X must be in the cone of positive symmetric definite matrices) by adjusting it to `X - eps * I >> 0` seems to work well. So the code below is never triggered on my random test cases. That being said, if it triggers it could save a user from an unfortunate situation in the future. Any thoughts on this would be appreciated. If you are agnostic, then I say we keep it - it makes a the code more robust at a not-too-high cost. I have no strong opinion however.

```python
  # We might get small eigenvalues due to numerics. Attempt to fix this by
  # recursively calling the solver with smaller values of epsilon. This is
  # an extra fail-safe that is very rarely triggered on actual data.
  is_symmetric = np.allclose(X, X.T)
  is_PD = np.linalg.eig(X).eigenvalues.min() > 0
  if not (is_symmetric and is_PD) and (eps > 1e-14):
      if verbose:
          print(f"Recursively calling solver with eps := {eps} / 10")
      return nearest_correlation_matrix(G, weights=H, eps=eps / 10, verbose=verbose)

  return X

```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
     "segyio",
     "xlrd",
     "xtgeo>=2.15",
+    "cvxpy",
 ]
 
 [project.urls]
--- a/src/semeio/fmudesign/create_design.py
+++ b/src/semeio/fmudesign/create_design.py
@@ -6,8 +6,8 @@ from collections import OrderedDict
 from datetime import datetime
 from pathlib import Path
 
+import cvxpy as cp
 import numpy as np
-import numpy.linalg as la
 import pandas as pd
 import scipy
 from scipy.stats import qmc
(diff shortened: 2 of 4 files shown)


dafeda:
This looks great.
As for the recursion, I don't mind it at all, but should we consider adding max recursion depth somehow?

dafeda (approved):

PR #657, opened 2024-11-26
Add Iman-Conover transformation

tommyod:
Updates the Iman-Conover transformation code. Non-functional changes: notation and using a class instead of a function.

Functional changes:
- `generate_van_der_waerden_scores` removed, replaced with vectorized version
- compute `S = P @ np.linalg.inv(Q)  # S = PQ^(-1)` without forming the inverse explicitly
- remove option `variance_reduction = False` (not in use, and likely will not be used)

I had to comment out a check for all ones on the diagonal user-supplied correlation matrix. Tests failed. This check will be added back in in the future, when we fix the `nearest correlation matrix` routine.


dafeda (approved):
Very nice 👏

PR #660, opened 2024-11-27
Use Latin Hypercube Sampling with Iman-Conover transformation

dafeda:
Resolves: https://github.com/equinor/semeio/issues/665

# Approach

- Switches from Monte Carlo to Latin Hypercube sampling
- Uses Iman-Conover instead of directy Cholesky to induce correlations. Iman-Conover preserver marginal distributions.
- Works directly with correlation matrices instead of covariance matrices, simplifying the code
- For uncorrelated variables, uses `LatinHypercube` with dimension=1, equivalent to stratified sampling

## Comparisons

Here I compare MC with LHS with n=30 for a few distributions. I've also added results from @Risk, but note that the seed is not the same.

image
image

image
image

## Simulating correlated variables with the Cholesky factorization

https://mlisi.xyz/post/simulating-correlated-variables-with-the-cholesky-factorization/

## Allowing correlations for discrete variables

https://kb.palisade.com/index.php?pg=kb.page&id=72


src/semeio/fmudesign/create_design.py
```diff
@@ -711,25 +835,36 @@ def generate(self, realnums, parameters, seedvalues, corrdict, rng):
...
-                    cov_matrix = design_dist.make_covariance_matrix(df_correlations)
-                    normalscoremeans = len(multivariate_parameters) * [0]
-                    normalscoresamples = rng.multivariate_normal(
-                        normalscoremeans, cov_matrix, size=numreals
-                    )
-                    normalscoresamples_df = pd.DataFrame(
-                        data=normalscoresamples, columns=multivariate_parameters
+                    correlations = df_correlations.values
+
+                    # Make correlation matrix symmetric
```
tommyod, line 840:
```suggestion
                    # Make correlation matrix symmetric by copying lower triangular part
```

src/semeio/fmudesign/design_distributions.py
```diff
@@ -147,17 +147,34 @@ def _check_dist_params_logunif(dist_params):
...
     """
+    if numreals < 0:
+        raise ValueError("numreal must be a positive integer")
+
+    if numreals == 0:
+        return np.array([])
+
+    sampler = qmc.LatinHypercube(d=1, seed=rng)
+    samples = sampler.random(n=numreals)
+    scaled = qmc.scale(samples, 0, 1)
```
tommyod, line 173: not sure if this line does anything, since samples are already in the `[0, 1)` interval?

src/semeio/fmudesign/design_distributions.py
```diff
@@ -166,33 +183,36 @@ def draw_values_normal(dist_parameters, numreals, rng, normalscoresamples=None):
     mean = float(dist_parameters[0])
     stddev = float(dist_parameters[1])
 
-    if len(dist_parameters) == 2:
+    if len(dist_parameters) == 2:  # normal
         if normalscoresamples is not None:
-            return scipy.stats.norm.ppf(
+            values = scipy.stats.norm.ppf(
```
tommyod, line 188: isn't this equal to `values = mean +  normalscoresamples * stddev`?

src/semeio/fmudesign/design_distributions.py
```diff
@@ -166,33 +183,36 @@ def draw_values_normal(dist_parameters, numreals, rng, normalscoresamples=None):
...
-        return scipy.stats.truncnorm.ppf(
-            scipy.stats.norm.cdf(normalscoresamples),
-            low,
-            high,
-            loc=mean,
-            scale=stddev,
-        )
+        else:
+            uniform_samples = generate_stratified_samples(numreals, rng)
+            values = scipy.stats.truncnorm.ppf(
```
tommyod, line 211: very nice. good to get rid of our own implementation and rely on scipy instead!

src/semeio/fmudesign/design_distributions.py
```diff
@@ -220,8 +239,10 @@ def draw_values_lognormal(dist_parameters, numreals, rng, normalscoresamples=Non
             scale=exp(mean),
         )
     else:
-        distribution = scipy.stats.lognorm(s=sigma, scale=exp(mean))
-        values = distribution.rvs(size=numreals, random_state=rng)
+        uniform_samples = generate_stratified_samples(numreals, rng)
+        values = scipy.stats.lognorm.ppf(
+            uniform_samples, s=sigma, loc=0, scale=exp(mean)
```
tommyod, line 244: nitpick: instead of using `exp` from math and everything else from scipy, use `np.exp` here while we're at it?

src/semeio/fmudesign/design_distributions.py
```diff
@@ -315,58 +342,49 @@ def draw_values_pert(dist_parameters, numreals, rng, normalscoresamples=None):
...
+        muval = (low + high + scale * mode) / (scale + 2)
+        if np.isclose(muval, mode):
+            alpha1 = (scale / 2) + 1
+        else:
+            alpha1 = ((muval - low) * (2 * mode - low - high)) / (
+                (mode - muval) * (high - low)
+            )
+        alpha2 = alpha1 * (high - muval) / (muval - low)
+
+        if normalscoresamples is not None:
```
tommyod, line 378: perhaps this IF can be written as:

```python
uniform_samples = generate_stratified_samples(numreals, rng) if (normalscoresamples is None) else normalscoresamples
values = scipy.stats.beta.ppf(uniform_samples, alpha1, alpha2)
```

or pulling the `values` out of the IF:

```python
if normalscoresamples is not None:
    uniform_samples = normalscoresamples 
else:
    uniform_samples = generate_stratified_samples(numreals, rng)

values = scipy.stats.beta.ppf(uniform_samples, alpha1, alpha2)
```

src/semeio/fmudesign/design_distributions.py
```diff
@@ -453,37 +468,46 @@ def draw_values(distname, dist_parameters, numreals, rng, normalscoresamples=Non
...
+            probabilities = [float(weight) for weight in weights]
+        except ValueError as e:
+            raise ValueError(
+                "All weights must be valid floating point numbers. "
+                f"Got weights: {weights}"
+            ) from e
+        # Just validate they're not negative
+        if any(p < 0 for p in probabilities):
+            raise ValueError("Weights cannot be negative")
+        fractions = [p / sum(probabilities) for p in probabilities]
```
tommyod, line 497: i would probably do this with numpy

tommyod (approved):
Looks good to me. Some nitpick here and there, some of my comments are not related to the code you wrote, but I just wrote a comment for it anyway.

src/semeio/fmudesign/design_distributions.py
dafeda: Yes I think so. Will change.

src/semeio/fmudesign/design_distributions.py
dafeda: I like the nitpick

src/semeio/fmudesign/design_distributions.py
dafeda: What about something like this:

```python
if normalscoresamples is not None:
    uniform_samples = scipy.stats.norm.cdf(normalscoresamples)
else:
    uniform_samples = generate_stratified_samples(numreals, rng)

values = scipy.stats.beta.ppf(uniform_samples, alpha1, alpha2)
```

src/semeio/fmudesign/design_distributions.py
dafeda: Yes sir, done

src/semeio/fmudesign/design_distributions.py
tommyod: yes. i had a mistake in my comment where i forgot to use CDF. this looks correct and the code looks better with the common part outside of the IF

dafeda: Alright, I've added the last version.

sondreso (approved):
🔢 🚀

PR #671, opened 2024-11-29
Run doctests in fmudesign as part of GH actions

tommyod:
In `fmudesign`, we added some doctests in previous PRs. They were not run, so I added a line in GH actions to run them. Some other examples did not actually run - here I simply replaced then `>>>` with `>>` syntax.


dafeda (approved):
Nice 👍

PR #673, opened 2024-12-02
Clean-up related to correlations

dafeda:


src/semeio/fmudesign/create_design.py
```diff
@@ -1020,9 +1013,9 @@ def _printwarning(corrgroup):
         "have to be included in correlation matrix. \n "
         "See documentation: \n"
         "https://equinor.github.io/fmu-tools/"
```
berland, line 1015: "fmu-tools" should probably not be here?

src/semeio/fmudesign/design_distributions.py
```diff
@@ -513,37 +514,28 @@ def is_number(teststring):
...
-            correlations = correlations.loc[
-                :, ~correlations.columns.str.contains("^Unnamed")
-            ]
-        else:
-            raise ValueError(
-                "Correlation matrix filename should be on "
-                "Excel format and end with .xlsx "
-            )
-    else:
+    if not str(excel_filename).endswith(".xlsx"):
```
berland, line 529: More canonical would be `if excel_filename.suffix != ".xlsx"`

src/semeio/fmudesign/create_design.py
dafeda: Actually, the docs are still in `fmu-tools` for now. Not been decided where to put them yet, so will leave this as is for now. But thanks.

src/semeio/fmudesign/design_distributions.py
dafeda: Thanks, but that causes a test to fail. I've fixed type hint to indicate that this can be both a str and Path.
Long way to go before this code-base is up to standard, but it's a start :)

berland (approved):
yay!

PR #675, opened 2024-12-06
Check if correlation matrix is consistent

dafeda:
Previous versions of fmudesign let the users know that a nearest correlation matrix is being used.
Bringing this functionality back in.
Set seed in design_input_onebyone to be able to check stdout in test.


src/semeio/fmudesign/create_design.py
```diff
@@ -17,6 +17,36 @@
 from semeio.fmudesign.iman_conover import ImanConover
 
 
+def is_consistent_correlation_matrix(matrix):
+    # Convert to numpy array if not already
+    matrix = np.array(matrix)
+
+    # Check if matrix is 2D and square
```
tommyod, line 24: this one is a bit different than the others. this is a true show-stopper, whereas the others can be "fixed". since `nearest_correlation_matrix` raises on this too, I would drop it from this function.

src/semeio/fmudesign/create_design.py
```diff
@@ -17,6 +17,36 @@
...
+    # Check elements are between -1 and 1
+    if np.any(matrix > 1 + 1e-8) or np.any(matrix < -1 - 1e-8):
+        raise ValueError("Correlation matrix elements must be between -1 and 1")
+
+    # Check positive semidefiniteness using Cholesky decomposition
+    try:
+        np.linalg.cholesky(matrix)
+    except np.linalg.LinAlgError:
+        print("Warning: Correlation matrix is not positive semidefinite")
+        return False
```
tommyod, line 45: the others checks `raise`, but this returns `False`. raise here too?

actually, the other conditions here would `raise`, and the matrix would not be fixed. there seems to be some inconsistencies here. if we can fix it we should, but we should print warnings. if there is no way to fix it, then we should `raise` on the outside.

we can also consider:
- print on "small errors"
- raise on "big errors" (diagonal is 1.0, not 1.000001)

tests/fmudesign/test_designmatrix.py
```diff
@@ -68,8 +69,39 @@ def test_endpoint(tmpdir, monkeypatch):
...
+    expected_output = dedent("""
+        Added sensitivity : seed
+        Added sensitivity : faults
+        Added sensitivity : velmodel
+        Added sensitivity : contacts
+        Added sensitivity : multz
+        Added sensitivity : sens6
+        Warning: Correlation matrix is not positive semidefinite
+        Input correlation matrix:
+        [[1. 0.9 0. 0. ]
```
tommyod, line 85: would the prints be prettier if the same number of digits was always used?
that way the columns will align.

src/semeio/fmudesign/create_design.py
```diff
@@ -17,6 +17,36 @@
...
+    # Check symmetry
+    if not np.allclose(matrix, matrix.T):
+        raise ValueError("Correlation matrix must be symmetric")
+
+    # Check diagonal elements are 1
+    if not np.allclose(np.diagonal(matrix), 1):
+        raise ValueError("Correlation matrix diagonal elements must be 1")
+
+    # Check elements are between -1 and 1
+    if np.any(matrix > 1 + 1e-8) or np.any(matrix < -1 - 1e-8):
```
tommyod, line 37: perhaps these errors should be propagated out to print with a more detailed description? 
```
element C[row=1, column=3] = 1.001, but every correlation should be in range [0, 1]
```

tommyod (approved):
I like it - very useful for the user to get feedback like this. In fact, maybe feedback should be even more detailed? See comments.

src/semeio/fmudesign/create_design.py
dafeda: Thanks, I've removed it.

src/semeio/fmudesign/create_design.py
dafeda: I have instead opted to just return True or False. I think it's cleaner. What do you think?

tests/fmudesign/test_designmatrix.py
dafeda: I've fixed the output to look like the following. What do you think?

```python
    expected_output = """Added sensitivity : seed
    Added sensitivity : faults
    Added sensitivity : velmodel
    Added sensitivity : contacts
    Added sensitivity : multz
    Added sensitivity : sens6

    Warning: Correlation matrix is not consistent
    Requirements:
    - Ones on the diagonal
    - Positive semi-definite matrix

    Input correlation matrix:
    [[1.00 0.90 0.00 0.00]
    [0.90 1.00 0.90 0.00]
    [0.00 0.90 1.00 0.00]
    [0.00 0.00 0.00 1.00]]

    Adjusted to nearest consistent correlation matrix:
    [[1.00 0.74 0.11 0.00]
    [0.74 1.00 0.74 0.00]
    [0.11 0.74 1.00 0.00]
    [0.00 0.00 0.00 1.00]]

    Added sensitivity : sens7
    Added sensitivity : sens8
    Provided number of background values (11) is smaller than number of realisations for sensitivity ('sens7', 'p10_p90') and parameter PARAM13. Will be filled with default values.
    Provided number of background values (11) is smaller than number of realisations for sensitivity ('sens7', 'p10_p90') and parameter PARAM14. Will be filled with default values.
    Provided number of background values (11) is smaller than number of realisations for sensitivity ('sens7', 'p10_p90') and parameter PARAM15. Will be filled with default values.
    Provided number of background values (11) is smaller than number of realisations for sensitivity ('sens7', 'p10_p90') and parameter PARAM16. Will be filled with default values.
    A total of 91 realizations were generated
    Designmatrix written to generateddesignmatrix.xlsx"""
```

src/semeio/fmudesign/create_design.py
dafeda: I removed this test as I think the definition of a consisten correlation matrix is that it's positive semi-definite and has ones on the diagonal. This is a different type of check I think. Might still be useful though. What do you think?

berland (approved):
👍🏻

PR #678, opened 2024-12-10
Require Python 3.11, update ruff and pyproject.toml

berland:


eivindjahren (approved):
Remember to check that the semeio version for python 3.8 in bleeding is fixed.

berland:
> Remember to check that the semeio version for python 3.8 in bleeding is fixed.

All of py38 is already purged from bleeding.

PR #681, opened 2025-01-08
Cast to float64 for xtgeo in ots

larsevj:
It appears that np.max returns np.float32 in numpy2 in ots.py, which is not compatible with xtgeo.

Tests passes locally with newest numpy and xtgeo.


berland (approved):
Lets try that!

PR #686, opened 2025-02-05
Make ahmanalysis not use libres facade

### LINKED ISSUE #685: AHMAnalysis not working after ErtConfig change
Workflows no longer get access to ErtConfig, meaning that [AHMAnalysis](https://github.com/equinor/semeio/blob/13710e63e67f563841efc5ab673bff238bd484e8/src/semeio/workflows/ahm_analysis/ahmanalysis.py#L121) no longer works. Most of the functionality should be possible to get by just using `ensemble` instead.

yngve-sk:
Resolves #685


src/semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -98,27 +105,49 @@
...
-        group_by="data_key",
-        output_dir=None,
-    ):
-        # (SemeioScript wraps this run method)
-
+        storage: Storage,
+        config_file: str,
+        target_name: str = "analysis_case",
+        prior_name: str | None = None,
+        group_by: str = "data_key",
```
oyvindeide, line 119: These would be workflow args normally?

src/semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -152,16 +181,34 @@ def run(
         except KeyError as err:
             raise ValidationError(f"Empty prior ensemble: {err}") from err
 
+        reports_dir = (
+            storage.path.parent
+            / "reports"
+            / Path(config_file).stem
+            / self.ensemble.name
+            / "AhmAnalysisJob"
```
oyvindeide, line 189: This could probably be a fixture from `ErtScript`?

src/semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -152,16 +181,34 @@ def run(
...
+            p.name
+            for p in experiment.parameter_configuration.values()
+            if isinstance(p, GenKwConfig)
+        ]
         if field_parameters:
             logger.warning(
                 f"AHM_ANALYSIS will only evaluate scalar parameters, skipping: {field_parameters}"
             )
 
-        scalar_parameters = sorted(self.facade.get_gen_kw())
```
oyvindeide, line 164: Nice! This was probably the last usage of that function?

tests/workflows/ahm_analysis/snapshots/test_integration/test_ahmanalysis_run_deactivated_obs/ks_df
```diff
@@ -1,11 +1,11 @@
...
 1,SNAKE_OIL_PARAM:BPR_555_PERSISTENCE,0.2,0.2,0.2,0.2
 2,SNAKE_OIL_PARAM:OP1_DIVERGENCE_SCALE,0.4,0.4,0.4,0.4
-3,SNAKE_OIL_PARAM:OP1_OCTAVES,0.2,0.2,0.2,0.2
-4,SNAKE_OIL_PARAM:OP1_OFFSET,0.3,0.3,0.3,0.3
+3,SNAKE_OIL_PARAM:OP1_OCTAVES,0.3,0.3,0.3,0.3
+4,SNAKE_OIL_PARAM:OP1_OFFSET,0.1,0.1,0.1,0.1
 5,SNAKE_OIL_PARAM:OP1_PERSISTENCE,0.2,0.2,0.2,0.2
 6,SNAKE_OIL_PARAM:OP2_DIVERGENCE_SCALE,0.2,0.2,0.2,0.2
-7,SNAKE_OIL_PARAM:OP2_OCTAVES,0.2,0.2,0.2,0.2
+7,SNAKE_OIL_PARAM:OP2_OCTAVES,0.3,0.3,0.3,0.3
```
oyvindeide, line 9: Why do the values change? Then rng?

tests/workflows/ahm_analysis/test_integration.py
```diff
@@ -68,8 +68,7 @@ def test_ahmanalysis_run_deactivated_obs(snake_oil_facade, snapshot, caplog):
     We simulate a case where some of the observation groups are completely
     disabled by outlier detection
     """
-
-    snake_oil_facade.config.analysis_config.observation_settings.alpha = 0.1
+    # Note: Unattainable run unless we can pass alpha to the workflow somehow
```
oyvindeide, line 71: Outdated comment?

src/semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -98,27 +105,49 @@
...
         self,
-        target_name="analysis_case",
-        prior_name=None,
-        group_by="data_key",
-        output_dir=None,
-    ):
-        # (SemeioScript wraps this run method)
-
+        storage: Storage,
+        config_file: str,
```
oyvindeide, line 116: Related to comment below, we might consider just passing an output path to workflows instead. Think that would be cleaner?

src/semeio/workflows/ahm_analysis/ahmanalysis.py
yngve-sk: Leaving as defaulted kwargs for now as support for it is added on ert side

src/semeio/workflows/ahm_analysis/ahmanalysis.py
yngve-sk: Done @ ert side

src/semeio/workflows/ahm_analysis/ahmanalysis.py
yngve-sk: Yep, removing

src/semeio/workflows/ahm_analysis/ahmanalysis.py
```diff
@@ -98,37 +106,69 @@
...
-            self._reports_dir = output_dir
 
-        obs_keys = list(self.facade.get_observations().obs_vectors.keys())
-        key_map = _group_observations(self.facade, obs_keys, group_by)
+        if not workflow_args:
+            target_name = None
+            prior_name = None
+            group_by = None
+        else:
+            target_name, prior_name, group_by = workflow_args
```
oyvindeide, line 132: User could provide less than 3 args, so should also check the length and only pick out the ones provided.

oyvindeide:
```python
    target_name = workflow_args[0] if len(workflow_args) > 0 else "analysis_case"
    prior_name = workflow_args[1] if len(workflow_args) > 1 else None
    sort_by = workflow_args[2] if len(workflow_args) > 2 else "data_key"
    ```

**Review by oyvindeide — COMMENTED** [review]
LGTM! But can you also add an integration test where this runs from the `ert` `cli`?

**Review by oyvindeide — APPROVED** [review]
LGTM! Some minor comments, also the tests are failing, but looks like some uv problem?

File: tests/workflows/ahm_analysis/test_integration.py [hunk 8]
```diff
@@ -51,7 +52,7 @@ def test_ahmanalysis_run(snake_oil_facade):
     ks_df = pd.read_csv(output_dir / "ks.csv")
     for keys in ks_df["Parameters"].tolist():
         assert keys in parameters
-    assert ks_df.columns[1:].tolist() == group_obs
+    assert set(ks_df.columns[1:].tolist()) == set(group_obs)
```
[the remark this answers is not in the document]
yngve-sk: The previous order is a bit arbitrary, and the `All-` usually come last in the dataframe as a consequence, I would opt for changing the test to be order agnostic rather than changing the order, though changing the order would be possible by sorting explicitly in `make_obs_groups` and then just putting those in there. Having the `All*` rows come last seems more sensible than a global sort to me at the moment.

PR #690, opened 2025-03-12
Have design2params validate Ert design matrix

jonathan-eq:
Resolves https://github.com/equinor/ert/issues/10271
This commit makes design2params also validate Ert design matrix, and log warning if it would fail.

### DIFF
--- a/src/semeio/forward_models/design2params/design2params.py
+++ b/src/semeio/forward_models/design2params/design2params.py
@@ -4,6 +4,7 @@ from pathlib import Path
 
 import numpy as np
 import pandas as pd
+from ert.config.design_matrix import DesignMatrix as ErtDesignMatrix
 
 from semeio._exceptions.exceptions import ValidationError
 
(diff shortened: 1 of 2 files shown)


src/semeio/forward_models/design2params/design2params.py
```diff
@@ -23,6 +24,21 @@
...
+        ert_design_matrix = ErtDesignMatrix(
+            xlsfilename, designsheetname, defaultssheetname
+        )
+        if ert_design_matrix_error := ert_design_matrix._validate_design_matrix(
+            ert_design_matrix
+        ):
+            raise Exception("\n".join(ert_design_matrix_error))
+    except Exception as err:
+        logger.warning(
+            f"Ert DESIGN_MATRIX keyword would have failed with error: {str(err)}"
```
xjules, line 38: Can you make this more descriptive? Eg. DESIGN_MATRIX validation of  DESIGN2PARAMS failed with .... ?

jonathan-eq: I think we should convey that this is not an error or failure, but just a warning that it will be in the future 🤔

xjules:
> I strongly believe that this should be implemented in ERT not in semeio. We do not want to run this for each realization, just once. And the forward model should not be dependent on ERT if possible I believe.

Very good point @larsevj ! I've completely missed that part that this actually runs for each realization. Indeed we just want to parse the config containing the `forward model == DESIGN2PARAMS` and validate it with the DESIGN MATRIX

jonathan-eq:
Closing this due to the issue point out by @larsevj

PR #695, opened 2025-04-15
Converge on using ECLBASE instead of ECL_BASE

berland:
ECL_BASE is aliased to ECLBASE when Ert parses configs, but there is no good reason to rely on that.

Related to https://github.com/equinor/ert/issues/10215


andreas-el (approved):
💯

PR #698, opened 2025-05-05
Stop using enspath from LibresFacade

frode-aarstad:


yngve-sk:
Would replace:
```
@pytest.fixture()
def snake_oil_facade(
    copy_snake_oil_case_storage,
):  # pylint: disable=unused-argument
    yield LibresFacade.from_config_file("snake_oil.ert")
```

with something just returning an ERT config and getting things from there, instead of hardcoding the strings

frode-aarstad:
> Would replace:
> 
> ```
> @pytest.fixture()
> def snake_oil_facade(
>     copy_snake_oil_case_storage,
> ):  # pylint: disable=unused-argument
>     yield LibresFacade.from_config_file("snake_oil.ert")
> ```
> 
> with something just returning an ERT config and getting things from there, instead of hardcoding the strings

I think hardcoding of strings are ok in tests. Makes things more explicit

yngve-sk:
> > Would replace:
> > ```
> > @pytest.fixture()
> > def snake_oil_facade(
> >     copy_snake_oil_case_storage,
> > ):  # pylint: disable=unused-argument
> >     yield LibresFacade.from_config_file("snake_oil.ert")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > with something just returning an ERT config and getting things from there, instead of hardcoding the strings
> 
> I think hardcoding of strings are ok in tests. Makes things more explicit

I don't really think it is a good practice in this case and in principle, the one source of truth for this stuff should be the config.

frode-aarstad:
> > > Would replace:
> > > ```
> > > @pytest.fixture()
> > > def snake_oil_facade(
> > >     copy_snake_oil_case_storage,
> > > ):  # pylint: disable=unused-argument
> > >     yield LibresFacade.from_config_file("snake_oil.ert")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > with something just returning an ERT config and getting things from there, instead of hardcoding the strings
> > 
> > 
> > I think hardcoding of strings are ok in tests. Makes things more explicit
> 
> I don't really think it is a good practice in this case and in principle, the one source of truth for this stuff should be the config.

I still think its fine to use hardcoded strings in tests

yngve-sk:
> > > > Would replace:
> > > > ```
> > > > @pytest.fixture()
> > > > def snake_oil_facade(
> > > >     copy_snake_oil_case_storage,
> > > > ):  # pylint: disable=unused-argument
> > > >     yield LibresFacade.from_config_file("snake_oil.ert")
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > with something just returning an ERT config and getting things from there, instead of hardcoding the strings
> > > 
> > > 
> > > I think hardcoding of strings are ok in tests. Makes things more explicit
> > 
> > 
> > I don't really think it is a good practice in this case and in principle, the one source of truth for this stuff should be the config.
> 
> I still think its fine to use hardcoded strings in tests

The way we refer to the storage path elsewhere in the code is always through config.ens_path. It should be possible to read this test without needing to read the ERT config file to understand it. The ERT config abstracts over these things so we don't have to do this. This is duplication of data without any benefit. It is more implicit because it has to be in sync with a string in an external file. To know that the hardcoded path makes sense, you have to go to the ERT config and verify that it matches the same thing, not ideal. If the path were in the config were to change, this path would change too. If we were to do this in all the tests in ERT surely it wouldn't be a good idea, therefore this seems more like a small step towards less code quality. The LibresFacade fixture was there for a reason, and just using an ERT config instead would let us remove it, and still keep this abstraction.

tests/workflows/ahm_analysis/test_ahm_analysis.py
```diff
@@ -17,7 +17,7 @@ def test_make_update_log_df(snake_oil_facade, snapshot):
     Note that this is now a snapshot test, so there is no guarantee that the
     snapshots are correct, they are just documenting the current behavior.
     """
-    with open_storage(snake_oil_facade.enspath, "w") as storage:
+    with open_storage("storage/snake_oil/ensemble", "w") as storage:
```
oyvindeide, line 20: Suggest rewriting the fixture to not use LibresFacade at all and give ErtConfig, as all the other use cases in this test is just using the config. We will still depend on ErtConfig here, but at least we drop the dependence on LibresFacade.

PR #707, opened 2025-06-10
Add fix for python3.13 stripping leading whitespace in docstrings

larsevj:
Tested locally to work with python3.13.


berland:
So the earlier code exploited the fact that the indent from the docstring was usable as the indent for the yaml code block. That was not so pretty.

Is it better to explicitly add a line that removes any common leading whitespace using `textwrap.dedent()` (so that it happens for all Python versions) and then add the indent when we need?

berland (approved):
👍🏻

PR #710, opened 2025-07-01
Add type annotations

### LINKED ISSUE #709: Thoughts on static type checking?
Hi

What are your thoughts about adding type annotations to this library? From what i've used of this library, it seems very well documented with docstring, so modern IDE's can infer a lot of the types. However, the developer experience using `semeio` could be improved by having type annotations.

If this is something you would consider, i am more than happy to help :)

einarwar:
Closes #709 

This PR adds type annotations mainly to the function/method signatures. In some cases, some type annotations in the body is added. Also added a `py.typed` file that should be included in the package manifest, [see more info here](https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information). [`basedpyright`](https://docs.basedpyright.com/latest/benefits-over-pyright/baseline/) was used to analyze the codebase. I see that this has introduced a set of type violations that was already present in the repo.

I also noticed that docstrings used multiple formats, and not always was up to date with the actual signature. I would suggest adding [`pydoclint`](https://docs.astral.sh/ruff/rules/#pydoclint-doc) as a Ruff rule to keep docstrings consistent and updated

I have really no knowledge of this repo, so please feel free to correct me if i made any mistakes :)

### DIFF
--- a/src/semeio/_exceptions/exceptions.py
+++ b/src/semeio/_exceptions/exceptions.py
@@ -1,10 +1,10 @@
 class ConfigurationError(Exception):
-    def __init__(self, message, errors):
+    def __init__(self, message: str, errors: Exception):
         super().__init__(message)
         self.errors = errors
         self.message = message
 
-    def __str__(self):
+    def __str__(self) -> str:
         msg = f"{self.message}"
         for error in self.errors:
             msg += f"\n{error}"
--- a/src/semeio/forward_models/overburden_timeshift/ots.py
+++ b/src/semeio/forward_models/overburden_timeshift/ots.py
@@ -17,7 +17,6 @@ from resdata.gravimetry import ResdataSubsidence
 from resdata.grid import Grid
 from resdata.resfile import Resdata3DKW, ResdataFile
 from scipy.interpolate import CloughTocher2DInterpolator
-from xtgeo.surface.regular_surface import RegularSurface
 
 from semeio._exceptions.exceptions import ConfigurationError
 from semeio.forward_models.overburden_timeshift import (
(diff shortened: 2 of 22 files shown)


src/semeio/_exceptions/exceptions.py
```diff
@@ -1,10 +1,10 @@
 class ConfigurationError(Exception):
-    def __init__(self, message, errors):
+    def __init__(self, message: str, errors: Exception):
```
einarwar, line 2: Comment: The one place i could find where this Exception was used, it seems like the `errors` parameter were of class `ValidationError`, i.e subclass of `Exception`. However, the `__str__()` implementation suggests it should be an iterable. Is this intened or a bug?

src/semeio/forward_models/overburden_timeshift/ots.py
```diff
@@ -279,7 +305,12 @@ def geertsma_ts_simple(self, vintage_pairs):
             vintage_pairs, self.subsidence.eval_geertsma, "TS_SIMPLE"
         )
 
-    def _geertsma_ts_custom(self, vintage_pairs, subsidence_func, method_name):
+    def _geertsma_ts_custom(
+        self,
+        vintage_pairs: Sequence[ConstrainedList],
+        subsidence_func: Callable[..., float],
+        method_name: str,
+    ) -> tuple[Literal[0], list[Any]] | list[xtgeo.RegularSurface]:
```
einarwar, line 313: Comment: The typing here is a bit wonky. The reason for this is that if `len(vintage_pairs) < 1` one returns a tuple `0, []`, which is not really consistent with the `list[xtgeo.RegularSurface]`. This typing also propagates since other functions use this function as well. This should probably raise an error instead, so the type of the output is `list[xtgeo.RegularSurface]`

PR #714, opened 2025-07-02
Add type annotations to `semeio.fmudesign`

einarwar:
This PR adds type annotations mainly to the function/method signatures of the `fmudesign` module. In some cases, some type annotations in the body is added. 

I have really no knowledge of this repo, so please feel free to correct me if i made any mistakes :)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,17 +1,17 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
+  - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v5.0.0
     hooks:
-    -   id: check-json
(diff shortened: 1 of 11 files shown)


einarwar:
Seems relevant to fix the `ANN` violations in the `fmudesign` in this PR. One could also add the following section to `pyproject.toml` to apply `ANN` ruleset to only the `fmudesign` directory:

```python
[tool.ruff.lint.per-file-ignores]
"!*/semeio/fmudesign/**.py" = ["ANN"]
```

Since ruff supports negated patterns, rules can be ignored except for the chosen paths, that way one can gradually apply and enforce rules. Do you want me to add that in this PR, or do you rather want to apply the rule once all existing violations are fixed?

PR #723, opened 2025-08-19
Use positive logic in excel2dict for readability

dafeda:


berland (approved):
Looks fine. Some changes to make mypy happy can be in a separate commit.

PR #724, opened 2025-08-20
Refactor

dafeda:


dafeda:
> I think this could benefit from being split into multiple commits. Not all changes here relates to using more idiomatic pandas.

I could rename the commit to something more generic. Don't really want to spend too much time splitting up these into separate commits.

src/semeio/fmudesign/_excel2dict.py
```diff
@@ -297,63 +297,66 @@ def _excel2dict_onebyone(
...
+            )
 
-    # Read decimals
     if "decimals" in designinput:
-        inputdict["decimals"] = OrderedDict()
-        for row in designinput.itertuples():
-            if _has_value(row.decimals) and _is_int(row.decimals):
-                inputdict["decimals"][row.param_name] = int(row.decimals)
+        # Convert to numeric, then filter for integers
+        numeric_decimals = pd.to_numeric(designinput["decimals"], errors="coerce")
```
berland, line 312: This decimals feature has only one integration test, thus not properly unit tested. It is more dangerous to refactor, can you add some unit tests for the feature that works before and after?

dafeda: `decimals` column is specified in `design_input_mc_with_correls.xlsx` which is run by for example `test_generate_full_mc_snapshot` so it should catch any changes. Many things are lacking unit-tests in `fmudesign`.

PR #726, opened 2025-08-20
Replace numpy RandomState with Generator

dafeda:
- Better performance and statistical properties
- Thread-safe unlike legacy RandomState
- RandomState is deprecated and in maintenance mode only


dafeda:
> This is breaking only with respect to exact reproduction of values from prior semeio versions?
> 
> I assume a user would have to work hard to figure out anything has changed.

Users will get different results even when using the same seed.

berland (approved):

PR #729, opened 2025-08-22
Convert triang dist validation to exception-based approach

dafeda:


src/semeio/fmudesign/design_distributions.py
```diff
@@ -82,27 +82,29 @@ def _check_dist_params_uniform(dist_params: Sequence[str]) -> tuple[bool, str]:
     return status, msg
 
 
-def _check_dist_params_triang(dist_params: Sequence[str]) -> tuple[bool, str]:
+def validate_triangular_params(
```
berland, line 85: Should we call it `parse_triangular_params` instead, or `parse_and_validate_triangular_params`?

Then you do not need the docstring, which is unclear about what kind of convertion it is doing.

dafeda: Not sure `parse` is a good name since there's no parsing going on. But perhaps we should remove the docstring as it's not really necessary.

berland: It is "parsing" from str|int|float into float

berland: At least it is doing more than validation

src/semeio/fmudesign/design_distributions.py
```diff
@@ -82,27 +82,29 @@ def _check_dist_params_uniform(dist_params: Sequence[str]) -> tuple[bool, str]:
...
-        msg = "Triangular distribution must have: low <= mode <= high. "
-    else:
-        status = True
-        msg = ""
 
-    return status, msg
+    try:
+        low, mode, high = [float(p) for p in dist_params]
+    except (ValueError, TypeError) as e:
+        raise ValueError("All parameters must be convertible to numbers") from e
```
berland, line 97: maybe nice for the user to have the "got [{low}, {mode}, {high}]" present in all exceptions?

dafeda: I have added this.

PR #730, opened 2025-08-22
Convert uniform dist validation to exception-based approach

dafeda:


tests/fmudesign/test_design_distributions.py
```diff
@@ -46,20 +46,36 @@ def test_check_dist_params_lognormal():
...
-def test_check_dist_params_uniform():
-    """Test lognormal dist param checker"""
-    assert not dists._check_dist_params_uniform([])[0]
-    assert not dists._check_dist_params_uniform(())[0]
-
-    assert not dists._check_dist_params_uniform([0])[0]
-    assert not dists._check_dist_params_uniform([0, 0, 0])[0]
+def test_parse_and_validate_uniform_params():
+    # Test wrong number of parameters
+    with pytest.raises(
```
berland, line 51: I think each of these should constitute their own `def test_something_descriptive`.

PR #732, opened 2025-08-25
Convert lognormal dist validation to exception-based approach

dafeda:


tests/fmudesign/test_design_distributions.py
```diff
@@ -81,20 +81,58 @@ def test_that_invalid_truncation_bounds_raise_ordering_error(self):
...
+    def test_that_non_numeric_parameters_raise_conversion_error(self):
+        with pytest.raises(
+            ValueError,
+            match=r"All parameters must be convertible to numbers. Got: \['mean', 'mu'\]",
+        ):
+            dists.parse_and_validate_lognormal_params(["mean", "mu"])
 
-    assert not dists._check_dist_params_lognormal(["mean", "mu"])[0]
+    def test_that_valid_lognormal_parameters_return_float_tuple(self):
+        assert dists.parse_and_validate_lognormal_params([0, 1]) == (0.0, 1.0)
```
berland, line 121: `"0"` and `"1"` are also valid parameters in parser code, but untested.

dafeda: Good catch, I've added a test.

PR #738, opened 2025-09-02
Enforce keyword args on calls. Some misc simplifications

tommyod:


src/semeio/fmudesign/fmudesignrunner.py
```diff
@@ -77,24 +77,20 @@ def main() -> None:
...
             f'Identical name "{args.config}" have been provided for the input'
             "file and the output file. "
-            " Exiting....."
         )
 
     design = DesignMatrix()
 
     design.generate(input_dict)
 
-    Path(args.destination).parent.mkdir(exist_ok=True, parents=True)
```
berland, line 96: Is this a behavioural change, or is this removal cancelled by the addition in `to_xls()` above?
I would have like to see this change isolated in its own commit with its own  comment (to ease future debugging).

tommyod: I moved it to `to_xls()`, so no behavioural change.

berland: Yes, I would like commits that describe what is being performed from the change, making the history readable. We try to follow https://cbea.ms/git-commit/

berland: (make it into a separate commit if it is unrelated to other changes in the code only. But it should be possible to read the commit log and check the diff in the future, and quickly understand which changes belong to which commit comment)

PR #739, opened 2025-09-02
Refactor distributions

tommyod:
- Refactor into classes to pull common validation into one parent class
- Better error messages, e.g. 

```python
    >>> Normal(mean=0, stddev=2, low=10, high=5)
        ...
    ValueError: Must have high > low in: Normal(mean=0.0, stddev=2.0, low=10.0, high=5.0)
```
- Match distributions with `distname.startswith("norm")` instead of `distname[0:4] == "norm"` - fewer indices to worry about
- Removed many "trivial tests", such as 

```python
    def test_that_insufficient_parameters_raises_count_error(self):
        with pytest.raises(
            ValueError,
            match="Loguniform distribution requires exactly 2 parameters, got 1",
        ):
            dists.parse_and_validate_loguniform_params([1])

    def test_that_excess_parameters_raises_count_error(self):
        with pytest.raises(
            ValueError,
            match="Loguniform distribution requires exactly 2 parameters, got 3",
        ):
            dists.parse_and_validate_loguniform_params([1, 2, 3])
```
since these tests basically just test the built-in `len`. Most non-trivial tests are kept. Some doctests were added too.


src/semeio/fmudesign/design_distributions.py
```diff
@@ -14,405 +13,299 @@
...
 
-    return low, mode, high
+    mean: float = 0.0
+    stddev: float = 1.0
+    low: float = -np.inf
+    high: float = np.inf
+
+    def __post_init__(self) -> None:
+        super().validate_params()
+        if self.stddev < 0:
```
dafeda, line 89: Could this check be in `validate_params` as well?

dafeda: Nevermind

dafeda (approved):

PR #747, opened 2025-09-08
Param names, unused comments, etc

tommyod:


src/semeio/fmudesign/design_distributions.py
```diff
@@ -289,6 +289,13 @@ def draw_values(
     if len(normalscoresamples) == 0:
         return np.array([])
 
+    match_names = {"disc", "const", "norm", "logn", "unif", "triang", "pert", "logunif"}
+    matches = [distname.lower().startswith(match_name) for match_name in match_names]
+    if sum(match_ for match_ in matches if match_) > 1:
+        raise ValueError(
+            f"Distribution prefix {distname} matched on several: {match_names}"
```
berland, line 296: (this could be accompanied by a test, e.g. by testing that `log` as a distribution would trigger this exception)

src/semeio/fmudesign/design_distributions.py
```diff
@@ -431,11 +431,20 @@ def read_correlations(excel_filename: str | Path, corr_sheet: str) -> pd.DataFra
...
     # Remove any 'Unnamed' columns that Excel/pandas may have automatically added.
-    correlations = correlations.loc[:, ~correlations.columns.str.contains("^Unnamed")]
-
-    return correlations
+    df_correlations = df_correlations.loc[
+        :, ~df_correlations.columns.str.contains("^Unnamed")
+    ]
+
+    columns, index = list(df_correlations.columns), list(df_correlations.index)
+    if columns != index:
```
berland, line 444: Looks like a nice extra validation. Can we test it?

src/semeio/fmudesign/create_design.py
```diff
@@ -507,7 +507,7 @@ def _fill_with_defaultvalues(self) -> None:
                 self.designvalues[key] = self.designvalues[key].fillna(
                     self.defaultvalues[key]
                 )
-            elif key not in ["REAL", "SENSNAME", "SENSCASE", "RMS_SEED"]:
+            elif key not in {"REAL", "SENSNAME", "SENSCASE", "RMS_SEED"}:
```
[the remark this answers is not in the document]
tommyod: This is rule [PLR6201](https://docs.astral.sh/ruff/rules/literal-membership/), but I do not know if we want to enable it. It's part of preview, so we either have to (1) enable all preview rules or (2) start adding specific rules. We could enable all preview rules, on the repo, but IMO that's a different PR

src/semeio/fmudesign/design_distributions.py
tommyod: Seems too trivial to test IMO.

PR #755, opened 2025-09-18
Remove MANIFEST.in file

jonathan-eq:
This file seems to no longer be used

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-include semeio/forward_models/config_forward_models/*


berland (approved):
This file is probably used by setuptools, but the files referred to does not exist -> commit message is not entirely correct.

jonathan-eq:
> This file is probably used by setuptools, but the files referred to do not exist -> commit message is not entirely correct.

There is no commit description, that is just on the PR

berland:
> There is no commit description, that is just on the PR

Ah, but then it can go into the commit message :)

berland:
Saying "Remove MANIFEST.in" in the commit message does not add any more information than what is already listed by git history on this commit, so you can use it to make it even clearer.

PR #759, opened 2025-09-22
Disallow merged cells in Excel input for fmudesign

### LINKED ISSUE #667: fmudesign: Detect merged cells in design matrix
If users have merged Excel cells in their input to the design matrix, this can lead to hard-to-understand error messages.
It is possible to detect merged cells in an Excel sheet and warn the user that this is not recommended, and unexpected behaviour should be expected.

tommyod:
Closes #667

### DIFF
--- a/src/semeio/fmudesign/_excel2dict.py
+++ b/src/semeio/fmudesign/_excel2dict.py
@@ -36,6 +36,7 @@ def excel2dict_design(
     """
 
     # Find sheets
+    _assert_no_merged_cells(input_filename)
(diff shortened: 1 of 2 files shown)


tommyod:
> What happens for users with xls files that have accicentally(?) merged some cells and using those xls files in their FMU setups? Could there be a situation where the merged cells are not involved in the operations and thus it would work fine, but break now?

It would fail. Yes, that is possible. As such it is a breaking change, but not a breaking change wrt expected, correct input. So I would consider it a bugfix.

Good that you did a check!

PR #773, opened 2025-10-07
Seed all random number generators in testing

tommyod:
Using RNG without seeds causes tests to fail every now and then. To test statistical properties, it is better to see the rng and test on many seeds. That way we can replicate across machines and we don't have to be stopped by false positives (what looks like test failures are just a bad rng).

### DIFF
--- a/tests/fmudesign/test_design_distributions.py
+++ b/tests/fmudesign/test_design_distributions.py
@@ -79,8 +79,9 @@ def test_draw_values_pert(seed):
     assert np.isclose(values.mean(), 5, atol=0.55)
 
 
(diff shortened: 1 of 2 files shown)


lucaeg (approved):
This looks good to me :) I only have two questions:

- Why use only one seed in some tests and a collection of seeds in others?
- How does using seeds exclude bad RNGs (that lead to test failures), do you mean that if a test fails for a particular seed then we will exclude that seed?

tommyod:
- depends on the test and its purpose!
- nope! in that case I would tweak tolerances!

PR #802, opened 2025-10-16
Fix distribution seed

### LINKED ISSUE #798: Always seed in fmudesign (?)
If "distribution_seed" is not given, then the default is silently set to None.

This means that two runs use two different RNGS, which produce two different results.

**Current behavior:**
```
distribution_seed = None    => do not seed the RNG
<no distribution_seed set>  => do not seed the RNG
distribution_seed = integer => seed the RNG
```

It this what we want? I would prefer a more **explicit behavoir**:
```
distribution_seed = None    => do not seed the RNG
<no distribution_seed set>  => (a) raise error or (b) set seed to e.g. 42 and report this
distribution_seed = integer => seed the RNG
```

lucaeg:
This PR forces `distribution_seed` to be set, either to a number (e.g. `38503`) or `None`. Before, users could avoid setting `distribution_seed` altogether, which would implicitly set `None` and make the analysis reproducible. We want to avoid this.

Closes #798

### DIFF
--- a/src/semeio/fmudesign/_excel_to_dict.py
+++ b/src/semeio/fmudesign/_excel_to_dict.py
@@ -241,7 +241,13 @@ def _excel_to_dict_onebyone(
         output[key] = int(generalinput[key])
     except KeyError as err:
         raise ValueError(
-            f"Missing key: {key}, please specify a value.\nValid types: None or class int."
+            "You did not specify a value for 'distribution_seed', which is used to seed "
+            "the random number generator that draws from distributions in Monte Carlo "
+            "sensitivities.\n"
+            "- Specify a number (e.g. a 6 digit integer) to seed the random number "
+            "generator and obtain reproducible results.\n"
+            "- Specify None if you do not want to seed the random number generator. "
+            "Your analysis will not be reproducible."
         ) from err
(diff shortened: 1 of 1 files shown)


src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -228,35 +228,39 @@ def _excel_to_dict_onebyone(
...
-    # Not reproducible right now since we set default_rng(None)
     key = "distribution_seed"
     try:
         output[key] = int(generalinput[key])
-    except (KeyError, TypeError):
-        output[key] = None
-    except ValueError:
+    except KeyError as err:
+        raise ValueError(
+            f"Missing key: {key}, please specify a value.\nValid types: None or class int."
```
tommyod: I think we should improve this, e.g.:

```
You did not specify a value for 'distribution_seed', which is 
used to seen the random number generator that draws from
distributions in Monte Carlo sensitivities.
- Specify a number (e.g. a 6 digit integer) to seed the random number generator and obtain reproducible results.
- Specify 'None' if you do not want to seed the random number generator. Your analysis will not be reproducible.
```

This is now enforced whereas it was not before, so users need a good message

PR #828, opened 2025-11-06
Add p10 p90 distributions

lucaeg:
Remaining:

- Go though all distributions and check consistent use of min/max vs low/high
- Add percentile distributions to documentation
- Add percentile distributions to test files
- Add pert_p10_p90
- Do we want to implement any other percentile distributions?


tommyod (approved):
Great! Starting to become a bit messy with all the cases and nested IFs, but I don't see any immediate way around it

PR #829, opened 2025-11-07
Use long options when building gendata command line

berland:
This makes jobs.json slightly more readable in runpaths.


andreas-el (approved):
Same saaaaame. 🏖️

PR #838, opened 2025-11-24
Fix sheet

lucaeg:


src/semeio/fmudesign/design_distributions.py
```diff
@@ -134,16 +135,25 @@ def to_probabilit(
 
     # Normal distribution
     if distname.lower().startswith("norm"):
+        if len(parameters) not in (2, 4):
+            raise ValueError(f"Normal must have 2 or 4 parameters, got: {parameters}")
```
SAKavli: Could it be nice to have `, got: {len(parameters)} ({parameters})"`? Idk how parameters looks like, but if the entries are long, a summary of how many could be nice first, as we are comparing lengths.

PR #879, opened 2026-04-13
Update development documentation to use uv

eivindjahren:


SAKavli:
Seems to me like Copilot makes some valid points

eivindjahren:
> Seems to me like Copilot makes some valid points

@SAKavli Fixed

SAKavli (approved):
🥓

PR #884, opened 2026-05-05
Fix OTS test

eivindjahren:


ajaust (approved):
Just let the tests run through and merge when they are green. 👍

ajaust (approved):

PR #889, opened 2026-05-12
Update rng for update strategy API

xjules:
A simple fix to account for a update strategy API


frode-aarstad (approved):

larsevj:
Need to require ert >= 22.0.0rc1

And update lockfile

PR #908, opened 2026-06-19
Fix observation name bug

### LINKED ISSUE #905: AHM_ANALYSIS fails if there is a colon in the observation name
**Describe the bug**
When running `AHM_ANALYSIS` with an observation name with `:` (and perhaps a combination with `-`/`_`, did not check in detail), the job will fail with a key error.

**To Reproduce**
In snake oil:

in snake_oil.ert:
```
CREATE_WORKFLOW_FROM_JOB ahm_analysis AHM_ANALYSIS
```
in the observations file modify one observation (note the `:` instead of `_`):

```
SUMMARY_OBSERVATION WOPR:OP1-190
{
    VALUE   = 0.015;
    ERROR   = 0.01;
    DATE    = 2015-03-15;  -- (RESTART = 190)
    KEY     = WOPR:OP1;
};
```

also reproduced in poly example, though the error message is a bit different

```
GENERAL_OBSERVATION POLY:MY-OBS {
   DATA       = POLY_RES;
   INDEX_LIST = 0,2,4,6,8;
   OBS_FILE   = poly_obs_data.txt;
};
```
in poly.ert:
```
CREATE_WORKFLOW_FROM_JOB ahm_analysis AHM_ANALYSIS
```

**Expected behavior**
This should work regardless of the observation name.

Stack trace:
```
Processing: SNAKE_OIL_WPR_DIFF
Processing: WOPR_OP1
The script 'AhmAnalysisJob' caused an error while running:
'Observations: WOPR_OP1-190 not in experiment'
  File "/path/to/ert/src/ert/config/ert_script.py", line 116, in initializeAndRun
    return self.run(*arguments)
           ^^^^^^^^^^^^^^^^^^^^
  File "/path/to/ert/.venv/lib/python3.11/site-packages/semeio/workflows/ahm_analysis/ahmanalysis.py", line 267, in run
    update_log = smoother_update(
                 ^^^^^^^^^^^^^^^^
  File "/path/to/ert/src/ert/analysis/_es_update.py", line 431, in smoother_update
    smoother_snapshot = perform_ensemble_update(
                        ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/ert/src/ert/analysis/_es_update.py", line 131, in perform_ensemble_update
    preprocessed_data = _preprocess_observations_and_responses(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/ert/src/ert/analysis/_update_commons.py", line 202, in _preprocess_observations_and_responses
    observations_and_responses = prior_ensemble.get_observations_and_responses(
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/ert/src/ert/storage/local_ensemble.py", line 981, in get_observations_and_responses
    raise KeyError(
KeyError: 'Observations: WOPR_OP1-190 not in experiment'
```

frode-aarstad:
Resolves #905 

The first commit is the actual fix. The second is extracting the problematic code into a new function and adding testing to it.

### DIFF
--- a/tests/workflows/ahm_analysis/test_ahm_analysis.py
+++ b/tests/workflows/ahm_analysis/test_ahm_analysis.py
@@ -1,8 +1,10 @@
 import copy
 import math
+from types import SimpleNamespace
 
 import numpy as np
 import pandas as pd
+import polars as pl
 import pytest
 from ert.analysis import build_strategy_map, smoother_update
 from ert.storage import open_storage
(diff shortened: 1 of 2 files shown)


tests/workflows/ahm_analysis/test_ahm_analysis.py
```diff
@@ -246,11 +248,116 @@ def test_raise_if_empty(misfit_data, prior_data, expected_msg):
...
+                        "observation_key": ["OBS:B", "OBS:A"],
+                        "response_key": ["RESP:1", "RESP:2"],
+                    }
+                ),
+            },
+            {
+                "RESP_1": ["OBS:B", "OBS:C"],
+                "RESP_2": ["OBS:A"],
+            },
+        ),
```
xjules, line 280: it would be nice to have an id too for the parametrization name. I guess the last one is a mixture of both types; `:` and `_`

xjules (approved):
I think it is correct. Just had a small comment on parametrization. Hence I think we should keep it as is with the tests and thus squash the commits.

tests/workflows/ahm_analysis/test_ahm_analysis.py
xjules: for instance in the first parametrization:
`id="response_renamed_for_data_key"`

PR #931, opened 2026-08-03
Add seed_strategy option for Monte Carlo sampling

### LINKED ISSUE #922: Add a sampling option that allows for per correlation group sampling
Currently all parameters are sampled jointly. If some parameters are added or removed, then all parameters are resampled, even across correlation groups. Potentially causing unnecessary disturbance.

larsevj:
Resolves #922

Sampling all parameters in one LHS call means adding, removing or reordering a single parameter reshuffles every other parameter, which makes design matrices hard to extend without invalidating existing runs.

The new 'seed_strategy' key in general_input selects between:

- 'joint' (default): existing behaviour, unchanged.
- 'independent': each uncorrelated parameter and each correlation group is keyed off distribution_seed, so editing one parameter leaves the others bit-identical. Correlation groups stay single sampling units, so requested correlations are induced as before.


src/semeio/fmudesign/config_validation.py
```diff
@@ -67,6 +67,24 @@ def validate_configuration(
             f"{key!r} must be a non-negative integer or None, got: {config[key]}"
         )
 
+    # 'seed_strategy' controls how Monte Carlo samples are seeded:
+    # - 'joint'       => all parameters drawn in one LHS call (the default).
+    #                    Adding/removing/reordering a parameter reshuffles all.
+    # - 'independent' => each parameter (or correlation group) is seeded
+    #                    separately, so changing one leaves the others stable.
+    key = "seed_strategy"
+    allowed_strategies = {"joint", "independent"}
```
SAKavli: Perhaps it would be nice with an enum for these strategies, as these strings are hardcoded multiple places now. 
Would be nice if those could be replaced with reference to a single enum.

src/semeio/fmudesign/create_design.py
```diff
@@ -35,6 +36,27 @@
 if TYPE_CHECKING:
     from collections.abc import Hashable, Sequence
 
+    # (group_name, correlation_matrix, member_parameter_names)
+    CorrelationGroup = tuple[str, pd.DataFrame, list[str]]
+
+
+def _derive_rng(base_seed: int, *keys: str) -> np.random.Generator:
+    """Return a numpy Generator seeded from ``base_seed`` and stable string keys.
```
SAKavli: What is meant by "stable string keys"?

src/semeio/fmudesign/create_design.py
```diff
@@ -59,6 +81,9 @@ def __init__(self, verbosity: int = 0, output_dir: Path | None = None) -> None:
         seedvalues: list of seed values
         verbosity: how much information to print
         output_dir: where to write debugging output and QC plots
+        rng: random number generator used by the 'joint' seed strategy
+        seed_strategy: 'joint' or 'independent' Monte Carlo seeding
+        base_seed: base seed each unit is keyed from under 'independent'
```
SAKavli: Is this accurate, as it also holds some other value for 'joint' strategy?

src/semeio/fmudesign/create_design.py
```diff
@@ -90,6 +118,18 @@ def generate(self, inputdict: dict[str, Any]) -> None:
         self.rng = np.random.default_rng(seed=inputdict.get("distribution_seed"))
         self.defaultvalues = inputdict["defaultvalues"]
 
+        # Seed strategy for Monte Carlo sampling. For the 'independent' strategy
+        # each sampling unit is seeded from a base seed, so that changing one
+        # parameter does not reshuffle the others. If no distribution_seed is
+        # given, self.rng is unseeded and the base seed differs between runs.
```
SAKavli: This is nice!
Could it make sense to replace the description in the docstring with this one?

src/semeio/fmudesign/create_design.py
```diff
@@ -785,6 +832,8 @@ def generate(
         corrdict: dict[str, Any] | None,
         rng: np.random.Generator,
         correlation_iterations: int = 0,
+        seed_strategy: str = "joint",
+        base_seed: int | None = None,
```
SAKavli, line 835: Is this ever `None`?

src/semeio/fmudesign/create_design.py
```diff
@@ -796,15 +845,31 @@ def generate(
...
+                  sensitivity name, and the parameter's own name, distribution
+                  and correlation group membership are unchanged.
+            base_seed (int): base seed used for the 'independent' strategy.
+              Required when ``seed_strategy='independent'``.
         """
         self.sensvalues = pd.DataFrame(
             columns=list(parameters.keys()), index=range(size)
         )
-        self.correlation_dfs_ = {}  # Store correlation matrices (dataframes)
+        self.correlation_dfs_: dict[str, pd.DataFrame] = {}  # correlation matrices
```
SAKavli, line 872: Unrelated to your PR:  Do you know why there's a trailing underscore here?

src/semeio/fmudesign/create_design.py
```diff
@@ -944,6 +946,165 @@ def generate(
...
+
+        Single-member groups are skipped (with a warning) and treated as
+        uncorrelated. Populates ``self.correlation_dfs_`` as a side effect.
+        Shared by both the 'joint' and 'independent' seed strategies.
+        """
+        if not corrdict:
+            return []
+
+        df_params = (
+            pd.DataFrame.from_dict(
```
SAKavli, line 970: I suppose this will be ported over to Ert, do we bother rewriting this already to polars or leave it be for later?

src/semeio/fmudesign/create_design.py
```diff
@@ -944,6 +946,165 @@ def generate(
...
+        groups = dict(iter(df_params.groupby("corr_sheet")))
+        groups.pop("nocorr", None)
+
+        loaded: list[CorrelationGroup] = []
+        for corr_group_name, corr_group in groups.items():
+            corr_group_name = cast("str", corr_group_name)
+
+            # A single correlation - print warning and skip it
+            if len(corr_group) == 1:
+                printwarning(corr_group_name)
```
SAKavli, line 990: What is the upperbound for number of correlation groups? 
This warning is quite long, could this become noisy?
Probably unrealistic to have single correlation for many correlation groups.

src/semeio/fmudesign/create_design.py
```diff
@@ -944,6 +946,165 @@ def generate(
...
+                continue
+
+            # Read correlation matrix and convert it to a proper matrix
```
SAKavli: What is meant by "proper" matrix here? Could this be elaborated in the comment?

src/semeio/fmudesign/create_design.py
```diff
@@ -944,6 +946,165 @@ def generate(
...
+            self.correlation_dfs_[corr_group_name] = df_correlations
+            loaded.append((corr_group_name, df_correlations, multivariate_parameters))
+
+        return loaded
+
+    @staticmethod
+    def _make_correlator(
+        correlation_iterations: int, rng: np.random.Generator
+    ) -> probabilit.correlation.Correlator:
+        """Iman-Conover, optionally followed by permutation iterations."""
```
SAKavli: Without context, I did not fully understand this docstring.
From the code, it looks like its something else than Iman Conover if correlation_iterations > 0?

src/semeio/fmudesign/create_design.py
```diff
@@ -944,6 +946,165 @@ def generate(
...
+        # Create a dummy NoOp node for sampling each parent distribution
+        expression = probabilit.modeling.NoOp(*distr_by_name.values())
+
+        for _name, df_correlations, members in corr_groups:
+            corrvars = [distr_by_name[name] for name in members]
+            expression.correlate(*corrvars, corr_mat=df_correlations.to_numpy())
+
+        correlator = self._make_correlator(correlation_iterations, rng)
+
+        # Sample the dummy node - this samples every parent and populates "samples_"
```
SAKavli: `samples_` of what object?
Could this be elaborated?

src/semeio/fmudesign/create_design.py
```diff
@@ -944,6 +946,165 @@ def generate(
...
+    ) -> None:
+        """Sample each correlation group and each uncorrelated parameter as a
+        separate unit, seeded from ``base_seed`` and a stable key. Adding,
+        removing or reordering a parameter therefore leaves the other
+        parameters unchanged (given a fixed seed and sample size).
+        """
+        grouped_params: set[str] = set()
+
+        # Each correlation group is one independently seeded unit.
+        for group_name, df_correlations, members in corr_groups:
```
SAKavli: Should we consider renaming the `members` reference to more specifically `parameters` or `member_params`?
I had to check what this third tuple value represents to understand this bit.

tests/fmudesign/snapshots/test_create_design/test_generate_full_mc_snapshot_independent/design_output_mc_with_correls.json
```diff
@@ -0,0 +1,9023 @@
...
+      "PARAM1": 0.036,
+      "PARAM2": -0.09,
+      "PARAM3": 2009830.0,
+      "REAL": 499.0,
+      "RMS_SEED": 1499.0,
+      "SENSCASE": "p10_p90",
+      "SENSNAME": "montecarlo"
+    }
+  ]
+}
```
SAKavli, line 9023: Will we be porting over this to ert?
I think @oyvindeide would like to avoid too many snapshot files laying around as much as possible, but if this is valuable we'll keep it.
Optionally we can keep this in semeio without porting over to ert.

tests/fmudesign/test_config_validation.py
```diff
@@ -0,0 +1,43 @@
+"""Tests for validation of the fmudesign config (the general_input sheet)."""
```
SAKavli: Is the description/title misleading for this file - as there is only tests regarding seeding strategies here?

tests/fmudesign/test_config_validation.py
```diff
@@ -0,0 +1,43 @@
...
+    return {
+        "designtype": "onebyone",
+        "repeats": 10,
+        "distribution_seed": 42,
+        "seeds": "default",
+        **extra,
+    }
+
+
+def test_seed_strategy_defaults_to_joint():
```
SAKavli, line 18: If porting over to ert later, it would be nice with the `test_that_` naming convention to avoid work later.

tests/fmudesign/test_config_validation.py
SAKavli: Optionally if existing validation tests should be moved here on some point?

tests/fmudesign/test_create_design.py
```diff
@@ -683,3 +710,400 @@ def test_fill_with_background_values_are_filled():
...
+# --------------------------------------------------------------------------
+# Monte Carlo seed strategies ('joint' default vs opt-in 'independent').
+#
+# 'joint' draws all parameters in one LHS call, so editing one parameter
+# reshuffles all of them. 'independent' seeds each parameter (and each
+# correlation group) separately, so unrelated parameters stay fixed.
+# --------------------------------------------------------------------------
+
+
+def _run(
```
SAKavli: Consider renaming to `_sample_mc` for less ambiguity in the code

tests/fmudesign/test_create_design.py
```diff
@@ -683,3 +710,400 @@ def test_fill_with_background_values_are_filled():
...
+        config["background"] = background
+    return config
+
+
+def test_default_strategy_is_joint():
+    """Omitting seed_strategy must equal explicitly passing 'joint'."""
+    params = {"A": ("normal", ["0", "1"], None), "B": ("uniform", ["0", "1"], None)}
+
+    def sample(**kwargs):
+        sens = MonteCarloSensitivity("s", verbosity=0)
```
SAKavli: Consider explicitly writing:
```py
sens = MonteCarloSensitivity(sensname="s", verbosity=0)
```
and also replacing "s" with "foo", as I was wondering if this was some important abbreviation for a mode - or something similar.

tests/fmudesign/test_create_design.py
```diff
@@ -683,3 +710,400 @@ def test_fill_with_background_values_are_filled():
...
+    assert np.isclose(n.mean(), 0.0, atol=0.1)
+    assert np.isclose(n.std(), 2.0, atol=0.1)
+    assert u.min() >= -5
+    assert u.max() <= 0
+    assert np.isclose(u.mean(), -2.5, atol=0.1)
+
+
+@pytest.mark.parametrize(("strategy", "stable"), EDIT_STABILITY)
+def test_adding_a_parameter(strategy, stable):
+    """'independent' leaves the other parameters untouched, 'joint' reshuffles."""
```
SAKavli: Perhaps it makes sense to split this into two tests with elaborated titles?

tests/fmudesign/test_create_design.py
```diff
@@ -1007,6 +1007,33 @@ def test_independent_group_may_share_a_name_with_a_parameter(tmp_path):
...
+            "corr2": (["Z", "W"], [[1, 0.8], [0.8, 1]]),
+        },
+    )
+    params = {
+        "X": ("normal", ["0", "1"], "corr1"),
+        "Y": ("normal", ["0", "1"], "corr1"),
+        "Z": ("normal", ["0", "1"], "corr2"),
+        "W": ("normal", ["0", "1"], "corr2"),
+    }
+    with pytest.raises(ValueError, match="'Z'"):
```
SAKavli: Why 'Z' - and not the others? Perhaps the match could be elaborated.
Also, if not necessary for testing overlapping correlation groups, the parameters set of `{X, Y, Z, W}` could be stripped down to a more concise example.

src/semeio/fmudesign/create_design.py
larsevj: I believe it is a sklearn convention, see https://scikit-learn.org/stable/developers/develop.html
Also used in probabilit.

src/semeio/fmudesign/create_design.py
larsevj: There is a separate issue for it, and I would think it makes the most sense to replace it in a different PR for all of fmu-design.

src/semeio/fmudesign/create_design.py
larsevj: This is the current behaviour, so will not become more noisy than before. When used with ERT we could probably give a more concise ConfigWarning or Error, later on.

src/semeio/fmudesign/create_design.py
larsevj: I assume they meant just convert to a numpy array. Will clean up.

src/semeio/fmudesign/create_design.py
larsevj: Also just a preexisting comment, but will try to clarify.

tests/fmudesign/snapshots/test_create_design/test_generate_full_mc_snapshot_independent/design_output_mc_with_correls.json
larsevj: I would think that as long as it is not binary files it should be okay, we already have a lot of snapshots in ert for every basemodel configs etc. Think it is valuable to "freeze" the output/behaviour, to guard against making unintended changes. Especially important now that we plan to do a lot of changes.

src/semeio/fmudesign/create_design.py
larsevj: Have rewritten now.

src/semeio/fmudesign/create_design.py
larsevj: Could be None for "joint" strategy, if called directly

tests/fmudesign/test_config_validation.py
larsevj: Agree, but will make an issue for it, since it concerns all fmu-design tests.

SAKavli (approved):
Nice work, @larsevj ⭐

PR #939, opened 2026-08-09
Setup rapid tests

SAKavli:


pyproject.toml
```diff
@@ -42,6 +42,10 @@ dependencies = [
     "probabilit>=0.4.1",
     "resfo-utilities>=0.1.1",
 ]
+[dependency-groups]
+dev = [
```
berland, line 46: these have typically been put in the optional dependencies, but maybe those (test dependencies) should be moved/merged into the dev dependency group (?)

SAKavli: Sure! If the separation `test, style, types` is not used for anything, might as well just merge them all into one.

.github/workflows/style.yml
```diff
@@ -35,7 +35,7 @@ jobs:
 
       - name: Install semeio with style deps
         run: |
-          uv sync --frozen --extra style
+          uv sync --frozen --all-extras
```
larsevj: extras are for optional dependencies. Since there are none left, this is a no-op. Should use groups instead, but the dev groups is included by default, so not strictly needed.

pyproject.toml
```diff
@@ -42,6 +42,27 @@ dependencies = [
...
+    'pytest-snapshot',
+    'pytest-xdist',
+    'rstcheck-core',
+    'rust-just',
+    'types-PyYAML',
+    'types-openpyxl',
+    'types-seaborn',
+    'types-setuptools',
+    'xlwt'
+]
```
larsevj, line 65: Will probably need an update to ci/testkomodo.sh and erts semeio test workflow to match the new dependency group instead of optional deps.

.github/workflows/style.yml
SAKavli: Ah, is this true?
When doing `uv sync` uv uninstalls packages like pytest, while `uv sync --all-extras` installs packages like pytest.

SAKavli: Yes, this is true

pyproject.toml
SAKavli: https://github.com/equinor/ert/pull/14132

larsevj (approved):
LGTM🚀 

(Could cleanup the `--all-extras` flag, but not strictly an issue.)

SAKavli:
Oops, forgot about that, will fix before merging!

PR #940, opened 2026-08-12
Report errors in background correlation setup

larsevj:
Errors while reading the background sheet were swallowed by a broad try/except, silently leaving all background parameters out of the design matrix. Correlation sheet mismatches and missing background sheets are now reported to the user.


SAKavli (approved):
Very nice!
I think we should spend some time on setting up a testing framework which is efficient to work with, as it is quite tedious to create all the sheets from scratch every time.

larsevj:
> Very nice! I think we should spend some time on setting up a testing framework which is efficient to work with, as it is quite tedious to create all the sheets from scratch every time.

Agree, can probably look to consolidate it when we try to migrate away from the excel test files.

PR #941, opened 2026-08-12
Refactor configuration validation

SAKavli:


src/semeio/fmudesign/config_validation.py
```diff
@@ -75,12 +89,30 @@ def validate_configuration(
...
+        - Checking that required keys exist
+        - Checking that values are set to valid types
+        - Setting default values if keys are not set
+
+    """
+    _validate_designtype(config)
+    _validate_repeats(config)
+    _validate_correlation_iterations(config, verbosity)
+    _validate_distribution_seed(config)
+    _validate_rms_seeds(config)
```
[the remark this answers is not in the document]
SAKavli: Sure why not, it's two lines reverted.

src/semeio/fmudesign/config_validation.py
```diff
@@ -62,11 +71,17 @@ def validate_configuration(
             "- Specify None if you do not want to seed the random number generator. "
             "Your analysis will not be reproducible."
         )
+
+    key = "distribution_seed"
+    if key not in config:
```
larsevj: this is checked above?  For consistency might move key = to the top as well?

src/semeio/fmudesign/config_validation.py
```diff
@@ -47,13 +55,14 @@ def validate_configuration(
         try:
             config[key] = int(config[key])
         except (ValueError, TypeError) as err:
-            raise ValueError(
+            raise ConfigValidationError(
                 f"{key!r} must be a non-negative integer, got: {config[key]}"
```
larsevj: Agree with the above, but should probably also consolidate the integer validation for `repeats`, `distribution_seed`, `rms_seed` and `correlation_iteration`.
Not sure if coercing to int is the answer, as it will allow for instance 10.5 -> 10, etc.

src/semeio/fmudesign/config_validation.py
```diff
@@ -7,28 +7,36 @@
...
-    This function is responsible for:
-        - Checking that required keys exist
-        - Checking that values are set to valid types
-        - Setting default values if keys are not set
-
-    """
-    config = copy.deepcopy(config)
 
+def _validate_designtype(config: dict[str, Any]) -> None:
+    if "designtype" not in config:
```
larsevj: Should probably have a test for this new check.

src/semeio/fmudesign/config_validation.py
larsevj: Should also add tests for this if done.

SAKavli:
Superseded by #950

PR #949, opened 2026-08-18
Cleanup tests

larsevj:


tests/fmudesign/test_create_design.py
```diff
@@ -23,7 +23,7 @@
 
 @pytest.mark.integration_test
 @pytest.mark.parametrize("correlations", [True, False])
-def test_distribution_statistis(tmpdir, monkeypatch, correlations):
+def test_distribution_statistics(tmp_path, correlations):
```
SAKavli, line 26: There is a `use_tmpdir` fixture if that makes life easier

SAKavli (approved):
Skimmed through the changes, looks very nice, @larsevj !

tests/fmudesign/test_create_design.py
larsevj: will keep in mind

PR #950, opened 2026-08-18
Add validation of parsed general input

SAKavli:
The existing validation is validating the config dict generated by various steps when parsing the excel file to a config dict.

It makes sense to have a validation step earlier in the loop which validates the raw input from the excel file to give a more relevant exception feedback to unchanged values.

This validation should not edit any values, only validate the input - except for validation of existing files relative to the excel `input_filename`


src/semeio/fmudesign/config_validation.py
```diff
@@ -60,29 +50,6 @@ def validate_configuration(
...
-            config[key] = int(config[key])
-        except (ValueError, TypeError) as err:
-            raise ValueError(
-                f"{key!r} must be a non-negative integer, got: {config[key]}"
-            ) from err
-
-    key = "distribution_seed"
-    if key not in config:
-        raise ValueError(
-            "You did not specify a value for 'distribution_seed', which is used to "
```
SAKavli, line 74: I thought it might be fine to remove this from the error message and consider adding it to the actual documentation document instead.

SAKavli:
I'm thinking the next step would be to create a class for GeneralInput to ease the validation and reduce the amount of dictionaries floating around.

src/semeio/fmudesign/general_input_validation.py
```diff
@@ -0,0 +1,175 @@
...
+
+def validate_file(maybe_file: Any, input_file: str, key: str) -> None:  # ruff: ignore[any-type]
+    try:
+        resolve_path(maybe_file, input_file)
+    except (ValueError, TypeError) as e:
+        raise GeneralInputError(
+            f"Could not resolve file '{maybe_file}' to an existing "
+            f"file for key '{key}'. "
+            f"Failed to validate '{maybe_file}'."
+        ) from e
```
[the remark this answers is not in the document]
SAKavli: Ye, this is valid, resolve_path does not really seem resolve paths after all

src/semeio/fmudesign/general_input_validation.py
```diff
@@ -0,0 +1,175 @@
...
+def validate_positive_int(maybe_int: Any, key: str) -> int:  # ruff: ignore[any-type]
+    try:
+        int_ = int(maybe_int)
+    except (TypeError, ValueError) as e:
+        raise PositiveIntValidationError(maybe_int, key, "positive int") from e
+
+    if int_ < 0:
+        raise PositiveIntValidationError(maybe_int, key, "positive int")
+
+    return int_
```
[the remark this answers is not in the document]
SAKavli: This is the same behavior as ert.
The same input will produce the same results every time, so I think this should be fine.

src/semeio/fmudesign/utils.py
```diff
@@ -214,3 +215,32 @@ def map_dependencies(
                     print(f" {from_} => {to_}")
 
     return df
+
+
+def resolve_path(input_filename: str, reference: str | None) -> str | None:
+    """The path `input_filename` is an Excel sheet, and `reference` is a cell
+    value that *might* be a reference to another file. Resolve the path to
+    `reference` and return. If no such file exists, return `reference`.
+    """
```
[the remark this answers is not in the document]
SAKavli: I dropped the resolve_path

src/semeio/fmudesign/config_validation.py
```diff
@@ -37,15 +36,6 @@ def validate_configuration(
     """
```
larsevj, line 37: The move of the validation into `_excel_to_dict` means that direct calls to DesignMatrix.generate is now not validated then?

SAKavli: Yes this is true.
`DesignMatrix.generate` is only called in one place in the code, but perhaps external libraries use this?

To be safe we can revert the last commit, then think about creating a validated pydantic class for the `config` dict (e.g. `DesignMatrixConfig`) in a later stage.

SAKavli:
We've discussed strategies of improving error messages, as pydantic can be quite verbose.
That will be resolved in a seperate issue.

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,120 @@
...
+        return value.strip()
+    # pd.isna(Collection) -> NDArray, which is ambiguous
+    if not isinstance(value, Collection) and pd.isna(value):  # type: ignore[call-overload]
+        return None
+    return value
+
+
+class GeneralInput(BaseModel):
+    designtype: Literal["onebyone"]
+    repeats: int
```
larsevj: pydantic has a `PositiveInt` and `NonNegativeInt` type

src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -49,29 +71,13 @@ def excel_to_dict(
...
-            header=None,
-            index_col=0,
-            engine="openpyxl",
-        )
-        .dropna(axis=0, how="all")
-        .dropna(axis=1, how="all")
-        .loc[:, 1]
-        .to_dict()
-    )
+    general_input_dict = _read_general_input(input_filename, general_input_sheet)
```
larsevj, line 74: Is it necessary to use this temp variable?

src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -198,95 +151,18 @@ def _excel_to_dict_onebyone(
...
     output: dict[str, Any] = {
-        "input_file": input_filename
+        "input_file": input_filename,
+        "designtype": generalinput.designtype,
+        "repeats": generalinput.repeats,
+        "distribution_seed": generalinput.distribution_seed,
+        "background": generalinput.background,
+        "seeds": generalinput.rms_seeds,
+        "correlation_iterations": generalinput.correlation_iterations,
+        "seed_strategy": generalinput.seed_strategy,
```
larsevj: Would maybe be safest to use `generalinput.seed_strategy.value` to ensure this is a string.

larsevj: to preserve behaviour when converting to yaml `inputdict_to_yaml`

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,120 @@
...
+    if isinstance(value, str):
+        return value.strip()
+    # pd.isna(Collection) -> NDArray, which is ambiguous
+    if not isinstance(value, Collection) and pd.isna(value):  # type: ignore[call-overload]
+        return None
+    return value
+
+
+class GeneralInput(BaseModel):
+    designtype: Literal["onebyone"]
```
larsevj, line 30: should maybe have model_config = ConfigDict(extra="forbid") ?

larsevj: Would allow to drop all the continue cases from the match case in from_dict

larsevj (commented):
See that there are some differences in validation between excel input and dict input.
Given excel input for instance you go through first the pydantic validation and then validate_configuration later, and the validation requirements are not the same.
Should probably be improved / consolidated at some point.

larsevj (commented):
Currently you get the pydantic url in the error message (for ValidationErrors), maybe we can catch it in the fmudesignrunner and exclude the url `include_url=False`?

src/semeio/fmudesign/_excel_to_dict.py
SAKavli: I think it reads nicely to split this into two lines, where it is emphasized that 
`.from_dict()` is called on a dict `general_input_dict`,
but if you have strong opinions we can oneline this.

Side by side comparison:
```py
general_input = GeneralInput.from_dict(
    _read_general_input(input_filename, general_input_sheet)
)
```
```py
general_input_dict = _read_general_input(input_filename, general_input_sheet)

general_input = GeneralInput.from_dict(general_input_dict)
```

src/semeio/fmudesign/_excel_to_dict.py
SAKavli: Yes, thanks for pointing this out!

src/semeio/fmudesign/general_input.py
SAKavli: Probably a good idea. 
The constructor is protected from extra keys already, but it's nice should `GeneralInput` be instantiated directly or gain new constructors. 👍

src/semeio/fmudesign/general_input.py
SAKavli: Didn't know that, thanks!
I made some custom ones as we want to disallow booleans and allow 0 for some cases, please review those.

SAKavli:
I've reworked the approach quite a bit from previous feedback, ended up squashing the fixups as I didn't bother constantly rebasing 12 fixup commits 😅

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,78 @@
...
+        input_folder = Path(input_filename).parent
+
+        rms_seeds = general_input.get("rms_seeds")
+        if isinstance(rms_seeds, str):
+            if (maybe_rms_seeds_file := input_folder / rms_seeds).is_file():
+                general_input["rms_seeds"] = str(maybe_rms_seeds_file)
+            elif not (
+                Path(rms_seeds).is_file() or rms_seeds is None or rms_seeds == "default"
+            ):
+                raise ValueError("rms_seeds must either be 'default' or a valid path.")
```
SAKavli: For rms_seeds we must control that the string passed in is an existing file or "default", and raise if it is any other string.

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,78 @@
...
+
+        background = general_input.get("background")
+        if isinstance(background, str):
+            if (maybe_background_file := input_folder / background).is_file():
+                general_input["background"] = str(maybe_background_file)
+            elif isinstance(background, str) and background.lower() == "none":
+                general_input["background"] = None
```
SAKavli: general_input may be a excel sheet, so we allow any strings as we don't validate existing excel sheets at this point.

src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -150,143 +132,55 @@ def _check_for_mixed_sensitivities(sens_name: str, sens_group: pd.DataFrame) ->
...
-        return str(reference_path.resolve())
-
-    raise ValueError(f"Failed to resolve path for file: {reference}")
-
-
 def _excel_to_dict_onebyone(
     input_filename: str,
     *,
-    general_input_sheet: str,
+    generalinput: GeneralInput,
```
larsevj: maybe consider `general_input`? for snake case convention?

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,78 @@
...
+                print(
+                    f"'{key}' not set in general input sheet. "
+                    f"Setting to default "
+                    f"{GeneralInput.model_fields[key].default}."
+                )
+
+        input_folder = Path(input_filename).parent
+
+        rms_seeds = general_input.get("rms_seeds")
+        if isinstance(rms_seeds, str):
```
larsevj: A bit confused why there is validation and checks of `rms_seeds` both here and in _excel_to_dict_onebyone ? Would it not be better to have the checks and resolve it once? Either here or in `_excel_to_dict_onebyone`?

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,78 @@
...
+            if (maybe_rms_seeds_file := input_folder / rms_seeds).is_file():
+                general_input["rms_seeds"] = str(maybe_rms_seeds_file)
+            elif not (
+                Path(rms_seeds).is_file() or rms_seeds is None or rms_seeds == "default"
+            ):
+                raise ValueError("rms_seeds must either be 'default' or a valid path.")
+
+        background = general_input.get("background")
```
larsevj: Apart from maybe converting "none" to None, is there any value in this? In addition to the handling in`_excel_to_dict_onebyone`?

src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -150,143 +132,55 @@ def _check_for_mixed_sensitivities(sens_name: str, sens_group: pd.DataFrame) ->
...
-        output[key] = {"extern": resolve_path(input_filename, background)}
-    else:
-        output[key] = _read_background(input_filename, background)
+            rms_seeds = seeds_from_extern(maybe_path)
+        else:
+            raise ValueError(
+                f"rms_seeds '{rms_seeds}' must be 'default', 'None' or relative path"
+            )
+
+    background: Any = generalinput.background
```
larsevj: Surely we know what values this must be in the end? Would be nice to avoid Any.

src/semeio/fmudesign/fmudesignrunner.py
```diff
@@ -312,6 +313,8 @@ def main() -> None:
     """semeio.fmudesign is a command line utility for generating design matrices
 
     Wrapper for the the semeio.fmudesign module"""
+
+    os.environ["PYDANTIC_ERRORS_INCLUDE_URL"] = "false"
```
larsevj: I think it would be nice to avoid mutating the global environment. Could maybe extend the try/except class, or just postpone this until later.

SAKavli:
The only issue I'm aware of now is the validation message of rms_seeds. The message is fine, but the error type is weird:
```
rms_seeds.function-after[validate_file(), lax-or-strict[lax=union[json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]],function-after[path_validator(), str]],strict=json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]]]]
  Path does not point to a file [type=path_not_file, input_value='defaultt', input_type=str]
rms_seeds.literal['default']
  Input should be 'default' [type=literal_error, input_value='defaultt', input_type=str]
```

src/semeio/fmudesign/fmudesignrunner.py
SAKavli: Added handling for `ValidationError`, example:
```
 fmudesign /Users/SKAV/git_repos/semeio/tests/fmudesign/data/config/design_input_onebyone.xlsx
Reading file: '/Users/SKAV/git_repos/semeio/tests/fmudesign/data/config/design_input_onebyone.xlsx'
'seed_strategy' not set in general input sheet. Setting to default joint.
'correlation_iterations' not set in general input sheet. Setting to default 0.
Validation error for 'designtype': Input should be 'onebyone', was 'onebyone1'
Validation error for 'repeats': Input should be greater than 0, was '-1'
Validation error for 'rms_seeds': Path does not point to a file, was 'defaultt'
Validation error for 'rms_seeds': Input should be 'default', was 'defaultt'

```
If we are still uncertain about how we want to handle this, we can postpone it.

src/semeio/fmudesign/_excel_to_dict.py
SAKavli: Yes, was battling typing. 
Is resolved by removing the validation from this function (handled in `GeneralInput`)

tests/fmudesign/test_fmudesignrunner.py
```diff
@@ -0,0 +1,16 @@
...
+
+@pytest.mark.integration_test
+def test_that_pydantic_validation_errors_do_not_include_urls(use_tmpdir, monkeypatch):
+    class Foo(BaseModel):
+        n: int
+
+    monkeypatch.setattr(fmudesignrunner, "get_parser", lambda: Foo(n="Not a int"))
+    with pytest.raises(ValidationError) as e:
+        main()
+    assert "https://errors.pydantic.dev" not in str(e)
```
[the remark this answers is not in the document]
SAKavli: Deleted file

src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -150,143 +132,50 @@ def _check_for_mixed_sensitivities(sens_name: str, sens_group: pd.DataFrame) ->
...
-    # or the name of a sheet in this workbook.
-    key = "background"
-    value = generalinput.get(key)
-    background = "" if value is None else str(value).strip()
-    if background.lower() in {"", "none"}:
-        output[key] = None
-    elif background.endswith(("csv", "xlsx")):
-        output[key] = {"extern": resolve_path(input_filename, background)}
+    if isinstance(bgr := general_input.background, Path):
+        background = {"extern": bgr}
```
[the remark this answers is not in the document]
SAKavli: Sure

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,83 @@
...
+                elif resolved.lower() == "none":
+                    general_input[key] = None
+                else:
+                    general_input[key] = resolved
+
+        return cls(**general_input)
+
+    @field_serializer("rms_seeds", "background")
+    def serialize_paths(self, field: Path) -> str:  # ruff: ignore[no-self-use]
+        return str(field)
```
larsevj, line 88: probably need an if check to only convert if is not None

src/semeio/fmudesign/general_input.py
```diff
@@ -0,0 +1,83 @@
...
+    return value
+
+
+class GeneralInput(BaseModel):
+    designtype: Literal["onebyone"]
+    repeats: PositiveInt
+    distribution_seed: NonNegativeInt | None
+    rms_seeds: FilePath | Literal["default"] | None
+    correlation_iterations: NonNegativeInt = 0
+    seed_strategy: SeedStrategy = SeedStrategy.JOINT
```
larsevj: Due to pydantics `use_enum_values=True` i think the default will differ from inputs, can set `Field(
     default=SeedStrategy.JOINT,
     validate_default=True,
 )`
to get the same behaviour in both cases.

larsevj:
> The only issue I'm aware of now is the validation message of rms_seeds. The message is fine, but the error type is weird:
> 
> ```
> rms_seeds.function-after[validate_file(), lax-or-strict[lax=union[json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]],function-after[path_validator(), str]],strict=json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]]]]
>   Path does not point to a file [type=path_not_file, input_value='defaultt', input_type=str]
> rms_seeds.literal['default']
>   Input should be 'default' [type=literal_error, input_value='defaultt', input_type=str]
> ```

Should be fine for now.

larsevj (approved):
Very good! 💯

src/semeio/fmudesign/general_input.py
SAKavli: Good call, also wrote test for this now.

src/semeio/fmudesign/general_input.py
SAKavli: Also good call, wrote test for this as well.

PR #956, opened 2026-08-25
Remove excel files from test data

### LINKED ISSUE #942: Rewrite tests to generate excel files on demand
We would like to avoid storing excel files(binary files) in test data. Should aim to rewrite tests to generate them when running tests instead.

larsevj:
Resolves #942 

In preparation of move of fmudesign into ert. Will leave excel files in examples untouched.


tests/fmudesign/test_use_cases.py
```diff
@@ -131,30 +127,13 @@ def test_constant_distribution(tmp_path, gen_input_sheet):
...
-        if filename.is_file():
-            shutil.copy2(filename, tmp_path)
-
-    _run_cli(designfile.name)
-    assert (tmp_path / "generateddesignmatrix.xlsx").is_file()
-
-
 @pytest.mark.integration_test
 @pytest.mark.parametrize("verbosity", [1, 2])
 def test_cli_verbosity_levels(tmp_path, monkeypatch, verbosity):
```
SAKavli, line 133: Is this testing anything in particular? If it's just running the `ex4_background_parameters.xslx` an additional time but with verbose flag and no asserts, we could consider removing this test imo.

I can't see many cases where verbosity is the cause of failure for a smoke test.

tests/fmudesign/test_create_design.py
```diff
@@ -486,6 +559,18 @@ def test_generate_background(tmp_path, monkeypatch):
...
     assert (background_vals.to_numpy() == velmodel_vals.to_numpy()).all()
 
+    sens9_vals = design.designvalues.loc[
+        design.designvalues["SENSNAME"] == "sens9", background_params
+    ]
+    assert design.backgroundvalues is not None
+    sampled_background = design.backgroundvalues[background_params]
+    assert sens9_vals.shape == sampled_background.shape
+    for parameter in background_params:
+        assert not np.array_equal(
```
SAKavli, line 572: Why are we testing that sens9 and background params are different?
Is there a specific property which should cause this?
Perhaps it could be clarified why this assert is here.

tests/fmudesign/test_use_cases.py
larsevj: Verbosity runs plotting, so is just to ensure that it runs without failure. "ex4_background_parameters.xlsx" just exercises a lot of these paths due to extensive config.
I can add some asserts.

tests/fmudesign/test_create_design.py
larsevj: have added some comments. It is to assert that the background params only fill in for a specific sensitivity where it is not specified.
I am not sure if this is a useful thing in practise, but it is how the code works as of now.

SAKavli (approved):
🎉

PR #957, opened 2026-08-27
Simplify resolve_path

### LINKED ISSUE #770: fmudesign: using file input from RMS seeds
Comment from Helena:

> I tried (not too hard) and failed to use the 3rd option "filename for file" in rms-seeds. I think I tried to add a .txt file on the same folder as the design matrix. Maybe some additional explanation on how to use it would suffice, or maybe it is not working as intended. If the scope of your work could include to check it more in detail, that is one suggestion.

Image

SAKavli:
Resolves #770

### DIFF
--- a/src/semeio/fmudesign/_excel_to_dict.py
+++ b/src/semeio/fmudesign/_excel_to_dict.py
@@ -238,7 +238,7 @@ def _excel_to_dict_onebyone(
 
         elif sens_type == "extern":
             sensdict["extern_file"] = resolve_path(
-                input_filename, str(group["extern_file"].iloc[0])
+                str(group["extern_file"].iloc[0]), input_filename
             )
             sensdict["senstype"] = sens_type
             sensdict["parameters"] = list(group["param_name"])
--- a/src/semeio/fmudesign/_excel_to_dict.py
+++ b/src/semeio/fmudesign/_excel_to_dict.py
@@ -238,7 +238,7 @@ def _excel_to_dict_onebyone(
 
         elif sens_type == "extern":
             sensdict["extern_file"] = resolve_path(
-                str(group["extern_file"].iloc[0]), input_filename
+                str(group["extern_file"].iloc[0]), base_file=input_filename
             )
             sensdict["senstype"] = sens_type
             sensdict["parameters"] = list(group["param_name"])
--- a/src/semeio/fmudesign/utils.py
+++ b/src/semeio/fmudesign/utils.py
@@ -262,15 +262,15 @@ def _is_int(teststring: str) -> bool:
 
 
 def resolve_path(target: str | None, base_file: str = "") -> str | None:
-    """Try to resolve the path of potential file 'target' as path either from cwd or
-    relative to 'base_file'."""
+    """Try to resolve the path of potential file 'target' as path either relative to
+    'base_file' or cwd."""
     if target is None:
         return None
 
-    if (path := Path(target)).is_file():
-        return str(path.resolve())
-
     if (relative_path := Path(base_file).parent / target).is_file():
         return str(relative_path.resolve())
+
+    if (path := Path(target)).is_file():
+        return str(path.resolve())
 
     return target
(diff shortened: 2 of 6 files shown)


src/semeio/fmudesign/utils.py
```diff
@@ -261,30 +261,16 @@ def _is_int(teststring: str) -> bool:
...
+    """Try to resolve the path of potential file 'target' as path either from cwd or
+    relative to 'base_file'."""
+    if target is None:
+        return None
 
-    # If the reference is e.g. 'doe1.xlsx'
-    full_path = Path(input_filename).parent / reference_path
-    if full_path.exists():
-        return str(full_path.resolve())
+    if (path := Path(target)).is_file():
```
larsevj, line 280: if target is a relative path with a file existing both in relation to `cwd` and `base_file`, then I think this reverses the earlier behaviour, choosing relative to `cwd` before `base_file`?

src/semeio/fmudesign/utils.py
```diff
@@ -261,30 +261,16 @@ def _is_int(teststring: str) -> bool:
...
+        return str(path.resolve())
 
-    if reference_path.exists():
-        return str(reference_path.resolve())
+    if (relative_path := Path(base_file).parent / target).is_file():
+        return str(relative_path.resolve())
 
-    raise ValueError(f"Failed to resolve path for file: {reference}")
+    return target
```
larsevj, line 283: I think the removal of the ValueError works well for both `background` and `rms_seeds`, but for `extern` this must be a file, and so might need to then raise the validation error somewhere else now.

src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -238,7 +238,7 @@ def _excel_to_dict_onebyone(
 
         elif sens_type == "extern":
             sensdict["extern_file"] = resolve_path(
-                input_filename, str(group["extern_file"].iloc[0])
+                str(group["extern_file"].iloc[0]), input_filename
```
larsevj: Ref other comment, might need validation here now?

src/semeio/fmudesign/utils.py
```diff
@@ -261,30 +261,16 @@ def _is_int(teststring: str) -> bool:
...
-
-    # It's a string but not a reference to another file
-    if not str(reference).endswith(("xlsx", "csv")):
-        return reference
-
-    # If the reference is e.g. 'C:/Users/USER/files/doe1.xlsx'
-    reference_path = Path(reference)
-    if reference_path.is_absolute() and reference_path.exists():
-        return str(reference_path.resolve())
+def resolve_path(target: str | None, base_file: str = "") -> str | None:
```
larsevj: Since this reverses argument order, might also consider making base_file a keyword argument;
```
          *,                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
          base_file: str | None = None,
```
, and having "" as default would mean checking CWD potentially twice.

src/semeio/fmudesign/utils.py
SAKavli: True, I can swap the order to maintain previous behavior 👍

src/semeio/fmudesign/utils.py
SAKavli: You are right about that.
The extern files end up in `seeds_from_extern` and `parameters_from_extern` where `seeds_from_extern` is called only once upon `rms_seeds` which is already validated by `GeneralInput`.

Will add a check for `parameters_from_extern` 👍

src/semeio/fmudesign/_excel_to_dict.py
SAKavli: Does https://github.com/equinor/semeio/pull/957/changes/50b83aa4a01445be773ec475e6e82bdeddb5928b suffice?

src/semeio/fmudesign/utils.py
SAKavli: Fixed!

larsevj (approved):

PR #965, opened 2026-08-28
Change from pandas to polars in GeneralInput

SAKavli:
As polars have stricter typing of columns, the behavior is now changed to reading the excel file into all strings in the dataframe.
This new assumption eases some of the logic in `GeneralInput`, while we leave the casting to the correct datatypes to Pydantic.

At the same time, `_read_general_input` is moved into the `GeneralInput` class as a static method, as it is only used there.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,8 @@ dependencies = [
     "seaborn",
     "probabilit>=0.4.1",
     "resfo-utilities>=0.1.1",
+    "polars",
+    "fastexcel",
 ]
 [dependency-groups]
 dev = [
@@ -61,7 +63,8 @@ dev = [
     'types-openpyxl',
     'types-seaborn',
     'types-setuptools',
-    'xlwt'
+    'xlwt',
+    'xlsxwriter'
 ]
 
 [project.urls]
--- a/src/semeio/fmudesign/_excel_to_dict.py
+++ b/src/semeio/fmudesign/_excel_to_dict.py
@@ -10,7 +10,6 @@ from typing import Any, Literal, cast
 
 import openpyxl
 import pandas as pd
-import polars as pl
 import yaml
 
 from semeio.fmudesign.general_input import GeneralInput
@@ -22,28 +21,6 @@ from semeio.fmudesign.utils import (
     resolve_path,
     seeds_from_extern,
 )
-
-
-def _read_general_input(
-    input_filename: str, general_input_sheet: str
-) -> dict[str, str | None]:
-    df = pl.read_excel(
-        input_filename,
-        sheet_name=general_input_sheet,
-        has_header=False,
-        read_options={"dtypes": "string"},
-        columns=[0, 1],
-    )
-    df = df.with_columns(
-        pl.col(col).str.strip_chars().alias(col) for col in df.columns
-    ).with_columns(
-        pl.when(pl.col(df.columns[1]).str.to_lowercase().is_in(["none", "null"]))
-        .then(None)
-        .otherwise(pl.col(df.columns[1]))
-        .alias(df.columns[1])
-    )
-    df = df.filter(pl.any_horizontal(pl.all().is_not_null()))
-    return dict(df.rows())
 
 
 def excel_to_dict(
@@ -75,9 +52,7 @@ def excel_to_dict(
     design_input_sheet = find_sheet(design_input_sheet, names=xlsx.sheetnames)
     default_values_sheet = find_sheet(default_values_sheet, names=xlsx.sheetnames)
 
-    general_input_dict = _read_general_input(input_filename, general_input_sheet)
-
-    general_input = GeneralInput.from_dict(general_input_dict, input_filename)
+    general_input = GeneralInput.from_xlsx(input_filename, general_input_sheet)
 
     return _excel_to_dict_onebyone(
         input_filename=input_filename,
--- a/src/semeio/fmudesign/_excel_to_dict.py
+++ b/src/semeio/fmudesign/_excel_to_dict.py
@@ -10,6 +10,7 @@ from typing import Any, Literal, cast
 
 import openpyxl
 import pandas as pd
+import polars as pl
 import yaml
 
 from semeio.fmudesign.general_input import GeneralInput
@@ -25,21 +26,24 @@ from semeio.fmudesign.utils import (
 
 def _read_general_input(
     input_filename: str, general_input_sheet: str
-) -> dict[str, Any]:
-    general_input = (
-        pd.read_excel(
-            input_filename,
-            general_input_sheet,
-            header=None,
-            engine="openpyxl",
-        )
-        .dropna(axis=0, how="all")
-        .dropna(axis=1, how="all")
-        .set_index(0)
-        .loc[:, 1]
-        .to_dict()
+) -> dict[str, str | None]:
+    df = pl.read_excel(
+        input_filename,
+        sheet_name=general_input_sheet,
+        has_header=False,
+        read_options={"dtypes": "string"},
+        columns=[0, 1],
     )
-    return {str(k): v for k, v in general_input.items()}
+    df = df.with_columns(
+        pl.col(col).str.strip_chars().alias(col) for col in df.columns
+    ).with_columns(
+        pl.when(pl.col(df.columns[1]).str.to_lowercase().is_in(["none", "null"]))
+        .then(None)
+        .otherwise(pl.col(df.columns[1]))
+        .alias(df.columns[1])
+    )
+    df = df.filter(pl.any_horizontal(pl.all().is_not_null()))
+    return dict(df.rows())
 
 
 def excel_to_dict(
--- a/src/semeio/fmudesign/general_input.py
+++ b/src/semeio/fmudesign/general_input.py
@@ -43,7 +43,11 @@ class GeneralInput(BaseModel):
         df = df.with_columns(
             pl.col(col).str.strip_chars().alias(col) for col in df.columns
         ).with_columns(
-            pl.when(pl.col(df.columns[1]).str.to_lowercase().is_in(["none", "null"]))
+            pl.when(
+                pl.col(df.columns[1])
+                .str.to_lowercase()
+                .is_in({"none", "null", "na", "nan"})
+            )
             .then(None)
             .otherwise(pl.col(df.columns[1]))
             .alias(df.columns[1])
--- a/src/semeio/fmudesign/general_input.py
+++ b/src/semeio/fmudesign/general_input.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 from typing import Literal, Self
 
+import polars as pl
 from pydantic import (
     BaseModel,
     ConfigDict,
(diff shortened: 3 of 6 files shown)


src/semeio/fmudesign/general_input.py
```diff
@@ -39,27 +35,45 @@ class GeneralInput(BaseModel):
...
-    def from_dict(cls, inputdict: dict[str, Any], input_filename: str = "") -> Self:
-        general_input: dict[str, Any] = {
-            str(key).strip(): parse_value(value) for key, value in inputdict.items()
-        }
+    @staticmethod
+    def _read_general_input(
+        input_filename: str, general_input_sheet: str
+    ) -> dict[str, str | None]:
+        df = pl.read_excel(
+            input_filename,
```
larsevj, line 37: could try the column keyword here to select first two columns directly?

src/semeio/fmudesign/general_input.py
```diff
@@ -39,27 +35,45 @@ class GeneralInput(BaseModel):
...
+            sheet_name=general_input_sheet,
+            has_header=False,
```
larsevj, line 39: forcing strings withs `read_options={"dtypes": "string"}` would avoid conversion later on.

src/semeio/fmudesign/general_input.py
```diff
@@ -39,27 +35,45 @@ class GeneralInput(BaseModel):
...
+        )
+        df = df.select(df.columns[:2])
+        df = df.with_columns(
+            pl.when(pl.col(col).str.to_lowercase().is_in(["none", "null"]))
```
larsevj: could strip all entries first to also allow for ` none` or `  null`

larsevj: or could move that check into parse_value and centralize.

tests/fmudesign/test_general_input.py
```diff
@@ -1,76 +1,44 @@
+from io import BytesIO
 from pathlib import Path
 from types import NoneType
 from typing import Any
 
 import hypothesis.strategies as st
-import numpy as np
-import pandas as pd
+import openpyxl
```
larsevj: would prefer xlsxwriter instead of openpyxl as that is what ert is using.

src/semeio/fmudesign/general_input.py
```diff
@@ -39,27 +35,45 @@ class GeneralInput(BaseModel):
...
-        # pydantic's validation. It can be caught using strict=True, but that
-        # removes the flexibility of allowing numeric strings for numeric fields.
-        # No values should be boolean, so we check for all.
-        for key, value in general_input.items():
-            if isinstance(value, bool):
-                raise ValueError(
-                    f"key '{key}' cannot have boolean value, got '{value}'"
-                )
+    @classmethod
+    def from_xlsx(cls, input_filename: str, general_input_sheet: str) -> Self:
```
larsevj, line 60: could not find any tests using this path? `from_xlsx`

tests/fmudesign/test_general_input.py
```diff
@@ -428,3 +363,81 @@ def test_that_background_path_is_serialized_as_string():
...
+        "rms_seeds": "default",
+        "distribution_seed": None,
+    }
+
+
+@pytest.mark.parametrize("none", [None, "none", "None", "NONE", "null", "NULL"])
+def test_that_null_rows_in_excel_is_filtered_out_in_read_general_input(none):
+    rows = [
+        ["designtype", "onebyone"],
+        [none, ""],
```
larsevj: should this maybe be key error? someone has then actually entered "none" as an `option`

tests/fmudesign/test_general_input.py
```diff
@@ -118,56 +80,46 @@ def test_that_designtype_onebyone_is_accepted():
     assert result.designtype == "onebyone"
 
 
-@given(ANY_TYPE)
+@given(st.text())
 def test_that_other_design_types_than_onebyone_raises_validation_error(text):
     assume(text != "onebyone")
```
larsevj, line 79: should maybe be `text.strip()` ?

src/semeio/fmudesign/general_input.py
SAKavli: Smart! 👍

src/semeio/fmudesign/general_input.py
SAKavli: Good idea!

tests/fmudesign/test_general_input.py
SAKavli: 🤷 
It was part of the old behavior:
```py
.dropna(axis=0, how="all")
.dropna(axis=1, how="all")
```
But I don't disagree with you.

tests/fmudesign/test_general_input.py
SAKavli: True, I will rewrite the strategy to only provide stripped text, as we expect the input to be stripped by `_read_general_input`

src/semeio/fmudesign/general_input.py
SAKavli: Will add it to `_read_general_input` then there's no need to parse values anymore.

src/semeio/fmudesign/general_input.py
SAKavli: Will write some tests 👍

tests/fmudesign/test_general_input.py
larsevj: i think that is only if it is empty? Or have you checked if `none` was dropped before?

src/semeio/fmudesign/_excel_to_dict.py
```diff
@@ -31,17 +31,19 @@ def _read_general_input(
         input_filename,
         sheet_name=general_input_sheet,
         has_header=False,
+        columns=[0, 1],
     )
-    df = df.select(df.columns[:2])
     df = df.with_columns(
+        pl.col(col).str.strip_chars().alias(col) for col in df.columns
```
SAKavli: This strips once beforehand to avoid stripping both in the None check and the `otherwise(pl.col(col).str.strip_chars()` section.

This now also captures cases like `"  None  "` -> `None`

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
     "seaborn",
     "probabilit>=0.4.1",
     "resfo-utilities>=0.1.1",
+    "xlsxwriter>=3.2.9",
```
larsevj: this should probably be a dev dependancy
and polars and fastexcel should be added as a direct dependency
(although both will be removed again when porting over to ert)

tests/fmudesign/test_general_input.py
larsevj: On main i get the following:
```
Reading file: 'src/semeio/fmudesign/examples/fmudesign_ex_onebyone.xlsx'
'correlation_iterations' not set in general input sheet. Setting to default 0.
Validation error for 'none': Extra inputs are not permitted, was 'None'

 
fmudesign failed. Read the error message above and fix the input file.
 - Documentation:           https://equinor.github.io/fmu-tools/fmudesign.html
 - Course docs:             https://fmu-docs.equinor.com/docs/fmu-coursedocs/fmu-howto/sensitivities/index.html 
 - Issues/feature requests: https://github.com/equinor/semeio/issues
If you believe this error is a bug or are unable to fix it, create an issue or contact the scout team 
```

SAKavli: The first commit adds tests for this before any functional changes take place for values of None.
I noticed now it doesnt have all the stringified variants as the later commits, so I will add that.
I've now tested that:
```py
    rows = [
        ["designtype", "onebyone"],
        ["none", "null"],
        ["NULL", "None"],
        [None, None],
        ["repeats", 10],
        ["rms_seeds", "default"],
        ["distribution_seed", none],
    ]
```
Generated this excel sheet:
image
Where the rows containing the various None values are dropped in parsing and the test passes.

pyproject.toml
SAKavli: Oh, I did not notice polars was not in there.
Is it transitive dependency of Ert?
I can add it 👍

tests/fmudesign/test_general_input.py
SAKavli: So `None`, `null` and `NULL` are valid cell values without fmudesign failing - also on main.

SAKavli: Then the question is whether it's ok to extend these allowed values with all capitalization versions?

src/semeio/fmudesign/general_input.py
```diff
@@ -39,27 +29,47 @@ class GeneralInput(BaseModel):
...
-        for key, value in general_input.items():
-            if isinstance(value, bool):
-                raise ValueError(
-                    f"key '{key}' cannot have boolean value, got '{value}'"
-                )
+    @classmethod
+    def from_dict(
+        cls, input_dict: dict[str, str | None], input_filename: str = ""
+    ) -> Self:
+        general_input: dict[str, str | Path | None] = dict(input_dict.items())
```
SAKavli, line 71: Typing disallows assigning input_dict to general_input due to the difference between
`dict[str, str | None]`
and
`dict[str, str | Path | None]`

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
     "seaborn",
     "probabilit>=0.4.1",
     "resfo-utilities>=0.1.1",
+    "polars>=1.43.2",
```
larsevj: also needs `fastexcel` for excel reading.

src/semeio/fmudesign/general_input.py
```diff
@@ -39,27 +29,47 @@ class GeneralInput(BaseModel):
...
+        df = df.filter(pl.any_horizontal(pl.all().is_not_null()))
+        return dict(df.rows())
+
     @classmethod
-    def from_dict(cls, inputdict: dict[str, Any], input_filename: str = "") -> Self:
-        general_input: dict[str, Any] = {
-            str(key).strip(): parse_value(value) for key, value in inputdict.items()
-        }
+    def from_xlsx(cls, input_filename: str, general_input_sheet: str) -> Self:
+        input_dict = GeneralInput._read_general_input(
```
larsevj: Might be cleaner to use `cls` instead of `GeneralInput` here, since it uses Self, and cls elsewhere.

tests/fmudesign/test_general_input.py
larsevj: What about `[none, "some_value"]` ?

pyproject.toml
```diff
@@ -41,6 +41,8 @@ dependencies = [
...
```
larsevj: These pins to absolute newest versions would break bleeding.

SAKavli: Should we just remove the pin for these versions or change to different pin?

larsevj: No pins should be fine

tests/fmudesign/test_general_input.py
larsevj: Then i get the error message:
```
Reading file: 'src/semeio/fmudesign/examples/fmudesign_ex_onebyone.xlsx'
'seed_strategy' not set in general input sheet. Setting to default joint.
'correlation_iterations' not set in general input sheet. Setting to default 0.
  File "/Users/LEVJE/dev_komodo/semeio/src/semeio/fmudesign/fmudesignrunner.py", line 343, in main
    args.func(args)
    ~~~~~~~~~^^^^^^
  File "/Users/LEVJE/dev_komodo/semeio/src/semeio/fmudesign/fmudesignrunner.py", line 260, in subcommand_run
    config = excel_to_dict(
        args.config,
    ...<2 lines>...
        default_val_sheet=args.defaultvalues,
    )
  File "/Users/LEVJE/dev_komodo/semeio/src/semeio/fmudesign/_excel_to_dict.py", line 55, in excel_to_dict
    general_input = GeneralInput.from_xlsx(input_filename, general_input_sheet)
  File "/Users/LEVJE/dev_komodo/semeio/src/semeio/fmudesign/general_input.py", line 60, in from_xlsx
    return cls.from_dict(
           ~~~~~~~~~~~~~^
        input_dict,
        ^^^^^^^^^^^
        input_filename,
        ^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/LEVJE/dev_komodo/semeio/src/semeio/fmudesign/general_input.py", line 88, in from_dict
    return cls(**general_input)
TypeError: keywords must be strings

 
fmudesign failed. Read the error message above and fix the input file.
```

used to be:
```
Reading file: 'src/semeio/fmudesign/examples/fmudesign_ex_onebyone.xlsx'
'correlation_iterations' not set in general input sheet. Setting to default 0.
Validation error for 'nan': Extra inputs are not permitted, was 'some'
```

which was a bit clearer although not great.

SAKavli:
One unhandled edge case is given empty general_input sheet, but this raises a nasty traceback on main as well.

src/semeio/fmudesign/general_input.py
```diff
@@ -39,27 +29,45 @@ class GeneralInput(BaseModel):
...
+            input_filename,
+            sheet_name=general_input_sheet,
+            has_header=False,
+            read_options={"dtypes": "string"},
+            columns=[0, 1],
+        )
+        df = df.with_columns(
+            pl.col(col).str.strip_chars().alias(col) for col in df.columns
+        ).with_columns(
+            pl.when(pl.col(df.columns[1]).str.to_lowercase().is_in(["none", "null"]))
```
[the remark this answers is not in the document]
larsevj: Could consider adding these to the list of is_in. (N/A apparently also used to work)

larsevj (approved):
Good job 🚀

PR #967, opened 2026-09-02
Support resdata 6 and 7 GRDECL writers

ajaust:
Use the legacy cwrap file interface for resdata 6 and a standard text stream for resdata 7 and later.

Tested locally with `resdata@main`


ajaust:
If we are happy with the changes, I will open an issue to track the compatibility fixes introduced to support `resdata<7` and `resdata>=7` at the same time. Once `resdata>=7` is published, we can bump the resdata version required by semeio and stick to resdata's new API.

andreas-el (approved):
Nicely done! 💪

ajaust:
Created issue to track these changes: https://github.com/equinor/semeio/issues/968

PR #969, opened 2026-09-02
Remove fmudesign from semeio

SAKavli:


larsevj:
Could also remove probabilit entry from mypy.ini file

SAKavli:
Also got rid of `mypy-sklearn` in `mypy.ini` as that was transitive dependency

pyproject.toml
larsevj: could remove this as well

larsevj (approved):
🚀

